PR #487, opened 2020-02-10
remove unnecessary reference to LDK

moneyball:


ariard:
ACK, will update my PR in consequence. (But still how someone who heard about LDK but think Rust-Lightning can be redirected to the LDK slack ?)

moneyball:
Yes we need to address LDK + RL but let's do that separately.

PR #494, opened 2020-02-12
Modular handshake

arik-so:
Refactor peer channel encryptor into a whole separate, isolated module and abstract away most of the internals of handshake handling from the peer handler


lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,173 @@
+use ln::peers::{chacha, hkdf};
+
+/// Returned after a successful handshake to encrypt and decrypt communication with peer nodes
```
TheBlueMatt: Can you add more details about how you'd use this, what the requirements are for use, etc? Same goes for pretty much everything pub in this PR.

lightning/src/ln/peers/chacha.rs
```diff
@@ -0,0 +1,34 @@
+use util::chacha20poly1305rfc::ChaCha20Poly1305RFC;
+
```
TheBlueMatt: Why is this in ln::peers? It seems to be pure crypto functions.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,173 @@
...
+
+	pub(super) read_buffer: Option<Vec<u8>>,
+}
+
+impl Conduit {
+	pub fn encrypt(&mut self, buffer: &[u8]) -> Vec<u8> {
+		let length = buffer.len() as u16;
+		let length_bytes = length.to_be_bytes();
+
+		let encrypted_length = chacha::encrypt(&self.sending_key, self.sending_nonce as u64, &[0; 0], &length_bytes);
```
TheBlueMatt: Oof. Lets avoid allocating a Vec for two bytes somehow.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,173 @@
...
+		let encrypted_message = chacha::encrypt(&self.sending_key, self.sending_nonce as u64, &[0; 0], buffer);
+		self.increment_sending_nonce();
+
+		let mut ciphertext = encrypted_length;
+		ciphertext.extend_from_slice(&encrypted_message);
+		ciphertext
+	}
+
+	pub(super) fn read(&mut self, data: &[u8]) {
+		let mut read_buffer = if let Some(buffer) = self.read_buffer.take() {
```
TheBlueMatt: This would probably be more readable with https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert

lightning/src/ln/peers/handshake/acts.rs
```diff
@@ -0,0 +1,33 @@
...
+);
+
+pub enum Act {
+	One(ActOne),
+	Two(ActTwo),
+	Three(ActThree),
+}
+
+impl Act {
+	pub fn serialize(&self) -> Vec<u8> {
```
TheBlueMatt: Can we name this something else? I almost responded with "why the hell would you ever want to serialize an Act for storage?" before I realized it meant "write act to a vec to send to a peer" (also, can we not return a slice here?).

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,398 @@
+use bitcoin_hashes::{Hash, HashEngine};
+use bitcoin_hashes::sha256::Hash as Sha256;
+use rand::{Rng, thread_rng};
```
TheBlueMatt: We don't support rand because it a) often breaks MSRV, and b) doesn't support lots of platforms (read:embedded) that we do, where RNG access is via some proprietary API. Any random values you want need to be passed in from the user.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+		let static_public_key = Self::private_key_to_public_key(&self.private_key);
+		let tagged_encrypted_pubkey = chacha::encrypt(&temporary_key, 1, &hash.value, &static_public_key.serialize());
+		hash.update(&tagged_encrypted_pubkey);
+
+		let ecdh = Self::ecdh(&self.private_key, &remote_ephemeral_public_key);
+		let (chaining_key, temporary_key) = hkdf::derive(&chaining_key, &ecdh);
+		let authentication_tag = chacha::encrypt(&temporary_key, 0, &hash.value, &[0; 0]);
+		let (sending_key, receiving_key) = hkdf::derive(&chaining_key, &[0; 0]);
+
+		let mut act_three_vec = [0u8].to_vec();
```
TheBlueMatt: We know how long this is going to be, we should just allocate a buffer instead of malloc()ing (and a few other places in this file).

lightning/src/ln/peers/chacha.rs
arik-so: Because the AEAD-based encryption methods are only used for handshakes and peer message encryption IIRC, and not for the onion construction.

lightning/src/ln/peers/conduit.rs
arik-so: Are you sure this is the right place to shave off memory usage at the expense of legibility? I'm a bit hesitant to do the whole decrypt in place approach.

lightning/src/ln/peers/chacha.rs
TheBlueMatt: Right, but its also a pure-crypto primitive. I guess my preference is for such things (even if it implements a lightning protocol crypto primitive) to be in some kind of crypto module.

lightning/src/ln/peers/conduit.rs
arik-so: Ooooh, TIL. There are a bunch of places where I will replace that.

lightning/src/ln/peers/conduit.rs
TheBlueMatt: I dont think it changes legibility, and its not a memory usage question, its a question of heap fragmentation and significant performance penalty to hit the heap. I'm dubious you can't make it readable without using a Vec.

lightning/src/ln/peers/handshake/acts.rs
arik-so: Perhaps `to_vec`? We could return a slice, but would have to mark its lifetime as linked to the act object's.

TheBlueMatt: What's wrong with using lifetimes? The callsite can chose to store it in a Vec or copy to some local buffer if they wont want it.

lightning/src/ln/peers/handshake/mod.rs
arik-so: Ok, I'll try to refactor it to have the user pass in entropy.

arik-so:
> Could you leave moving macros and reducing indentations to a separate PR? For the latter, you can introduce scopes if needed. This will make reviewing the PR much easier as the diff will be smaller. It will also help me avoid some nasty merge conflicts when rebasing. :)

I'm afraid not, because some versions of Rust are confused about variable scopes with macro invocation.

arik-so:
But for what it's worth, the only file that I am really modifying that existed prior is peer_handler.rs. And this diff is gonna impact it pretty severely one way or another.

jkczyz:
> > Could you leave moving macros and reducing indentations to a separate PR? For the latter, you can introduce scopes if needed. This will make reviewing the PR much easier as the diff will be smaller. It will also help me avoid some nasty merge conflicts when rebasing. :)
> 
> I'm afraid not, because some versions of Rust are confused about variable scopes with macro invocation.

Actually, I may have been a bit confused here. The primary cause of all unrelated diffs is the indentation change. Could you introduce scopes to prevent this?

lightning/src/ln/peer_handler.rs
```diff
@@ -501,266 +531,214 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
 								}
 							}
 
-							macro_rules! insert_node_id {
```
jkczyz: Why remove this? We would no longer insert into the map without it, AFAICT. This breaks other code that needs the map to look up peers by id.

arik-so: None of the unit tests are failing, but I will re-add it to the graph, and then think of unit tests for this in a separate diff

arik-so:
> > > Could you leave moving macros and reducing indentations to a separate PR? For the latter, you can introduce scopes if needed. This will make reviewing the PR much easier as the diff will be smaller. It will also help me avoid some nasty merge conflicts when rebasing. :)
> > 
> > 
> > I'm afraid not, because some versions of Rust are confused about variable scopes with macro invocation.
> 
> Actually, I may have been a bit confused here. The primary cause of all unrelated diffs is the indentation change. Could you introduce scopes to prevent this? FYI, this comment was suppose to be on `do_read_event`.

Honestly, I think the indentation change might be because I moved the macro definitions. But I stand by the moves :)

lightning/src/ln/peer_handler.rs
```diff
@@ -501,266 +544,214 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
-									peer.pending_outbound_buffer.push_back(act_three.to_vec());
-									peer.pending_read_buffer = [0; 18].to_vec(); // Message length header is 18 bytes
-									peer.pending_read_is_header = true;
-
-									peer.their_node_id = Some(their_node_id);
-									insert_node_id!();
-									let mut features = InitFeatures::supported();
-									if self.message_handler.route_handler.should_request_full_sync(&peer.their_node_id.unwrap()) {
-										features.set_initial_routing_sync();
+							match message {
```
jkczyz: If you insert three scopes around this, the diffs become saner.

lightning/src/ln/peer_handler.rs
```diff
@@ -450,26 +463,55 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
+							}
 						}
 
-						if peer.pending_read_buffer_pos == peer.pending_read_buffer.len() {
-							peer.pending_read_buffer_pos = 0;
+						if let Some(key) = remote_pubkey_option {
+							peer.their_node_id = Some(key);
+						}
+
+						if let Some(conduit) = conduit_option {
```
jkczyz: Can't all this be done within the `PeerState::Handshake` block above? Then you wouldn't need `conduit_option` and `remote_pubkey_option` there with the de-structured moves in the `if let` statements above. You would just place this code there instead.

lightning/src/ln/peer_handler.rs
```diff
@@ -819,7 +814,9 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
 						let (mut descriptor, peer) = get_peer_for_forwarding!(node_id, {
 								//TODO: Drop the pending channel? (or just let it timeout, but that sucks)
 							});
-						peer.pending_outbound_buffer.push_back(peer.channel_encryptor.encrypt_message(&encode_msg!(msg)));
+						if let PeerState::Connected(ref mut conduit) = peer.encryptor {
+							peer.pending_outbound_buffer.push_back(conduit.encrypt(&encode_msg!(msg)));
+						}
```
jkczyz: If we have a peer in `node_id_to_descriptor`, it should also have a conduit. Any chance this can be done within `get_peer_for_forwarding` so the repetition across the method could be removed? We probably want to assert this as well.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	/// Decrypt a single message. Buffer is an undelimited amount of bytes
+	pub(crate) fn decrypt(&mut self, buffer: &[u8]) -> (Option<Vec<u8>>, usize) { // the response slice should have the same lifetime as the argument. It's the slice data is read from
+		if buffer.len() < 18 {
+			return (None, 0);
+		}
+
+		let encrypted_length = &buffer[0..18]; // todo: abort if too short
+		let length_vec = chacha::decrypt(&self.receiving_key, self.receiving_nonce as u64, &[0; 0], encrypted_length).unwrap();
```
jkczyz: Consider naming this `decrypted_length` rather than encoding the type in the name. It makes the relationship between it and `encrypted_length` above explicit. The actual types aren't important in understanding the code conceptually.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+		let mut length_bytes = [0u8; 2];
+		length_bytes.copy_from_slice(length_vec.as_slice());
+		let message_length = byte_utils::slice_to_be16(&length_bytes) as usize;
```
jkczyz: I think these three lines can be reduced to one line:

```rust
let message_length = byte_utils::slice_to_be16(&length_vec) as usize;
```

It would also reduce the number of "length" variables from four to three. :) I think that would help reduce the cognitive load.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+
+		let message_end_index = message_length + 18 + 16; // todo: abort if too short
```
jkczyz: Likewise, is this TODO addressed below?

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+		Self::increment_nonce(&mut self.sending_nonce, &mut self.sending_chaining_key, &mut self.sending_key);
+	}
+
+	fn increment_receiving_nonce(&mut self) {
+		Self::increment_nonce(&mut self.receiving_nonce, &mut self.receiving_chaining_key, &mut self.receiving_key);
+	}
+
+	fn increment_nonce(nonce: &mut u32, chaining_key: &mut [u8; 32], key: &mut [u8; 32]) {
+		*nonce += 1;
+		if *nonce == 1000 {
```
jkczyz: Please add a constant for this.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+		let mut chaining_key = [0u8; 32];
+		chaining_key.copy_from_slice(&chaining_key_vec);
+
+		let sending_key_vec = hex::decode("969ab31b4d288cedf6218839b27a3e2140827047f2c0f01bf5c04435d43511a9").unwrap();
+		let mut sending_key = [0u8; 32];
+		sending_key.copy_from_slice(&sending_key_vec);
+
+		let receiving_key_vec = hex::decode("bb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442").unwrap();
+		let mut receiving_key = [0u8; 32];
+		receiving_key.copy_from_slice(&receiving_key_vec);
```
jkczyz: Could you link to the test data in a comment for the test?

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+		assert_eq!(encrypted_messages[1000], hex::decode("4a2f3cc3b5e78ddb83dcb426d9863d9d9a723b0337c89dd0b005d89f8d3c05c52b76b29b740f09").unwrap());
+		assert_eq!(encrypted_messages[1001], hex::decode("2ecd8c8a5629d0d02ab457a0fdd0f7b90a192cd46be5ecb6ca570bfc5e268338b1a16cf4ef2d36").unwrap());
+
+		for _ in 0..1002 {
+			let encrypted_message = encrypted_messages.remove(0);
+			let mut decrypted_messages = remote_peer.decrypt_message_stream(Some(&encrypted_message));
+			assert_eq!(decrypted_messages.len(), 1);
+			let decrypted_message = decrypted_messages.remove(0);
+			assert_eq!(decrypted_message, hex::decode("68656c6c6f").unwrap());
+		}
```
jkczyz: This looks like it is testing a different behavior so it should probably be a different test.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+			let encrypted_message = connected_peer.encrypt(&message);
+			encrypted_messages.push(encrypted_message);
+		}
+
+		assert_eq!(encrypted_messages[0], hex::decode("cf2b30ddf0cf3f80e7c35a6e6730b59fe802473180f396d88a8fb0db8cbcf25d2f214cf9ea1d95").unwrap());
+		assert_eq!(encrypted_messages[1], hex::decode("72887022101f0b6753e0c7de21657d35a4cb2a1f5cde2650528bbc8f837d0f0d7ad833b1a256a1").unwrap());
+		assert_eq!(encrypted_messages[500], hex::decode("178cb9d7387190fa34db9c2d50027d21793c9bc2d40b1e14dcf30ebeeeb220f48364f7a4c68bf8").unwrap());
+		assert_eq!(encrypted_messages[501], hex::decode("1b186c57d44eb6de4c057c49940d79bb838a145cb528d6e8fd26dbe50a60ca2c104b56b60e45bd").unwrap());
+		assert_eq!(encrypted_messages[1000], hex::decode("4a2f3cc3b5e78ddb83dcb426d9863d9d9a723b0337c89dd0b005d89f8d3c05c52b76b29b740f09").unwrap());
+		assert_eq!(encrypted_messages[1001], hex::decode("2ecd8c8a5629d0d02ab457a0fdd0f7b90a192cd46be5ecb6ca570bfc5e268338b1a16cf4ef2d36").unwrap());
```
jkczyz: Looks like this is testing two different behaviors: chaining and key rotation? Ideally, tests should only test one behavior. Could we split this up? I realize this is the test data from BOLT 4, so happy to hear the argument if you think it should be one big test.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+			} else {
+				break;
+			}
+		}
+
+		messages
+	}
+
+	/// Decrypt a single message. Buffer is an undelimited amount of bytes
+	pub(crate) fn decrypt(&mut self, buffer: &[u8]) -> (Option<Vec<u8>>, usize) { // the response slice should have the same lifetime as the argument. It's the slice data is read from
```
jkczyz: This trailing comment seems incomplete. Could it be folded into the documentation comment? How about documenting the return type and the various cases?

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+
+		let mut hash = HandshakeHash::new(initial_hash_preimage.as_slice());
+		hash.update(&public_key.serialize());
+
+		(hash, chaining_key) // hash, chaining_key
+	}
+
+	/// Process act dynamically
+	/// The role must be set before this method can be called
+	pub fn process_act(&mut self, input: &[u8], remote_public_key: Option<&PublicKey>) -> Result<(Vec<u8>, Option<Conduit>, Option<PublicKey>), String> {
```
jkczyz: Some idle thoughts on this interface (`process_act`) and that of `process_act_two` and `process_act_three`:

Rather than returning tuples of `Option`s in `process_act` and the non-`Option` values in the other two methods, could they (i.e., `PublicKey` and `Conduit`) instead be set as fields in the appropriate act states (without using `Option`)? Then accessor methods on `PeerHandshake` could be provided for retrieving them (using `Option`). That way the code is a bit more self-documented and the return values of the act processing methods would be simpler (i.e., just returning the next Act to send).

lightning/src/ln/peer_handler.rs
```diff
@@ -300,12 +313,12 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
 	/// Panics if descriptor is duplicative with some other descriptor which has not yet has a
 	/// disconnect_event.
 	pub fn new_inbound_connection(&self, descriptor: Descriptor) -> Result<(), PeerHandleError> {
-		let peer_encryptor = PeerChannelEncryptor::new_inbound(&self.our_node_secret);
-		let pending_read_buffer = [0; 50].to_vec(); // Noise act one is 50 bytes
+		let mut handshake = PeerHandshake::new(&self.our_node_secret, &self.get_ephemeral_key());
+		handshake.make_inbound();
```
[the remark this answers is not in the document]
arik-so: Or rather new_inbound and new_outbound, sure

lightning/src/ln/peer_handler.rs
```diff
@@ -105,8 +107,23 @@ enum InitSyncTracker{
 	NodesSyncing(PublicKey),
 }
 
+enum PeerState {
+	Handshake(PeerHandshake),
```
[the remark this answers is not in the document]
arik-so: that works

lightning/src/ln/peer_handler.rs
```diff
@@ -105,8 +107,23 @@ enum InitSyncTracker{
...
+}
+
+impl PeerState {
+	fn is_ready_for_encryption(&self) -> bool {
+		if let &PeerState::Connected(_) = self {
+			true
+		} else {
+			false
+		}
+	}
```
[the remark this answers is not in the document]
arik-so: ok

lightning/src/ln/peer_handler.rs
```diff
@@ -450,26 +463,55 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
-							let data_to_copy = cmp::min(peer.pending_read_buffer.len() - peer.pending_read_buffer_pos, data.len() - read_pos);
-							peer.pending_read_buffer[peer.pending_read_buffer_pos..peer.pending_read_buffer_pos + data_to_copy].copy_from_slice(&data[read_pos..read_pos + data_to_copy]);
-							read_pos += data_to_copy;
-							peer.pending_read_buffer_pos += data_to_copy;
+						let mut conduit_option = None;
+						let mut remote_pubkey_option = None;
+
+						if let &mut PeerState::Handshake(ref mut handshake) = &mut peer.encryptor {
+							let (response, conduit, remote_pubkey) = handshake.process_act(&peer.pending_read_buffer, None).unwrap();
+							peer.pending_read_buffer.drain(..); // we read it all
```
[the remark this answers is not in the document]
arik-so: Ok, I've got an idea for this. Add a method to handshake to read only one message, but store the buffer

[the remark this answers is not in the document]
arik-so: Oh wait, that's already happening

lightning/src/ln/peer_handler.rs
```diff
@@ -450,26 +463,55 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
 						}
 
-						if peer.pending_read_buffer_pos == peer.pending_read_buffer.len() {
-							peer.pending_read_buffer_pos = 0;
+						if let Some(key) = remote_pubkey_option {
+							peer.their_node_id = Some(key);
+						}
+
+						if let Some(conduit) = conduit_option {
+							// Rust 1.22 does not allow assignment in a borrowed context, even if mutable
```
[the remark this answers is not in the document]
arik-so: This statement could have been done inside the big if clause preceding these single line assignments.

lightning/src/ln/peer_handler.rs
```diff
@@ -450,26 +463,55 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
+							};
+						}
 
+						if let &mut PeerState::Connected(ref mut conduit) = &mut peer.encryptor {
 							macro_rules! encode_and_send_msg {
 								($msg: expr) => {
 									{
 										log_trace!(self, "Encoding and sending message of type {} to {}", $msg.type_id(), log_pubkey!(peer.their_node_id.unwrap()));
-										peer.pending_outbound_buffer.push_back(peer.channel_encryptor.encrypt_message(&encode_msg!(&$msg)[..]));
+										// we are in a context where conduit is known
```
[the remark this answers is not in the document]
arik-so: I think it doesn't make sense because the other variables are assumed to be known, too. Once this is refactored from a macro into a regular method, it will solve itself.

lightning/src/ln/peer_handler.rs
arik-so: Not in rust 1.22

lightning/src/ln/peer_handler.rs
arik-so: Absolutely, though here too, I think that it will be much clearer once get_peer_for_forwarding is refactored from being a macro, which I strongly believe it doesn't need to be.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
+//! Handles all over the wire message encryption and decryption upon handshake completion.
+
+use ln::peers::{chacha, hkdf};
+use util::byte_utils;
+
+/// Returned after a successful handshake to encrypt and decrypt communication with peer nodes.
+/// It should not normally be manually instantiated.
+/// Automatically handles key rotation.
+/// For decryption, it is recommended to call `decrypt_message_stream` for automatic buffering.
+pub struct Conduit {
```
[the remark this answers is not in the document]
arik-so: This makes perfect sense and I would love to do that, though what would you call the module?

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+	pub(crate) sending_key: [u8; 32],
+	pub(crate) receiving_key: [u8; 32],
+
+	pub(crate) sending_chaining_key: [u8; 32],
+	pub(crate) receiving_chaining_key: [u8; 32],
```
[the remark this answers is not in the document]
arik-so: These aren't constants as much as they are types, but definitely.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,191 @@
...
+
+			if let Some(message) = current_message {
+				messages.push(message);
+			} else {
+				break;
+			}
+		}
+
+		messages
+	}
```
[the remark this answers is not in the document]
arik-so: I have a better idea. It will be in the next revision.

lightning/src/ln/peers/conduit.rs
arik-so: Done

lightning/src/ln/peers/conduit.rs
arik-so: Sure

lightning/src/ln/peers/conduit.rs
arik-so: You are right, it is testing both. Arguably the key rotation is part of the encryption. I can split it up into two tests if you like, but in a separate commit.

lightning/src/ln/peers/conduit.rs
arik-so: Ok

lightning/src/ln/peers/handshake/acts.rs
```diff
@@ -0,0 +1,38 @@
+/// Wrapper for the first act message
```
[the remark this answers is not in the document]
arik-so: Frankly, just easier navigation through the components in the file system.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+impl PeerHandshake {
+	/// Instantiate a new handshake with a node identity secret key and an ephemeral private key
+	pub fn new(private_key: &SecretKey, ephemeral_private_key: &SecretKey) -> Self {
+		let handshake = PeerHandshake {
+			state: Some(HandshakeState::Blank),
+			private_key: (*private_key).clone(),
+			ephemeral_private_key: (*ephemeral_private_key).clone(),
+			read_buffer: Vec::new(),
+		};
+		handshake
```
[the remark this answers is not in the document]
arik-so: good point

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+		sha.input(protocol_name);
+		let chaining_key = Sha256::from_engine(sha).into_inner();
+
+		let mut initial_hash_preimage = chaining_key.to_vec();
+		initial_hash_preimage.extend_from_slice(prologue.as_ref());
+
+		let mut hash = HandshakeHash::new(initial_hash_preimage.as_slice());
+		hash.update(&public_key.serialize());
+
+		(hash, chaining_key) // hash, chaining_key
```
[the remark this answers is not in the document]
arik-so: lol

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+
+		let mut chacha_tag = [0u8; 16];
+		chacha_tag.copy_from_slice(&act.0[50..66]);
+
+		let mut hash = act_three_expectation.hash;
+
+		let remote_pubkey_vec = chacha::decrypt(&act_three_expectation.temporary_key, 1, &hash.value, &tagged_encrypted_pubkey)?;
+		let mut remote_pubkey_bytes = [0u8; 33];
+		remote_pubkey_bytes.copy_from_slice(remote_pubkey_vec.as_slice());
+		// todo: replace unwrap with handleable error type
```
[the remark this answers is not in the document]
arik-so: yes, in a subsequent commit

lightning/src/ln/peers/conduit.rs
arik-so: ah, you meant the 1,000

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+		let mut response: Vec<u8> = Vec::new();
+		let mut connected_peer = None;
+		let mut remote_pubkey = None;
+
+		self.read_buffer.extend_from_slice(input);
+		let read_buffer_length = self.read_buffer.len();
+
+		match &self.state {
+			&Some(HandshakeState::Blank) => {
+				let remote_public_key = remote_public_key.ok_or("remote_public_key must be Some for outbound connections")?;
```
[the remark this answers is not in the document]
arik-so: in new_outbound in peer_handler.rs, the handshake is manually initiated because unlike new_inbound, that method immediately returns the first byte vector to send to the peer

lightning/src/ln/peers/handshake/mod.rs
arik-so: No, the act states are only what's necessary for the next step of the handshake. I think the handshake can be thought of as a chemical reaction, where there is an educt, a product, as well as some external inputs and byproducts. The handshake states are educts and products, and the conduit is a byproduct.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+				let act_one = self.initiate(&remote_public_key)?;
+				response = act_one.0.to_vec();
+			}
+			&Some(HandshakeState::AwaitingActOne(_)) => {
+				let act_length = 50;
```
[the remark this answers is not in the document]
arik-so: Done

lightning/src/ln/peer_handler.rs
jkczyz: Hmm, yeah for conduit at least. At very least you shouldn't have destruct the `Option` when assigning to `conduit_option`. You can simply unconditionally assign:

```rust
conduit_option = conduit;
```

lightning/src/ln/peers/conduit.rs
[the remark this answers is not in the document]
jkczyz: How about simply "encryption.rs"? Or you could break them into "encryptor.rs" and "decryptor.rs" with key rotation living elsewhere.

lightning/src/ln/peers/handshake/mod.rs
jkczyz: What are the acts and the remote public key in this analogy?

lightning/src/ln/peers/handshake/mod.rs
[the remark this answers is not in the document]
jkczyz: `remote_public_key` is passed as a parameter to `process_act` and it is only ever passed `None` in peer_handler.rs. Shouldn't you need to pass it `peer.their_node_id` instead?

lightning/src/ln/peers/handshake/acts.rs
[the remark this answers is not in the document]
jkczyz: Given these are small and either are implementation details or form the public interface of handshake/mod.rs, I'd prefer if they were all in the same file. Otherwise, navigation between files is a bit excessive, IMHO.

lightning/src/ln/peers/handshake/mod.rs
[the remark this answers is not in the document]
jkczyz: Could you also use it here? :)

lightning/src/ln/peers/conduit.rs
jkczyz: Not done?

lightning/src/ln/peers/conduit.rs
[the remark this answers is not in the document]
arik-so: Maybe. But I'd reserve it for a separate PR

lightning/src/ln/peers/handshake/mod.rs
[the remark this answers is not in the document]
arik-so: Normally, yes, but because new_outbound needs the response immediately, the initiate() method is called right there. It is the only act transition that needs the remote pubkey. I think once peer_handler is refactored, we will make new_inbound and new_outbound have more similar calling mechanisms, without one having a special condition for returning data.

arik-so:
Travis seems to have forgotten about this

lightning/src/ln/peer_handler.rs
```diff
@@ -450,317 +462,277 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
+									}
+								};
+							}
+						}
+
+						macro_rules! insert_node_id {
+							() => {
+								match peers.node_id_to_descriptor.entry(peer.their_node_id.unwrap()) {
+									hash_map::Entry::Occupied(_) => {
+										log_trace!(self, "Got second connection with {}, closing", log_pubkey!(peer.their_node_id.unwrap()));
```
ariard: b71b7ea

nit: precise you're closing both connections in the log_trace

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+					return Err("need at least 50 bytes".to_string());
+				}
+
+				let mut act_two_buffer = [0u8; 50];
+				act_two_buffer.copy_from_slice(&self.read_buffer[..act_length]);
+				self.read_buffer.drain(..act_length);
+
+				let (act_three, mut conduit) = self.process_act_two(ActTwo(act_two_buffer))?;
+
+				if self.read_buffer.len() > 0 { // have we received more data still?
```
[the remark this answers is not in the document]
arik-so: I hear you, but haven't really figured out a nice alternative yet.

lightning/src/ln/peer_handler.rs
arik-so: The log entry is just getting moved; changing it will also require changing the no breakage unit test, which I strongly wish to avoid in such a massive refactor.

lightning/src/ln/peers/handshake/mod.rs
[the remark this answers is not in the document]
ariard: Nevermind, I figure out after pursing review than I agree it's not that much easy

lightning/src/ln/peers/handshake/mod.rs
jkczyz: Discussed a bit offline. One idea is to have `PeerHandshake` initialized with the optional remote public key, so that it is not both an input and an output of `process_act`. And then have an accessor for accessing it. Similarly with the conduit.

Then the only return value would be the act data which could even be the `Act` enum instead given it simply wraps the vector, which would make the interface more explicit.

lightning/src/ln/peer_handler.rs
```diff
@@ -1115,19 +1115,22 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
+				}
 
+				if needs_to_write_data {
 					let mut descriptor_clone = descriptor.clone();
 					self.do_attempt_write_data(&mut descriptor_clone, peer);
-
-					peer.awaiting_pong = true;
 				}
 
+				peer.awaiting_pong = true;
```
jkczyz: This seems to obfuscate the code. We don't need to await a pong until a ping is sent, so this could simply be:

```rust
if let PeerState::Connected(ref mut conduit) = peer.encryptor {
	let ping = msgs::Ping {
		ponglen: 0,
		byteslen: 64,
	};
	peer.pending_outbound_buffer.push_back(conduit.encrypt(&encode_msg!(&ping)));
	peer.awaiting_pong = true
	self.do_attempt_write_data(&mut descriptor.clone(), peer);
}
```

Arguably, the last few lines at some point may be refactored to simply:

```rust
peer.send_message(ping);
```

arik-so: alas, Rust 1.22.0 does not allow such simplicity. That's what it was originally.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,237 @@
...
+
+const MESSAGE_LENGTH_HEADER_SIZE: usize = 2;
+const TAGGED_MESSAGE_LENGTH_HEADER_SIZE: usize = MESSAGE_LENGTH_HEADER_SIZE + chacha::TAG_SIZE;
+
+const KEY_ROTATION_INDEX: u32 = 1000;
+
+/// Returned after a successful handshake to encrypt and decrypt communication with peer nodes.
+/// It should not normally be manually instantiated.
+/// Automatically handles key rotation.
+/// For decryption, it is recommended to call `decrypt_message_stream` for automatic buffering.
```
jkczyz: This documentation is out-of-date.

lightning/src/ln/peers/handshake/acts.rs
```diff
@@ -0,0 +1,42 @@
+pub const ACT_ONE_LENGTH: usize = 50;
+pub const ACT_TWO_LENGTH: usize = 50;
+pub const ACT_THREE_LENGTH: usize = 66;
```
jkczyz: There are still places in mod.rs where these aren't used.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+	/// Process act dynamically
+	/// # Arguments
+	/// `input`: Byte slice received from peer as part of the handshake protocol
+	/// `remote_public_key`: If outbound, the peer's static identity public key needs is required for the handshake initiation
+	///
+	/// # Return values
+	/// Returns a tuple with the following components:
+	/// `.0`: Byte vector containing the next act to send back to the peer per the handshake protocol
+	/// `.1`: Conduit option if the handshake was just processed to completion and messages can now be encrypted and decrypted
+	/// `.2`: Public key option if the handshake was inbound and the peer's static identity pubkey was just learned
```
jkczyz: This documentation is out-of-date.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+		let mut hash = HandshakeHash::new(initial_hash_preimage.as_slice());
+		hash.update(&public_key.serialize());
+
+		(hash, chaining_key)
+	}
+
+	/// Process act dynamically
+	/// # Arguments
+	/// `input`: Byte slice received from peer as part of the handshake protocol
+	/// `remote_public_key`: If outbound, the peer's static identity public key needs is required for the handshake initiation
```
jkczyz: This documentation is out-of-date

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+		let public_key = Self::private_key_to_public_key(&private_key);
+		let (hash, chaining_key) = Self::initialize_state(&public_key);
+		handshake.state = Some(HandshakeState::AwaitingActOne(ActOneExpectation {
+			hash,
+			chaining_key,
+		}));
+		handshake
+	}
+
+	/// Getter accessor for the remote public key. It is useful for inbound connections to obtain
```
jkczyz: "Getter accessor" is redundant and unnecessary. Simply state as:

> Return the remote public key once it has been extracted from the third act. Useful for inbound connections.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+		Ok(ActOne(act_one))
+	}
+
+	/// Process a peer's incoming first act and return the second act
+	pub(crate) fn process_act_one(&mut self, act: ActOne) -> Result<ActTwo, String> {
+		let state = self.state.take();
+		let act_one_expectation = match state {
+			Some(HandshakeState::AwaitingActOne(act_state)) => act_state,
+			Some(HandshakeState::Uninitiated) => {
+				// this can also be initiated from a blank state
```
jkczyz: This comment can be left out IMHO. I think it will confuse the casual reader more than anything. "blank" is referring to the old name of "HandshakeState::Uninitiated" so without that context seems like this comment is referring to `private_key_to_public_key`.

lightning/src/ln/peers/handshake/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+				let public_key = Self::private_key_to_public_key(&self.private_key);
+				let (hash, chaining_key) = Self::initialize_state(&public_key);
+				ActOneExpectation {
+					hash,
+					chaining_key,
+				}
+			}
+			_ => {
+				self.state = state;
+				return Err("unexpected state".to_string());
```
jkczyz: Can we just panic in cases where the code is called incorrectly? Same in the other methods.

lightning/src/ln/peers/handshake/acts.rs
arik-so: In places where it's not a concrete act, but either 1 or 2, they shouldn't be used. But I found one spot to replace

lightning/src/ln/peers/handshake/mod.rs
arik-so: Yeah, but I'd rather do that in the next PR. It's always easier to remove a Result wrapper than to add one in.

lightning/src/ln/peer_handler.rs
```diff
@@ -596,33 +591,32 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
 										log_trace!(self, "Received message of type {} from {}", message.type_id(), log_pubkey!(peer.their_node_id.unwrap()));
 
 										// Need an Init as first message
-										if let wire::Message::Init(_) = message {
```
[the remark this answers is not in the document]
arik-so: I'm sorry, I can't see these changes anymore due to squashing. Was it an indentation change?

lightning/src/ln/peers/handshake/acts.rs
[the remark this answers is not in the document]
arik-so: I would prefer to not stuff the handshake logic with all the accessory structs. I would be open to having one separate file where both the enums/states and the acts would live?

lightning/src/ln/peers/handshake/mod.rs
jkczyz: I'm saying we should _never_ return a `Result` in these situations and should _always_ panic instead. So if that argument is valid, we shouldn't merge something that we never want to do.

lightning/src/ln/peers/handshake/acts.rs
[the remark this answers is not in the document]
arik-so: For what it's worth, so does the Act enum

lightning/src/ln/peers/chacha.rs
```diff
@@ -0,0 +1,40 @@
+use util::byte_utils;
+use util::chacha20poly1305rfc::ChaCha20Poly1305RFC;
+
+pub const TAG_SIZE: usize = 16;
```
jkczyz: Seems like there's a few places where this can be used throughout the file.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,237 @@
...
+		(connected_peer, remote_peer)
+	}
+
+	#[test]
+	fn test_empty_message() {
+		let (mut connected_peer, mut remote_peer) = setup_peers();
+
+		let message: Vec<u8> = vec![];
+		let encrypted_message = connected_peer.encrypt(&message);
+		assert_eq!(encrypted_message.len(), 2 + 16 + 16);
```
jkczyz: Could you use the constants that you've defined for these?

lightning/src/ln/peer_handler.rs
```diff
@@ -262,24 +276,22 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
 			their_features: None,
 
 			pending_outbound_buffer: LinkedList::new(),
 			pending_outbound_buffer_first_msg_offset: 0,
 			awaiting_write_event: false,
 
-			pending_read_buffer: pending_read_buffer,
-			pending_read_buffer_pos: 0,
-			pending_read_is_header: false,
+			pending_read_buffer: Vec::new(),
```
TheBlueMatt: If we're gonna have a buffer in the Conduit (which makes sense - it knows how much to read because it knows how to decrypt), can we not just drop it from Peers here?

lightning/src/ln/peer_handler.rs
```diff
@@ -262,24 +276,22 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
 	/// Panics if descriptor is duplicative with some other descriptor which has not yet has a
 	/// disconnect_event.
 	pub fn new_outbound_connection(&self, their_node_id: PublicKey, descriptor: Descriptor) -> Result<Vec<u8>, PeerHandleError> {
-		let mut peer_encryptor = PeerChannelEncryptor::new_outbound(their_node_id.clone(), self.get_ephemeral_key());
-		let res = peer_encryptor.get_act_one().to_vec();
-		let pending_read_buffer = [0; 50].to_vec(); // Noise act two is 50 bytes
+		let mut handshake = PeerHandshake::new_outbound(&self.our_node_secret, &their_node_id, &self.get_ephemeral_key());
+		let (act, ..) = handshake.process_act(&[]).unwrap();
```
TheBlueMatt: This feels super awkward. Why not just have new_outbound return the first act data and return that?

lightning/src/ln/peer_handler.rs
```diff
@@ -887,21 +891,33 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
 								//TODO: Do whatever we're gonna do for handling dropped messages
 							});
 						for msg in update_add_htlcs {
-							peer.pending_outbound_buffer.push_back(peer.channel_encryptor.encrypt_message(&encode_msg!(msg)));
+							if let PeerState::Connected(ref mut conduit) = peer.encryptor {
```
TheBlueMatt: You should be able to just do this once instead of inside each loop.

lightning/src/ln/peer_handler.rs
TheBlueMatt: I tend to agree the redundancy of all these calls suck. Happy for y'all to hash out which PR it gets fixed in, though.

lightning/src/ln/peer_handler.rs
```diff
@@ -501,69 +545,17 @@ impl<Descriptor: SocketDescriptor, CM: Deref> PeerManager<Descriptor, CM> where
...
-										peer.pending_read_is_header = true;
+							let message_option = conduit.decrypt_single_message(Some(&peer.pending_read_buffer.clone()));
+							peer.pending_read_buffer = Vec::new(); // empty the pending read buffer
+							let msg_data = if let Some(message) = message_option {
+								message
+							} else {
+								break;
+							};
 
+							{
```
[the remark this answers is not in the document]
jkczyz: I requested this to prevent a slew of merge conflicts and to keep the blame layer sane. Eventually, we'll remove this as we refactor the code.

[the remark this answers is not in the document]
arik-so: Yeah, let's please not keep messing with the indentation until _after_ we merge this. This will be a very trivial two-line commit.

lightning/src/ln/peer_handler.rs
arik-so: Yeah, but not quite yet. The very core part of peer_handler's logic depends on `while peer.pending_read_buffer.len() > 0`. That is not something where we should have visibility into other objects, so I think we can come up with a much better approach than simply checking if the conduit's read buffer is empty.

lightning/src/ln/peer_handler.rs
arik-so: a constructor should return an instance. Returning a tuple of an instance and the first act would be more confusing to readers of our code.

dr-orlovsky:
I've described some practical use of the changes introduced by PR: https://github.com/rust-bitcoin/rust-lightning/pull/574

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,237 @@
...
+	/// returned in lieu of anything new, even if new data is provided.
+	pub fn decrypt_single_message(&mut self, new_data: Option<&[u8]>) -> Option<Vec<u8>> {
+		let mut read_buffer = if let Some(buffer) = self.read_buffer.take() {
+			buffer
+		} else {
+			Vec::new()
+		};
+
+		if let Some(data) = new_data {
+			read_buffer.extend_from_slice(data);
```
TheBlueMatt, line 64: Hmm, this is kinda painful. Can we avoid copying the new data if we don't need to store it here? Maybe by returning an iterator over new messages that holds a `&mut self` to to this Conduit we can structure it so that it reads from the existing data first, then from the new data, then when the Iterator is dropped pushes the still-pending data into the buffer.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,237 @@
...
+
+	/// Decrypt a message from the beginning of the provided buffer. Returns the consumed number of bytes.
+	fn decrypt(&mut self, buffer: &[u8]) -> (Option<Vec<u8>>, usize) {
+		if buffer.len() < TAGGED_MESSAGE_LENGTH_HEADER_SIZE {
+			// A message must be at least 18 bytes (2 for encrypted length, 16 for the tag)
+			return (None, 0);
+		}
+
+		let encrypted_length = &buffer[0..TAGGED_MESSAGE_LENGTH_HEADER_SIZE];
+		let mut length_bytes = [0u8; MESSAGE_LENGTH_HEADER_SIZE];
```
TheBlueMatt, line 81: It feels kinda awkward not to cache this result somehow - like, if we know we're gonna need 65KB of data to read the full message, decrypting the message header every time we get a new packet of 1.5KB seems super wasteful.

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,297 @@
...
+		self.decryptor.decrypt(buffer)
+	}
+
+	fn increment_sending_nonce(&mut self) {
+		self.encryptor.increment_nonce()
+	}
+
+	fn increment_receiving_nonce(&mut self) {
+		self.decryptor.increment_nonce()
+	}
```
dr-orlovsky, line 93: This private functions are not used from anywhere – and I can't imagine where they may be needed
```suggestion
```

lightning/src/ln/peers/conduit.rs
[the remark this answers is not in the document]
dr-orlovsky: I propose to use name `Transcoder`

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,297 @@
...
+const TAGGED_MESSAGE_LENGTH_HEADER_SIZE: usize = MESSAGE_LENGTH_HEADER_SIZE + chacha::TAG_SIZE;
+
+const KEY_ROTATION_INDEX: u32 = 1000;
+
+/// Returned after a successful handshake to encrypt and decrypt communication with peer nodes.
+/// It should not normally be manually instantiated.
+/// Automatically handles key rotation.
+/// For decryption, it is recommended to call `decrypt_message_stream` for automatic buffering.
+pub struct Conduit {
+	pub(super) encryptor: Encryptor,
```
dr-orlovsky, line 18: In my rust-lightning node I handle input and output parts of TCP socket in different threads; so it will be nice to have an ability to split encryptor and decryptor as well. So I propose to make those fields public — or add `split` and `join` functions (see my proposed changes below)

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,297 @@
...
+			decryptor: Decryptor {
+				receiving_key,
+				receiving_chaining_key: chaining_key,
+				receiving_nonce: 0,
+				read_buffer: None,
+				pending_message_length: None
+			}
+		}
+	}
+
```
dr-orlovsky, line 64:
```suggestion
	pub fn join(encryptor: Encryptor, decryptor: Decryptor) -> Self {
		Self {
			encryptor,
			decryptor
		}
	}
	
	pub fn split(self) -> (Encryptor, Decryptor) {
		(self.encryptor, self.decryptor)
	}
	
```

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,297 @@
...
+/// It should not normally be manually instantiated.
+/// Automatically handles key rotation.
+/// For decryption, it is recommended to call `decrypt_message_stream` for automatic buffering.
+pub struct Conduit {
+	pub(super) encryptor: Encryptor,
+	pub(super) decryptor: Decryptor
+
+}
+
+pub(super) struct Encryptor {
```
dr-orlovsky, line 23:
```suggestion
pub struct Encryptor {
```

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,297 @@
...
+	sending_key: SymmetricKey,
+	sending_chaining_key: SymmetricKey,
+	sending_nonce: u32,
+}
+
+pub(super) struct Decryptor {
```
dr-orlovsky, line 29:
```suggestion
pub struct Decryptor {
```

dr-orlovsky:
FYI, I've done a PR merging the latest main into this branch: https://github.com/arik-so/rust-lightning/pull/1

lightning/src/ln/peers/handler.rs
```diff
@@ -503,145 +557,118 @@ impl<Descriptor: SocketDescriptor, CM: Deref, RM: Deref, L: Deref> PeerManager<D
...
+					let mut send_init_message = false;
+
+					let data_processing_decision = peer.encryptor.process_peer_data(data, &mut peer.pending_outbound_buffer);
+					match data_processing_decision {
+						PeerDataProcessingDecision::Disconnect(e) => {
+							log_trace!(self.logger, "Invalid act message; disconnecting: {}", e);
+							return Err(e);
 						}
 
-						if peer.pending_read_buffer_pos == peer.pending_read_buffer.len() {
```
julianknutsen, line 518: During the handshake `pending_read_buffer` is initialized to an act_length vector filled with 0s. So, until the full act arrives this branch will not be taken. This is a different behavior than this PR changes which return an `Err` if the `process_act()` doesn't receive the full act and results in a `PeerDataProcessingDecision::Disconnect`.

#672

lightning/src/ln/peers/conduit.rs
```diff
@@ -0,0 +1,297 @@
...
+			length
+		} else {
+			if buffer.len() < TAGGED_MESSAGE_LENGTH_HEADER_SIZE {
+				// A message must be at least 18 bytes (2 for encrypted length, 16 for the tag)
+				return (None, 0);
+			}
+
+			let encrypted_length = &buffer[0..TAGGED_MESSAGE_LENGTH_HEADER_SIZE];
+			let mut length_bytes = [0u8; MESSAGE_LENGTH_HEADER_SIZE];
+			length_bytes.copy_from_slice(&chacha::decrypt(&self.receiving_key, self.receiving_nonce as u64, &[0; 0], encrypted_length).unwrap());
```
julianknutsen, line 170: Found this bug while adding to the fuzz testing today. It should return an error here in the event of a decryption error instead of panicking. This is a bit involved since the `Iterator` implementation now needs to return `Result` objects, but it is doable.

ariard:
@julianknutsen @arik-so How do we move forward with this PR and "Enum Dispatch NOISE State Machine" ? 

Should we focus review on this PR first then open a rebased "Enum Dispatch NOISE State Machine" ? Or "Enum Dispatch NOISE State" should supersede it with a new PR ?

arik-so:
I think the nature of the NOISE PR updating the branch that will them update the modular handshake PR automatically requires that we focus on the NOISE one first.

julianknutsen:
> I think the nature of the NOISE PR updating the branch that will then update the modular handshake PR automatically requires that we focus on the NOISE one first.

I agree. I think that the end of this is going to look like a "feature branch" represented by https://github.com/arik-so/rust-lightning/tree/modular_handshake and a set of PRs against that branch that address some of the outstanding review comments as well as this handshake state machine to get it into a point where the modular handshake PR is ready to go.

There are still a few outstanding action items from this original PR that I am addressing in future PRs targetting ariks's branch so it seems like vetting code into that branch and then a final PR from his branch into master is the path of least resistance. Still going to be quite a bit of work, but that is just the nature of long-standing dependent PRs.

If there is additional legwork you would like me to do (rebasing, merge PRs, etc)  to make it easy just let me know. I know that this wasn't planned work so I want to make it as easy as possible to integrate the changes and additional testing.

ariard:
Okay I'm even on this, will review the NOISE PR against Arik's repo so. At least as we track back review context.

julianknutsen:
Superseded by PR stack ending at https://github.com/rust-bitcoin/rust-lightning/pull/694

ariard:
Closing, as it's superseded.

PR #599, opened 2020-04-20
Fix build warnings

D4nte:
I have noticed few build warnings on master and I thought I'd give a go at suppressing them. So far I have not changed any software behaviour, just removed unused parentheses and removed the warning using an `allow` attribute.

I have justified the used of the `allow` attribute in the commit.
Do you prefer justifying the use of `allow` attribute with in code comments?

Also, I assume that we are keeping the used of `description()` in purpose. Let me know if you'd prefer to see an implementation and use of the `Display` trait instead (as per Rust's recommendation).

I have not removed the warning `warning: field is never read: 'logger'` for `ChainWatchInterfaceUtil` and `KeysManager` because indeed, it is never read and I cannot seem to grasp why this field is present.

Some guidance would be greatly appreciated 🙂


TheBlueMatt:
This seems duplicative with #594. Reviews there would be appreciated :).

TheBlueMatt:
#594 was merged, so gonna close this. As for the missing logger reads, those were added cause we figured we'd have something worth logging. We probably do, in fact, have things worth logging, we just never do. I think the right fix is to log *something* at least at the trace level.

Issue #605, opened 2020-04-23
Experimenting with DLC-style of contract on Rust-Lightning


ariard:
@nkohen and Yancy Ribbens are interested to experiment offchain DLC. RL suits very well for this, I think ideally DLC extension should be downstream as another under rust-bitcoin org repository to make progress as its own rhythm but right now opening this issue to track what we should do.

* what custom message ?
* how to avoid weaking commitment transaction ?
* what design of 2nd-stage transaction ?
* maintaining a rust-secp256k-zkp in rust-bitcoin with ecdsa_adaptor module ?

@nkohen I'm willingly to start the rust-dlc skeleton compatible with Rust-Lighning and let you fulfill holes from then to learn Rust but I need a better description of the protocol do you have in one of your blog or in dlcspecs ?

cc @arik-so (you should be definitely interested, DLC where my main reason to experiment with PTLC a while ago :) )

Christewart:
If anyone is interested in organizing a "formal" effort for this, it appears Fulmo is having another hackday over the weekend of May 9th-11th. It was pretty useful for us all to collaborate on the first version of PTLCs over that weekend and get something working. 

>@channel Save the date! The next #LightningHacksprint is coming up on the weekend of May 9th/10th 2020.  Please use https://wiki.fulmo.org to update your challenges, bounties and self-organized sessions. Ping me or @rootzoll is you have any questions.

LLFourn:
@Christewart I'll join again :). Another approach we could take would just to implement a normal PTLC payment without randomization at each hop. i.e. just making the one-way function for the secret a point multiplication rather than a hash. This seems like a pre-requisite to doing DLCs anyway. Here's what i think are the most minimal script changes to do a PTLC:

## Offered HTLC
### Before
```
OP_DUP OP_HASH160 <RIPEMD160(SHA256(revocationpubkey))> OP_EQUAL
OP_IF
    OP_CHECKSIG
OP_ELSE
    <remote_htlcpubkey> OP_SWAP OP_SIZE 32 OP_EQUAL
    OP_NOTIF
        # To local node via HTLC-timeout transaction (timelocked).
        OP_DROP 2 OP_SWAP <local_htlcpubkey> 2 OP_CHECKMULTISIG
    OP_ELSE
        # To remote node with preimage.
        OP_HASH160 <RIPEMD160(payment_hash)> OP_EQUALVERIFY
        OP_CHECKSIG
    OP_ENDIF
OP_ENDIF
[...]

PR #619, opened 2020-05-11
[Experimental] Generic Commitment Transaction Output - WIP DLC support

ariard:
This is a starter work to support generic payload on the commitment transaction. It mostly switches InboundHTLCOutput/OutboundHTLCOutput behind InboundGenericOutput/OutboundGenericOutput traits. This is a headstart on #605 

Adding Box<InboundGenericOutput> in Channel on its performance decrease is yet to be confirmed as an approach but other approach explored were really counter-intuititive, specially for an implementator.

Thanks to @nkohen there is also new DLC messages and a DLC-compatible state machine flow in test_send_dlc. 

Next steps:
* support `update_countersign_dlc`
* support `update_fulfill_dlc`
* add `ChannelManager::send_dlc`
* abstract some transactions builders such as `Channel::build_commitment_transaction` or `Channel::build_htlc_transaction`


devrandom:
Is the longer term goal to allow this to be something that can be provided by the developer rather than being in the RL crate?

TheBlueMatt:
Yes, ideally a developer could specify the exact scripts and write the protocol using custom p2p messages.

TheBlueMatt:
This is pretty long out-of-date. Gonna go ahead and close but can obviously be re-opened if it gets picked back up again.

PR #642, opened 2020-06-19
Anchor-outputs (1/3): Refactoring chain backend to extract PackageTemplate

ariard:
_This is 1/3 of anchor ouput implementation_

I invite reviewers to only Concept ACK/Approach ACK until I open the complete serie of patchset to see if implementation makes sense.

### Rational

Anchor ouptut proposal is a spec move towards introducing dynamic fee-setting for LN transactions. Actually, feerate is selected by channel opener through `update_fee` and must be adapted to follow network-wise mempool increases/decreases. This model is broken with regards to quick fee spikes, adversial counterparty exploiting mempool policy rules or `update_fee` in itself introducing funny attack vectors ([Flood & Loot](https://arxiv.org/abs/2006.08513) among others). 

Anchor output modifies commitment transaction format by introducing a special per-party output for adjusting feerate through CPFP. Note, without base layer changes this is still insecure and can only be relied for favorable mempool congestion in case of intentional unilateral closure.

Implementing anchor output consists in both onchain changes to support CPFP logic and offchain ones to dynamically add this new anchor output. This CPFP logic and the further onchain backend will likely get more complex in the future both by security and fee efficiency (like "aggressive" aggregation, cross-commitment CPFP, don't-go-onchain-if-your-HTLC-is-non-economic, don't-bump-more-than-HTLC-in-danger, ...). So this PR envision this by removing code complexity from `OnchainTxHandler` to some new interface `OnchainRequest`/`PackageTemplate` mimicking "packets". 

Motivations behind `OnchainRequest`/`PackageTemplate` is to provide a generic interface for `OnchainTxHandler` without this component being aware of actual transaction type processed. It's nice to introduce later pre-signed justice transactions for watchtowers or anchor output garbage-collecting transactions.

Interface holds high-level methods like `package_weight` or `package_amounts`. Concretely it kills the monster `OnchainTxHandler::generate_claim_tx` to get it ready for CPFP support.  Current interface is an enum, we may want to switch to trait object but it comes with trade-offs in Rust (thoughts ?)

### PR scope

This PR SHOULD NOT introduce any behavior change.

Overview of PR:
* move `InputDescriptors` in new `onchain_utils`
* replace `InputDescriptors` by `PackageTemplate`, switching from per-input claim to holistic package make it easier to compute weight, which is less footgunish
* replace `ClaimTxBumpMaterial`/`ClaimRequest` by a unique structure, easing aggregatiion
* extract fee computing utilitties from `OnchainTxHandler


lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+				}
+				None
+			},
+			_ => {
+				// Note, we may try to split on remote transaction for
+				// which we don't have a competing one (HTLC-Success before
+				// timelock expiration). This explain we don't panic!.
+				// We should refactor OnchainTxHandler::block_connected to
+				// only test equality on competing claims.
+				return None;
```
arik-so, line 329: I think `None` would do the trick, too

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+			}
+		}
+	}
+	pub(crate) fn package_merge(&mut self, mut template: PackageTemplate) {
```
arik-so, line 333: Same as `package_split`.

lightning/src/ln/onchaintx.rs
```diff
@@ -462,235 +293,71 @@ impl<ChanSigner: ChannelKeys> OnchainTxHandler<ChanSigner> {
...
+		let mut preprocessed_requests = Vec::with_capacity(requests.len());
+		let mut aggregated_request = OnchainRequest::default();
 
 		// Try to aggregate outputs if their timelock expiration isn't imminent (absolute_timelock
 		// <= CLTV_SHARED_CLAIM_BUFFER) and they don't require an immediate nLockTime (aggregable).
-		for req in claimable_outpoints {
+		for req in requests {
 			// Don't claim a outpoint twice that would be bad for privacy and may uselessly lock a CPFP input for a while
-			if let Some(_) = self.claimable_outpoints.get(&req.outpoint) { log_trace!(logger, "Bouncing off outpoint {}:{}, already registered its claiming request", req.outpoint.txid, req.outpoint.vout); } else {
+			if let Some(_) = self.claimable_outpoints.get(req.content.outpoints()[0]) { log_trace!(logger, "Bouncing off outpoint {}:{}, already registered its claiming request", req.content.outpoints()[0].txid, req.content.outpoints()[0].vout); } else {
```
arik-so, line 337: you keep accessing the first outpoint, so I'd just put it in a variable for easier legibility

devrandom (commented):
Here are my initial thoughts on this PR

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+/// An enum to describe a claim content which is generated by ChannelMonitor and
+/// used by OnchainTxHandler to regenerate feerate-bump transactions to settle claims.
+///
+/// Template may be either malleable (a justice tx, a counterparty HTLC tx) or lockdown (a holder htlc
+/// tx, a holder commitment tx, a pre-signed justice tx). Bumping can be a Replace-by-Fee, that way
+/// the claim-settlement tx in itself has its feerate increased or Child-Pay-For-Parent, a child
+/// of the claim tx has its feerate increased. For the latter case, access to the whole package
+/// sizea and pre-committed fee is required to compute an efficient bump.
+#[derive(Clone, PartialEq)]
+pub(crate) enum PackageTemplate {
```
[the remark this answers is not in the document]
ariard: Good point, I think we can aggregate MalleableJusticeTx/CounterpartyHTLCTx on a lot of cases. Though I'm holding this on the discussion to switch to dynamic trait objects instead of an enum pseudo-interface.

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+				return outpoints;
+			},
+			PackageTemplate::HolderCommitmentTx { ref input } => {
+				let mut outpoints = Vec::with_capacity(1);
+				outpoints.push(&input.0);
+				return outpoints;
+			},
+		}
+	}
+	pub(crate) fn package_split(&mut self, outp: &BitcoinOutPoint) -> Option<PackageTemplate> {
```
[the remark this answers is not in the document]
ariard: Good, done for `package_finalize` and `package_merge` too. Not for `package_amounts` it's an attribute?

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+							base_inputs.insert(k, v);
+						}
+					},
+					_ => panic!("Merging templates of different types")
+				}
+			},
+			_ => panic!("Merging template on non-malleable packages")
+		}
+	}
+	pub(crate) fn package_amounts(&self) -> u64 {
```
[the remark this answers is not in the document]
ariard: Yes done.

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+			lock_time: 0,
+			input,
+			output: vec![TxOut {
+				script_pubkey: destination_script.clone(),
+				value: 0
+			}],
+		};
+		bumped_tx.get_weight() + witnesses_weight
+	}
+	pub(crate) fn package_finalize<L: Deref, ChanSigner: ChannelKeys>(&self, onchain_handler: &mut OnchainTxHandler<ChanSigner>, value: u64, destination_script: Script, logger: &L) -> Option<Transaction>
```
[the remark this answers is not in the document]
ariard: Yes in the future I think we can just pass `key_signer` as we're just interested with signer for construction and signing. Maybe we can remove entirely signer from `OnchainTxHandler` and just give it to `PackageTemplate` constructor ?

Note, you still have `get_fully_signed_htlc_tx`/`get_fully_signed_holder_tx` blocking to remove the passing of `OnchainTxHandler`, it's a leftover from the past which is my bad. `OnchainTxHandler` shouldn't have any commitment transaction state directly exposed and it should be wrapped inside package. I started to do this move while writing this PR but it was already a big refactor...

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+		};
+		Ok(bump_strategy)
+	}
+}
+
+/// A structure to describe a claim content and its metadatas which is generated
+/// by ChannelMonitor and used by OnchainTxHandler to generate feerate-competive
+/// transactions.
+///
+/// Metadatas are related to multiple fields playing a role in packet lifetime.
```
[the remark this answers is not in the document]
ariard: Good point!

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+	let new_fee = if previous_feerate < fee_estimator.get_est_sat_per_1000_weight(ConfirmationTarget::HighPriority) as u64 {
+		if let Some((new_fee, _)) = subtract_high_prio_fee(input_amounts, predicted_weight, fee_estimator, logger) {
+			new_fee
+		} else {
+			log_trace!(logger, "Can't new-estimation bump new claiming tx, amount {} is too small", input_amounts);
+			return None;
+		}
+	// ...else just increase the previous feerate by 25% (because that's a nice number)
+	} else {
+		let fee = previous_feerate * (predicted_weight as u64) / 750;
```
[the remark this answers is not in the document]
ariard: I think you're right :

* (feerate=1000) * (weight=100) / (weight_denominator=1000) = 100
* (feerate=1000) * (weight=100) / (weight_denominator=750) = 133

So good catch! I need to update tests so not marking at resolved yet. If you have a better idea on the % heuristic lmk it's quite arbitrary rn :)

ariard:
Thanks for review @devrandom, updated with at least the minor fixed.

I think before to go further with this PR review there is a discussion between using enum as a pseudo-interface or moving to trait objects. The downside with trait objects (experimented a bit for offchain generic output in #619) is you need a Box and a memory over-head. It might be okay for onchain stuff which isn't a hot path.

Note, I see this PR as a real blocker to work further on third-party watchtower support, future PackageTemplate should be DelegatedJusticeTx/HolderDLCTx. Once we resolve the code design point above, I would be glad to land this PR as soon as we can.

devrandom:
Regarding efficiency - I usually don't try to optimize this early in a project's lifetime.  This especially applies to crypto code, where sign/encrypt/verify operations are usually the bulk of the CPU load.

lightning/src/ln/onchain_utils.rs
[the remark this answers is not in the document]
devrandom: yeah, sounds like a good topic for a separate PR

lightning/src/ln/onchain_utils.rs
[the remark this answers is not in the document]
devrandom: well, to be obvious, this has to be high enough so it tracks feerate spikes within the expiration time, especially when there's an attack on the mempool.  if I understand the code correctly, a bump happens on every block, right?

so, the desired bump % is dependent on the expiration time in blocks and the ratio of the spike to original feerate.  one issue is if the original feerate was very low, let's say because the network was quiet.  then perhaps we can get a very high spike ratio. 

for example, with a 100x spike height and 144 block expiration, the bump should be around 3.2%.  we could handle much higher spikes with a 10% increase - up to a `(1.1 ^ 144) = 900000` feerate increase over 144 blocks.

we should have this in a const with some documentation on the const justifying the value based on assumptions on maximum spike height and minimum expiration time.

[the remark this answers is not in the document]
ariard: Bump happens accordingly to `OnchainTxHandler::get_height_timer`, if expiration is coming in next 3 blocks, bumps at every block.

What do you qualify as spike height here ? The original height at which the transaction is broadcast for first time ? Ideally you want the fee-bump to be exponential, save on fees if timelock is far away, be ready to burn as funds at stake when the expiration is really near.

Agree on documenting our assumptions.

ariard:
Updated at de93efc.

Currently, this is the main area where I'm curious of feedback. If we prefer trait objects, I'll need to re-work this PR, so first I would like to settle on trait objecs or enum as a abstract package interface.

> I think before to go further with this PR review there is a discussion between using enum as a pseudo-interface or moving to trait objects. The downside with trait objects (experimented a bit for offchain generic output in #619) is you need a Box and a memory over-head. It might be okay for onchain stuff which isn't a hot path.

lightning/src/ln/onchain_utils.rs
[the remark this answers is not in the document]
devrandom: apologies, missed this reply.  I should have said "spike factor" - which I meant as the ratio of the feerate when we start to try to publish and the feerate at the peak of the spike.

I agree that the bumps should be exponential (and they are in your code, right?).

But my assumption about the bump frequency was incorrect.  Looking at `get_height_timer`, if we have 144 blocks, we bump 9 times every 15 blocks, then twice every 3 blocks then another 3 times, for a total of 14 or so.  I might be slightly off.  so `1.25 ^ 14` = 22.

I'm not sure that's actually enough to be safe, since we do stand to lose the funds in the worst case.  I think a factor of 100 would be better.

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+		let funding_outp = HolderFundingOutput {
+			funding_redeemscript,
+		};
+		PackageTemplate::HolderCommitmentTx {
+			input: (BitcoinOutPoint { txid, vout }, funding_outp)
+		}
+	}
+}
+
+impl Default for PackageTemplate {
```
TheBlueMatt, line 568: Wait, why is this needed?

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+	// Timeout tx must have nLocktime set which means aggregating multiple
+	// ones must take the higher nLocktime among them to satisfy all of them.
+	// Sadly it has few pitfalls, a) it takes longuer to get fund back b) CLTV_DELTA
+	// of a sooner-HTLC could be swallowed by the highest nLocktime of the HTLC set.
+	// Do simplify we mark them as non-aggregable.
+	pub(crate) aggregation: bool,
+	// Content may lockdown with counter-signature of our counterparty
+	// or fully-malleable by our own. Depending on this bumping strategy
+	// must be adapted.
+	pub(crate) bump_strategy: BumpStrategy,
```
TheBlueMatt, line 722: Isn't this determined by the actual output itself? Shouldn't this be in the PackageTemplate?

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
...
+	// Based feerate of previous broadcast. If resources available (either
+	// output value or utxo bumping).
+	pub(crate) feerate_previous: u64,
+	// At every block tick, used to check if pending claiming tx is taking too
+	// much time for confirmation and we need to bump it.
+	pub(crate) height_timer: Option<u32>,
+	// Block height before which claiming is exclusive to one party,
+	// after reaching it, claiming may be contentious.
+	pub(crate) absolute_timelock: u32,
```
TheBlueMatt, line 731: Same here, shouldn't this be a function on PackageTemplate?

TheBlueMatt (commented):
Overall looks like a good refactor. Have a couple of API questions.

TheBlueMatt:
> Regarding efficiency - I usually don't try to optimize this early in a project's lifetime. This especially applies to crypto code, where sign/encrypt/verify operations are usually the bulk of the CPU load.

This I agree with for CPU operations, but memory allocation operations are more tricky. My personal experience on Core made me very, very scared of too much allocation, its very easy to have a program that ends up with 3x RSS as is required from memory fragmentation and other overhead, and not really have an easy way to fix it - trying to reduce memory allocations after the fact can be a very challenging rebase, doubly so because there aren't any good tools for debugging or profiling these issues.

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,640 @@
...
+		Ok(HolderFundingOutput {
+			funding_redeemscript: Readable::read(reader)?,
+		})
+	}
+}
+
+/// An enum to describe a claim content which is generated by ChannelMonitor and
+/// used by OnchainTxHandler to regenerate feerate-bump transactions to settle claims.
+///
+/// Template may be either malleable (a justice tx, a counterparty HTLC tx) or lockdown (a holder htlc
```
devrandom, line 259: Why is "counterparty HTLC tx" malleable, but "holder HTLC tx" not?  is the term "HTLC tx" used differently in the two cases?

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,892 @@
+//! Utilities for computing witnesses weight and feerate computation for onchain operation
```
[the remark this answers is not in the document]
ariard: Yeah, moved. I guess `chain/` wasn't a thing when I started to work on this branch :)

lightning/src/ln/onchain_utils.rs
ariard: Dropped, and used an Option instead. Needed some base in onchaintx on which to potentially aggregate OnchainRequest.

lightning/src/ln/onchain_utils.rs
ariard: Added a is_malleable() to PackageTemplate, we should get the same effect than BumpStrategy. Yes for now bump_strategy == malleability, but it might differ in the future with watchtowers/channel monitors replicas and where the keys are laying.

lightning/src/ln/onchain_utils.rs
ariard: Yeah, I'll should study if I can merge OnchainRequest with PackageTemplate, now that I've isolated outpoints and their solving materials in PackageSolvingData.

lightning/src/ln/onchain_utils.rs
ariard: Holder HTLC tx only become malleable with anchor changes introduced in other branches of this patchset :)
Though they're only malleable under SIGHASH_SINGLE. Claiming transactions on counterparty HTLC outputs have always been malleable beyond witness satisfaction (i.e settup nLocktime correctly) ?

ariard:
Okay pushed a refactored version of the refactor at 8b2797. It gets rides of code duplication by encapsulating outputs special solving data in a new `PackageSolvingData`. Still needs heavily documentation but will do once we agree on onchain_utils's structs.

ariard:
Rebased at 238dd96. Working on API documentation and test coverage.

lightning/src/chain/channelmonitor.rs
```diff
@@ -1565,10 +1420,10 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
 		// *we* sign a holder commitment transaction, not when e.g. a watchtower broadcasts one of our
 		// holder commitment transactions.
 		if self.broadcasted_holder_revokable_script.is_some() {
-			let (claim_reqs, _) = self.get_broadcasted_holder_claims(&self.current_holder_commitment_tx);
+			let (claim_reqs, _) = self.get_broadcasted_holder_claims(&self.current_holder_commitment_tx, /*XXX(ariard) Option<height> */ 0);
```
TheBlueMatt, line 1423: I assume these weren't meant to slip in :)

lightning/src/chain/onchaintx.rs
```diff
@@ -286,9 +150,9 @@ pub struct OnchainTxHandler<ChannelSigner: Sign> {
 	// us and is immutable until all outpoint of the claimable set are post-anti-reorg-delay solved.
 	// Entry is cache of elements need to generate a bumped claiming transaction (see ClaimTxBumpMaterial)
 	#[cfg(test)] // Used in functional_test to verify sanitization
-	pub pending_claim_requests: HashMap<Txid, ClaimTxBumpMaterial>,
+	pub pending_claim_requests: HashMap<Txid, PackageTemplate>,
```
TheBlueMatt, line 153: rustc doesn't like this - even though the whole module is only `pub(crate)` the fact that this includes a `pub(crate)` thing makes it sad. We can just make this `pub(crate)` instead.

lightning/src/chain/onchain_utils.rs
```diff
@@ -0,0 +1,884 @@
...
+/// transaction or second-stage HTLC transaction.
+///
+/// Amount and revocation secret are used in both cases.
+#[derive(Clone, PartialEq)]
+pub(crate) struct RevokedOutput {
+	per_commitment_point: PublicKey,
+	counterparty_delayed_payment_base_key: PublicKey,
+	counterparty_htlc_base_key: PublicKey,
+	per_commitment_key: SecretKey,
+	input_descriptor: InputDescriptors,
```
TheBlueMatt, line 144: Its a little confusing that a `RevokedOutput` could have type `OfferedHTLC`/`ReceivedHTLC`. Can we split out the type here or assert that the type is a `Revoked`... one in `build` (it looks like the non-revoked types are really just used for `get_witnesses_weight`).

lightning/src/chain/onchain_utils.rs
```diff
@@ -0,0 +1,884 @@
...
+			PackageSolvingData::HolderHTLCOutput(..) => { 0 },
+			PackageSolvingData::HolderFundingOutput(..) => { 0 },
+		};
+		amt
+	}
+	fn weight(&self) -> usize {
+		let weight = match self {
+			PackageSolvingData::RevokedOutput(ref outp) => { get_witnesses_weight(&[outp.input_descriptor]) },
+			PackageSolvingData::CounterpartyHTLCOutput(ref outp) => { get_witnesses_weight(if outp.preimage.is_some() { &[InputDescriptors::OfferedHTLC] } else { &[InputDescriptors::ReceivedHTLC] }) },
+			PackageSolvingData::HolderHTLCOutput(..) => { 0 },
```
TheBlueMatt, line 352: Can you add a comment describing why its ok that these are 0 here, or what the calling preconditions are for this function?

TheBlueMatt (commented):
Note that a number of tests fail, and a few appear to hang.

lightning/src/chain/channelmonitor.rs
ariard: Removed, they're markers for the remaining patchset. For height values passed they don't matter as we can't yet bump them.

lightning/src/chain/onchain_utils.rs
ariard: So I dried-up InputDescriptors and instead I'm using new named constants. It turns out InputDescriptors and the associated helper `get_witnesses_weight` don't serve a real purpose. At detection, spend weights can be marked as static, we won't touch them anymore until transaction generation.

lightning/src/chain/onchain_utils.rs
ariard: Added a comment, values should change with the remaining anchor patchset, introducing fee-bumping of holder outputs spending packages.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,15 +1,48 @@
...
+
+/// witnessScript markers used to compute the expected witness weight.
 #[derive(PartialEq, Clone, Copy)]
 pub(crate) enum InputDescriptors {
 	RevokedOfferedHTLC,
 	RevokedReceivedHTLC,
 	OfferedHTLC,
 	ReceivedHTLC,
-	RevokedOutput, // either a revoked to_local output on commitment tx, a revoked HTLC-Timeout output or a revoked HTLC-Success output
+	RevokedOutput, // either a revoked to_holder output on commitment tx, a revoked HTLC-Timeout output or a revoked HTLC-Success output
```
TheBlueMatt: In the previous commit (while moving this block) you change it from to_holder to to_local, and then change it back here.

lightning/src/ln/onchain_utils.rs
```diff
@@ -88,3 +123,649 @@ pub(crate) fn get_witnesses_weight(inputs: &[InputDescriptors]) -> usize {
...
+			input_descriptor,
+			amount,
+			htlc,
+			on_counterparty_tx_csv
+		}
+	}
+}
+
+impl Writeable for RevokedOutput {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ::std::io::Error> {
```
TheBlueMatt: nit here and below: can we implement this with `impl_writeable!()`?

lightning/src/ln/onchain_utils.rs
```diff
@@ -88,3 +123,649 @@ pub(crate) fn get_witnesses_weight(inputs: &[InputDescriptors]) -> usize {
...
+		writer.write_all(&byte_utils::be64_to_array(self.amount))?;
+		self.htlc.write(writer)?;
+		self.on_counterparty_tx_csv.write(writer)?;
+		Ok(())
+	}
+}
+
+impl Readable for RevokedOutput {
+	fn read<R: ::std::io::Read>(reader: &mut R) -> Result<Self, DecodeError> {
+		let per_commitment_point = Readable::read(reader)?;
```
TheBlueMatt: style nit: If we're reading this in-order we can also put the `Readable::read` calls inside the struct creation below.

lightning/src/ln/onchain_utils.rs
```diff
@@ -88,3 +123,649 @@ pub(crate) fn get_witnesses_weight(inputs: &[InputDescriptors]) -> usize {
...
+			PackageSolvingData::HolderHTLCOutput(..) => { 0 },
+			PackageSolvingData::HolderFundingOutput(..) => { 0 },
+		};
+		amt
+	}
+	fn weight(&self) -> usize {
+		let weight = match self {
+			PackageSolvingData::RevokedOutput(ref outp) => { get_witnesses_weight(&[outp.input_descriptor]) },
+			PackageSolvingData::CounterpartyHTLCOutput(ref outp) => { get_witnesses_weight(if outp.preimage.is_some() { &[InputDescriptors::OfferedHTLC] } else { &[InputDescriptors::ReceivedHTLC]}) },
+			// Note: Currently, weights of holder outputs spending witnesses aren't used
```
TheBlueMatt: If this is unused can we at least have a `debug_assert!(false)` instead of returning a bogus value?

lightning/src/chain/channelmonitor.rs
```diff
@@ -1571,7 +1571,7 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
 			// First, process non-htlc outputs (to_holder & to_counterparty)
 			for (idx, outp) in tx.output.iter().enumerate() {
 				if outp.script_pubkey == revokeable_p2wsh {
-					let revk_outp = RevokedOutput::build(per_commitment_point, per_commitment_key, self.counterparty_tx_cache.counterparty_delayed_payment_base_key, self.counterparty_tx_cache.counterparty_htlc_base_key, InputDescriptors::RevokedOutput, outp.value, None, self.counterparty_tx_cache.on_counterparty_tx_csv);
+					let revk_outp = RevokedOutput::build(per_commitment_point, per_commitment_key, self.counterparty_tx_cache.counterparty_delayed_payment_base_key, self.counterparty_tx_cache.counterparty_htlc_base_key, WEIGHT_REVOKED_OUTPUT, outp.value, None, self.counterparty_tx_cache.on_counterparty_tx_csv);
```
TheBlueMatt, line 1574: This feels like even more of a layer violation than the original `InputDescriptors` enum. I think my original complaint about it was more about how its confusing to have two separate enums (and that you can have a non-`Revoked` `InputDescriptor` inside of a `RevokedOutput` which is obviously nonsensical). Instead, can we make `RevokedOutput` at least contain an enum so that the `htlc` field isn't an option and we don't need to store a weight passed in from another layer and can calculate it locally?

lightning/src/chain/channelmonitor.rs
```diff
@@ -1716,8 +1571,9 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
...
-					claimable_outpoints.push(ClaimRequest { absolute_timelock: height + self.counterparty_tx_cache.on_counterparty_tx_csv as u32, aggregable: true, outpoint: BitcoinOutPoint { txid: commitment_txid, vout: idx as u32 }, witness_data});
+					let revk_outp = RevokedOutput::build(per_commitment_point, per_commitment_key, self.counterparty_tx_cache.counterparty_delayed_payment_base_key, self.counterparty_tx_cache.counterparty_htlc_base_key, WEIGHT_REVOKED_OUTPUT, outp.value, None, self.counterparty_tx_cache.on_counterparty_tx_csv);
+					let justice_package = PackageTemplate::build_package(commitment_txid, idx as u32, PackageSolvingData::RevokedOutput(revk_outp), PackageMalleability::Malleable, height + self.counterparty_tx_cache.on_counterparty_tx_csv as u32, true, 0, None, height);
```
TheBlueMatt, line 1575: Similarly - can we simplify this further and push more logic down into `onchain_utils`? eg even having `channelmonitor` aware of what types of outputs are and are not Malleable seems like split logic - ideally `ChannelMonitor` can *only* know about the transactions its looking at and then tell `onchain_utils` or `onchaintx` which of a given enum an output is and let it handle the rest (maybe with different constructors depending on the types out outputs being spent instead of one big constructor?).

TheBlueMatt (commented):
This is great, but I think we can simplify the logic a bunch more and pull more out of `channelmonitor`. It may be that some of it makes more sense in a followup, but probably it could go in here.

lightning/src/ln/onchain_utils.rs
ariard: Right, corrected in "Add onchain_utils"

lightning/src/ln/onchain_utils.rs
ariard: Yep good idea, done.

lightning/src/ln/onchain_utils.rs
ariard: Amount is exercised in `generate_claim_tx` so added a debug_assert(amt == 0) otherwise added for weight.

lightning/src/chain/channelmonitor.rs
ariard: Good call, this one is super straightforward to address. Instead of splitting constructor, I did embed malleability assignment based on output type. `channelmonitor` is now blind w.r.t to package malleability

lightning/src/chain/channelmonitor.rs
ariard: Addressed with new RevokedHTLCOutput and corresponding constructor

ariard:
Rebased on top of #923, with comments fixed at 59a27af

TheBlueMatt:
Concept ACK, this looks really good. I'll give it a more thorough review ASAP.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,698 @@
...
+impl Writeable for PackageTemplate {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ::std::io::Error> {
+		writer.write_all(&byte_utils::be64_to_array(self.inputs.len() as u64))?;
+		for (ref outpoint, ref rev_outp) in self.inputs.iter() {
+			outpoint.write(writer)?;
+			rev_outp.write(writer)?;
+		}
+		self.malleability.write(writer)?;
+		self.absolute_timelock.write(writer)?;
+		self.aggregation.write(writer)?;
```
jkczyz: Is there a need to serialize `malleability` and `aggregation`? Can't they be inferred from the other data? Possibly other fields, too.

jkczyz (commented):
Concept ACK. Feel free to ignore some of the more nitty comments for now. Interested more in constructing packages and serialization.

TheBlueMatt (approved):
I think I've basically gone through the whole thing in detail now. A few methods could be squashed to clean it up even more but it looks good.

lightning/src/chain/channelmonitor.rs
```diff
@@ -20,7 +20,7 @@
 //! security-domain-separated system design, you should consider having multiple paths for
 //! ChannelMonitors to get out of the HSM and onto monitoring devices.
 
-use bitcoin::blockdata::block::{Block, BlockHeader};
+use bitcoin::blockdata::block::BlockHeader;
```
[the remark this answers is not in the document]
ariard: Yes inadvertent corrected.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,698 @@
...
+		Ok(malleability)
+	}
+}
+
+/// A structure to describe a package content which is generated by ChannelMonitor and
+/// used by OnchainTxHandler to generate and broadcast transactions settling onchain claims.
+///
+/// A package is defined as one or more transactions claiming onchain outputs in reaction
+/// to confirmation of a channel transaction. Those packages might be aggregated to save on
+/// fees, if satisfaction of outputs's witnessScript let's us do so.
```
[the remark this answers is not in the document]
ariard: Yes a package is just a chain of transactions. A package template a chain of transactions with a superset of constraints such as "should be confirmed in chain before height X" or "must not be aggregated".

I concede that currently chain of transactions are one-sized which might make it confusing but in the rest of the patchset, commitment transactions are attached a CPFP, such becoming two-sized.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,698 @@
...
+	}
+	pub(crate) fn get_timer(&self) -> Option<u32> {
+		if let Some(ref timer) = self.height_timer {
+			return Some(*timer);
+		}
+		None
+	}
+	pub(crate) fn load_timer(&mut self, new_timer: Option<u32>) {
+		self.height_timer = new_timer;
+	}
```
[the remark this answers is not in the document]
ariard: Switched feerate/timer to idiomatic Rust naming. Will try to remember in the future :)

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,698 @@
...
+	absolute_timelock: u32,
+	// Timelocked outputs belonging to the same transaction might have differing
+	// satisfying heights. Picking up the highest height among the output set would be a valid
+	// aggregation strategy but it comes with at least 2 trade-offs :
+	// * soonest-output fund are going to take longer to come back
+	// * CLTV delta backing up a corresponding HTLC on an upstream channel could be swallowed
+	// by the requirement of the highest-output part of the set
+	// For now, we mark such timelocked outputs as non-aggregable, though we might introduce
+	// smarter aggregation strategy in the future.
+	aggregation: bool,
```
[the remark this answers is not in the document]
ariard: Beyond rewording comment, switched to `aggregable`, hopefully making field clearer.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,698 @@
...
+/// fees, if satisfaction of outputs's witnessScript let's us do so.
+///
+/// As packages are time-sensitive, we fee-bump and rebroadcast them at scheduled intervals.
+/// Failing to confirm a package translate as a loss of funds for the user.
+#[derive(Clone, PartialEq)]
+pub(crate) struct PackageTemplate {
+	// List of onchain outputs and solving data to generate satisfying witnesses.
+	inputs: Vec<(BitcoinOutPoint, PackageSolvingData)>,
+	// Packages are deemed as malleable if we have local knwoledge of at least one set of
+	// private keys yelling a satisfying witnesses. Malleability implies that we can aggregate
```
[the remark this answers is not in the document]
ariard: Ah, I meant yielding. Does it sound correct ?

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,698 @@
...
+		}
+	}
+	pub(crate) fn merge_package(&mut self, mut merge_from: PackageTemplate) {
+		assert_eq!(self.height_original, merge_from.height_original);
+		if self.malleability == PackageMalleability::Untractable || merge_from.malleability == PackageMalleability::Untractable {
+			panic!("Merging template on untractable packages");
+		}
+		if !self.aggregation || !merge_from.aggregation {
+			panic!("Merging non aggregatable packages");
+		}
```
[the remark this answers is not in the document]
ariard: Can you precise further your thinking here ? Ofc I can encode deeper aggregation/malleability at the output-level and if aggregation logic is correct it should ensure similarity of aggregation/malleability among package members. Though it is a big win in term of API clarity ? 

Further it will inflate serialized package size as you will redundantly encode properties ?

lightning/src/ln/onchain_utils.rs
ariard: Yep added new CounterpartyHTLCOfferedOutput/CounterpartyHTLCReceivedOutput to remove ambiguity around aggregation and thus enable inference at deserialization.

lightning/src/ln/onchain_utils.rs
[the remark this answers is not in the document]
jkczyz: Ah, yes it does.

lightning/src/ln/onchain_utils.rs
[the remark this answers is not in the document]
jkczyz: > Can you precise further your thinking here ? Ofc I can encode deeper aggregation/malleability at the output-level (i.e `PackageSolvingData`) and if aggregation logic is correct it should ensure similarity of aggregation/malleability among package members. Though is it a big win in term of API clarity ?
> 

I'm thinking something similar to `InvoiceBuilder` where `PackageTemplate` would be constructed piecemeal (i.e., chaining method calls for each field) and where these properties are enforced by the type system by conditionally implementing methods (e.g., a `PackageTemplate` marked by a type parameter as `Malleable` could only be merged with another such `PackageTemplate`). Similarly with `aggregation`.

lightning/src/chain/onchaintx.rs
```diff
@@ -305,7 +169,7 @@ pub struct OnchainTxHandler<ChannelSigner: Sign> {
 
 	latest_height: u32,
 
-	secp_ctx: Secp256k1<secp256k1::All>,
+	pub(super) secp_ctx: Secp256k1<secp256k1::All>,
```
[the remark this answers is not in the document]
ariard: It's still used in `onchain_utils`, actually I could initiate PackageTemplate with a Signer/secp256k1::All, though we don't have anymore Signer access in ChannelMonitor.

If we make OnchainTxhandler its own interface in the future, you may be able to pass a differing Signer than the ChannelMonitor though does this flexibility is really worthy and do we want to keep signing operations in OnchainTxHandler?

lightning/src/chain/onchain_utils.rs
```diff
@@ -0,0 +1,935 @@
...
+	} else {
+		new_fee
+	};
+	Some((new_fee, new_fee * 1000 / (predicted_weight as u64)))
+}
+
+/// Deduce a new proposed fee from the claiming transaction output value.
+/// If the new proposed fee is superior to the consumed outpoint's value, burn everything in miner's
+/// fee to deter counterparties attacker.
+pub(crate) fn compute_output_value<F: Deref, L: Deref>(predicted_weight: usize, input_amounts: u64, previous_feerate: u64, fee_estimator: &F, logger: &L) -> Option<(u64, u64)>
```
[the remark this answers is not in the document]
ariard: Right, they will be fulfilled at first transaction generation. Dropped.

lightning/src/chain/onchain_utils.rs
```diff
@@ -0,0 +1,935 @@
...
+						log_trace!(logger, "Adding claiming input for outpoint {}:{}", outpoint.txid, outpoint.vout);
+						log_trace!(logger, "Finalized transaction {} ready to broadcast", final_tx.txid());
+						return Some(final_tx);
+					}
+					return None;
+				} else { panic!("API Error: Package must not be inputs empty"); }
+			},
+		}
+	}
+	pub (crate) fn build_package(txid: Txid, vout: u32, input_solving_data: PackageSolvingData, absolute_timelock: u32, aggregation: bool, feerate_previous: u64, height_timer: Option<u32>, height_original: u32) -> Self {
```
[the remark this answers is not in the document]
ariard: Right, they will be fulfilled at first transaction generation. Dropped.

lightning/src/chain/onchaintx.rs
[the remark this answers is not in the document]
TheBlueMatt: Huh, I thought I compile-checked it. Its fine to leave as-is if its actually used, I just didn't see where.

lightning/src/chain/onchain_utils.rs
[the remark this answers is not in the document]
ariard: Make it part of the package interface, though note weight and amounts are still generated on their own as we use them for assertions/logs in `generate_claim_tx` path.

lightning/src/chain/channelmonitor.rs
```diff
@@ -1915,35 +1767,34 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
...
-				}});
+				let htlc_output = HolderHTLCOutput::build(if !htlc.offered {
+					if let Some(preimage) = self.payment_preimages.get(&htlc.payment_hash) {
+						Some(preimage.clone())
+					} else {
+						// We can't build an HTLC-Success transaction without the preimage
+						continue;
+					}
+				} else { None }, htlc.amount_msat);
+				let htlc_package = PackageTemplate::build_package(holder_tx.txid, transaction_output_index, PackageSolvingData::HolderHTLCOutput(htlc_output), height, false, 0, None, height);
```
TheBlueMatt, line 1796: I'm a bit confused by the timelock semantics. Previously we were always setting `absolute_timelock` to none here, but now we set it to the current height (though it should be current height + 1 if we're trying to do the "always timelock transactions" thing). Should we be moving towards `absolute_timelock` really meaning "the real timelock when this transaction can hit the chain"? I do think that would make it much more useful.

In that case, does the absolute_timelock here not need to be HTLC timelock + height (for HTLC-Timeout transactions)? May be worth adding some debug assertions that the transactions generated have a timelock that matches the absolute_timelock in the PackageTemplate.

ariard: I think I would agree on defining `absolute_timelock` as "the real hit when counterparty transaction can hit the chain" ?

Right, timelock should not be height but in fact `htlc.cltv_expiry`. Though current value of `height` is just a tweak to avoid generating non-bumped HTLC-transactions for nothing, which would break some tests. Note, it should be changed in the rest of the patchset iirc.

TheBlueMatt: Right, I think the other few calls already use that definition, just not this one, though it may also make sense to always do max(HTLC timeout, current height + 1)?

lightning/src/ln/onchain_utils.rs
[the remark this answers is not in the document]
ariard: TBH, I don't see a straightforward way to do such conditional methods implementation and considering the API should change a bit with remaining PRs of the patchset (e.g some output type becoming malleable), I would wait for them to land before to solidify the API now ? I think that's a good direction, just maybe too early.

lightning/src/chain/onchaintx.rs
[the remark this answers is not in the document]
ariard: Have a look in `PackageSolvingData::finalize_input` ?

lightning/src/ln/onchain_utils.rs
```diff
@@ -0,0 +1,90 @@
+//! Utilities for computing witnesses weight and feerate computation for onchain operation
+
+use ln::msgs::DecodeError;
+use util::ser::{Readable, Writer, Writeable};
+
+#[derive(PartialEq, Clone, Copy)]
+pub(crate) enum InputDescriptors {
```
jkczyz: Could be left for a later clean up: s/InputDescriptors/InputDescriptor

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
 
+//! Various utilities to assemble claimable outpoints in package of one or more transactions. Those
+//! packages are attached metadatas, guiding their aggregable or fee-bumping re-schedule. This file
```
jkczyz: s/metadatas/metadata

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+/// used by OnchainTxHandler to generate and broadcast transactions settling onchain claims.
+///
+/// A package is defined as one or more transactions claiming onchain outputs in reaction
+/// to confirmation of a channel transaction. Those packages might be aggregated to save on
+/// fees, if satisfaction of outputs's witnessScript let's us do so.
+///
+/// As packages are time-sensitive, we fee-bump and rebroadcast them at scheduled intervals.
+/// Failing to confirm a package translate as a loss of funds for the user.
+#[derive(Clone, PartialEq)]
+pub(crate) struct PackageTemplate {
```
jkczyz: No need to define structs inside-out order. That's mostly a C-ism. The topmost struct can be defined first in the file. Feel free to ignore, but it's somewhat of pet peeve of mine as it requires reading the file backwards. :)

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+// number_of_witness_elements + sig_length + revocation_sig + pubkey_length + revocationpubkey + witness_script_length + witness_script
+const WEIGHT_REVOKED_RECEIVED_HTLC: u64 = 1 + 1 + 73 + 1 + 33 + 1 +  139;
+// number_of_witness_elements + sig_length + counterpartyhtlc_sig  + preimage_length + preimage + witness_script_length + witness_script
+const WEIGHT_OFFERED_HTLC: u64 = 1 + 1 + 73 + 1 + 32 + 1 + 133;
+// number_of_witness_elements + sig_length + revocation_sig + pubkey_length + revocationpubkey + witness_script_length + witness_script
+const WEIGHT_RECEIVED_HTLC: u64 = 1 + 1 + 73 + 1 + 1 + 1 + 139;
+// number_of_witness_elements + sig_length + revocation_sig + true_length + op_true + witness_script_length + witness_script
+const WEIGHT_REVOKED_OUTPUT: u64 = 1 + 1 + 73 + 1 + 1 + 1 + 77;
+
+/// A struct to describe a revoked output and corresponding informations to generate a solving
```
jkczyz: s/informations/information

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+	per_commitment_point,
+	counterparty_delayed_payment_base_key,
+	counterparty_htlc_base_key,
+	per_commitment_key,
+	weight,
+	amount,
+	on_counterparty_tx_csv
+});
+
+/// A struct to describe a revoked offered output and corresponding informations to generate a
```
jkczyz: s/informations/information

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+/// solving witness.
+///
+/// HTLCOuputInCommitment (hash timelock, direction) and pubkeys are used to generate the suiting
```
jkczyz: s/the suiting/a suitable?

Likewise elsewhere.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+			1 => {
+				PackageMalleability::Untractable
+			},
+			_ => return Err(DecodeError::UnknownVersion)
+		};
+		Ok(malleability)
+	}
+}
+
+/// A structure to describe a package content which is generated by ChannelMonitor and
```
jkczyz: s/which/that

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+			// Note, Revoked*Output are sharing the same category as their aggregable
+			// doesn't raise issue, assuming timelocks are respected.
+			PackageSolvingData::RevokedOutput(..) => { 0 },
+			PackageSolvingData::RevokedHTLCOutput(..) => { 0 },
+			PackageSolvingData::CounterpartyOfferedHTLCOutput(..) => { 1 },
+			PackageSolvingData::CounterpartyReceivedHTLCOutput(..) => { 1 },
+			PackageSolvingData::HolderHTLCOutput(..) => { 2 },
+			PackageSolvingData::HolderFundingOutput(..) => { 3 },
+		};
+		category
```
jkczyz: What is category suppose to represent? Is it defined anywhere? Could you use an enum instead?

lightning/src/chain/onchain_utils.rs
```diff
@@ -0,0 +1,972 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Various utilities to assemble claimable outpoints in package of one or more transactions. Those
```
jkczyz, line 10: Consider naming this module `package` rather than using a grab bag `onchain_utils`. The parent module is already about the chain.

lightning/src/chain/onchain_utils.rs
```diff
@@ -0,0 +1,972 @@
...
+//! packages are attached metadatas, guiding their aggregable or fee-bumping re-schedule. This file
```
jkczyz, line 11: s/metadats/metadata

jkczyz (commented):
Mostly nits on `chain_utils`. Can take a closer look at later commits tonight.

lightning/src/ln/onchain_utils.rs
```diff
@@ -476,26 +476,26 @@ impl Readable for PackageMalleability {
...
 	// Determines if this package can be aggregated.
 	// Timelocked outputs belonging to the same transaction might have differing
-	// satisfying heights. Picking up the highest height among the output set would be a valid
+	// satisfying heights. Picking up the later height among the output set would be a valid
 	// aggregable strategy but it comes with at least 2 trade-offs :
-	// * soonest-output fund are going to take longer to come back
+	// * earlier-output fund are going to take longer to come back
 	// * CLTV delta backing up a corresponding HTLC on an upstream channel could be swallowed
-	// by the requirement of the highest-output part of the set
+	// by the requirement of the later-output part of the set
```
jkczyz: Looks like these are in the wrong commit.

lightning/src/chain/mod.rs
```diff
@@ -23,6 +23,8 @@ pub mod chainmonitor;
 pub mod channelmonitor;
 pub mod transaction;
 pub mod keysinterface;
+pub(crate) mod onchaintx;
+pub(crate) mod onchain_utils;
```
jkczyz, line 27: "chain" seems redundant in these names after the move, though maybe not worth changing right now.

lightning/src/chain/onchaintx.rs
```diff
@@ -347,7 +333,7 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
 
 		// Compute new height timer to decide when we need to regenerate a new bumped version of the claim tx (if we
 		// didn't receive confirmation of it before, or not enough reorg-safe depth on top of it).
-		let new_timer = Some(Self::get_height_timer(height, cached_request.timelock()));
+		let new_timer = Some(onchain_utils::get_height_timer(height, cached_request.timelock()));
```
jkczyz, line 336: Why move this to `onchain_utils` if it is only used here? I think we should avoid "utils" modules as they tend to collect unrelated code.

jkczyz (commented):
Code mostly looks good. Comments are primarily around doc confusion and code organization.

lightning/src/ln/onchain_utils.rs
ariard: InputDescriptors are removed after this PR, so not sure if it's worthy to correct.

lightning/src/ln/onchain_utils.rs
ariard: Well I'm coming from a strong C background so quite spontaneous for me to sort structs inside-out :) I'll leave as it is for now but think about it next time I'm defining structs.

lightning/src/ln/onchain_utils.rs
ariard: It is supposed to represent an output type. As `type` is a keyword` in Rust I did opt-in for "category". Switched to an enum.

lightning/src/ln/onchain_utils.rs
```diff
@@ -1,8 +1,741 @@
...
+				PackageMalleability::Malleable
+			},
+			1 => {
+				PackageMalleability::Untractable
+			},
+			_ => return Err(DecodeError::UnknownVersion)
+		};
+		Ok(malleability)
+	}
+}
```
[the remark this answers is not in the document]
ariard: No, dropped!

lightning/src/ln/onchain_utils.rs
```diff
@@ -821,3 +822,104 @@ pub(crate) fn get_witnesses_weight(inputs: &[InputDescriptors]) -> usize {
 	}
 	tx_weight
 }
+
+/// Attempt to propose a fee based on consumed outpoints's values and predicted weight of the claiming
```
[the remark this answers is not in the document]
ariard: Good remarks, it's should be _output's values_, we may have multiple outputs spent aggregated in one transaction and we're willingly to guess its fee.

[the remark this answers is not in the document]
ariard: Yes, reword a bit, let me know if it's clearer.

lightning/src/ln/onchain_utils.rs
```diff
@@ -821,3 +822,104 @@ pub(crate) fn get_witnesses_weight(inputs: &[InputDescriptors]) -> usize {
...
+/// transaction. We start with the highest priority feerate returned by the node's fee estimator
+/// then fall-back to lower priorities until we have enough value available to suck from.
+/// If the proposed fee is more than the claimable balance, we return None
+fn subtract_high_prio_fee<F: Deref, L: Deref>(input_amounts: u64, predicted_weight: usize, fee_estimator: &F, logger: &L) -> Option<(u64, u64)>
```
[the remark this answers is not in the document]
ariard: Right, it's used to subtract directly from spent input amounts, now it's just proposing a fee and subtraction is done in `compute_output_value`. Corrected.

lightning/src/chain/mod.rs
ariard: Well switched to "package.rs" so it's a bit better.

lightning/src/chain/onchaintx.rs
ariard: Right, made it part of the package interface as it's belong to it.

lightning/src/chain/channelmonitor.rs
ariard: IIRC, if HTLC timeout is already behind the chain tip, we're going to bump it at every block according to `get_block_height_timer`, so doing max(HTLC timeout, current height + 1) isn't going to change anything.

lightning/src/ln/package.rs
```diff
@@ -1,8 +1,719 @@
...
+
+/// A enum marking output category
+#[derive(PartialEq)]
+enum OutputCategory {
+	Revoked,
+	CounterpartyOfferedHTLC,
+	CounterpartyReceivedHTLC,
+	HolderHTLC,
+	HolderFunding,
+}
```
jkczyz: Looking at how this is used, it doesn't really seem necessary. You can instead have a method on `PackageSolvingData` that takes another `PackageSolvingData` and returns whether they are compatible. No need to introduce the term "category" which isn't defined. This would also eliminate the redundancy between `OutputCategory` and `PackageSolvingData`.

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+	// Cache of package feerate committed at previous (re)broadcast. If bumping resources
+	// (either claimed output value or external utxo), it will keep increasing until holder
+	// or counterparty successful claim.
+	feerate_previous: u64,
+	// Cache of next height at which fee-bumping and rebroadcast will be attempted. In
+	// the future, we might abstract it to an observed mempool fluctuation.
+	height_timer: Option<u32>,
+	// Confirmation height of the claimed outputs set transaction. In case of reorg reaching
+	// it, we wipe out and forget the package.
+	height_original: u32,
```
TheBlueMatt, line 491: This appears to be entirely unused except for assertions, is it worth keeping?

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+	// Packages are deemed as malleable if we have local knwoledge of at least one set of
+	// private keys yielding a satisfying witnesses. Malleability implies that we can aggregate
+	// packages among them to save on fees or rely on RBF to bump their feerates.
+	// Untractable packages have been counter-signed and thus imply that we can't aggregate
+	// them without breaking signatures. Fee-bumping strategy will also rely on CPFP.
+	malleability: PackageMalleability,
+	// Block height after which the earlier-output belonging to this package is mature for a
+	// competing claim by the counterparty. As our chain tip becomes nearer from the timelock,
+	// the fee-bumping frequency will increase. See `OnchainTxHandler::get_height_timer`.
+	absolute_timelock: u32,
```
TheBlueMatt, line 471: This name really confused me - `absolute_timelock` implies the timelock on a transaction (presumably the transaction being created), but this variable really means "the time at which our counterparty can potentially spend the claims here", maybe call it something like `soonest_conf_deadline` or something?

lightning/src/chain/package.rs
TheBlueMatt: Actually, I take that back, I need it for later things, so definitely keep it.

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+impl HolderHTLCOutput {
+	pub(crate) fn build(preimage: Option<PaymentPreimage>, amount: u64) -> Self {
+		HolderHTLCOutput {
+			preimage,
+			amount
+		}
+	}
+}
+
+impl_writeable_len_match!(HolderHTLCOutput, {
```
TheBlueMatt, line 219: We shouldn't use `len_match`, I dont think, for structs that are contained in larger structs contained in larger structs, if anything its a pessimism instead of an optimization.

lightning/src/chain/package.rs
```diff
@@ -642,6 +642,19 @@ impl PackageTemplate {
...
+	/// height than once reached we should generate a new bumped "version" of the claim tx to be sure than we safely claim outputs before
+	/// than our counterparty can do it too. If timelock expires soon, height timer is going to be scale down in consequence to increase
+	/// frequency of the bump and so increase our bets of success.
+	pub(crate) fn get_height_timer(&self, current_height: u32) -> u32 {
+		if self.absolute_timelock <= current_height + 3 {
+			return current_height + 1
+		} else if self.absolute_timelock - current_height <= 15 {
+			return current_height + 3
+		}
+		current_height + 15
```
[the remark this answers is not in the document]
ariard: Added constants `*_FREQUENCY_BUM_INTERVAL`, though for now values are quite arbitrary.

lightning/src/ln/package.rs
```diff
@@ -1,8 +1,719 @@
...
+
+/// A wrapper encapsulating all in-protocol differing outputs categories.
```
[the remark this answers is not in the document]
ariard: In my mind that's the purpose of a wrapper, encapsulating an object without altering type. If you have a finer suggestion w.r.t. to category, which is in this context designates a type, I'll take it. Naming is hard :/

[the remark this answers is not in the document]
jkczyz: I think you can just say "output types".

lightning/src/ln/package.rs
jkczyz: BTW, such comparison can be accomplished for complex enum types like `PackageSolvingData` using `std::mem::discriminant`.

https://doc.rust-lang.org/std/mem/fn.discriminant.html

ariard: Okay finally dropped OutputCategory/OutputType! And leverage usage of `std::mem::discriminant`. There is still a bunch of special logic in `output_type_eq` to handle the forced-equality for revoked outputs.

lightning/src/chain/package.rs
ariard: Ah good to know, updated for HolderHTLCOutput/HolderFundingOutput!

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+				}
+			},
+			PackageSolvingData::RevokedHTLCOutput(..) => {
+				match input {
+					PackageSolvingData::RevokedOutput(..) => { return true },
+					PackageSolvingData::RevokedHTLCOutput(..) => { return true },
+					_ => { return false }
+				}
+			},
+			_ => { return mem::discriminant(&self) == mem::discriminant(&input) }
```
TheBlueMatt, line 311:
```
error: calling `mem::discriminant` on non-enum type `&chain::package::PackageSolvingData`
   --> lightning/src/chain/package.rs:311:18
```

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+	}
+}
+
+impl_writeable!(HolderFundingOutput, 0, {
+	funding_redeemscript
+});
+
+/// A wrapper encapsulating all in-protocol differing outputs types.
+///
+/// The generic API offers access to outputs attributes (amount, weight, output_type) or allow
```
jkczyz, line 254: This seems like a circular definition. If each variant is an output type, then it can't have an output type as an attribute. I'd drop this list entirely and say "common attributes"

I think instead you want to rename `output_type_eq` to something descriptive like `is_compatible` (or whatever is most accurate) -- best way to figure out how to name something is to first document it.

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+			// Note: Currently, weights of holder outputs spending witnesses aren't used
+			// as we can't malleate spending package to increase their feerate. This
+			// should change with the remaining anchor output patchset.
+			PackageSolvingData::HolderHTLCOutput(..) => { debug_assert!(false); 0 },
+			PackageSolvingData::HolderFundingOutput(..) => { debug_assert!(false); 0 },
+		};
+		weight
+	}
+	fn output_type_eq(&self, input: &PackageSolvingData) -> bool {
+		match self {
```
jkczyz, line 296: Can drop all the `return` keywords from these matches.

lightning/src/chain/package.rs
jkczyz, line 254: s/outputs attributes/an output's attributes

lightning/src/chain/package.rs
```diff
@@ -0,0 +1,1037 @@
...
+			PackageSolvingData::RevokedOutput(..) => {
+				match input {
+					PackageSolvingData::RevokedHTLCOutput(..) => { return true },
+					PackageSolvingData::RevokedOutput(..) => { return true },
```
jkczyz, line 300: If these variants only differ by one field but are considered "equal", would it be simpler to make that particular field an enum and collapse these two variants into one variant rather than repeat the six other fields across each?

ariard: I was anticipating future implementation of disaggregation-after-timelock to solve this transaction pinning : https://github.com/lightningnetwork/lightning-rfc/pull/803. If this mitigation is implemented they won't be considered "equal" anymore.

lightning/src/chain/package.rs
ariard: Solved!

jkczyz (approved):
Looks good! Thanks for addressing everything.

TheBlueMatt (approved):
Looks good. Anything else can come in a followup.

PR #667, opened 2020-08-10
Remove Channel's ChannelMonitor copy

### LINKED ISSUE #657: Get rid of unnecessary ChannelMonitor in Channel
In preparation for refactoring how we persist channel data, we need to get rid of the `Channel`'s `ChannelMonitor`. The only reason it exists now is because of the `ChannelManager` calling `would_broadcast_at_height`. `would_broadcast_at_height` provides information that should be accessible in a different way (that doesn't require the `Channel` to hold its `ChannelMonitor`).

valentinewallace:
In service to the larger refactor of removing the Channel's reference
to its ChannelMonitor.

This also allows us to remove the Channel's `channel_monitor()` function. Next we can remove all of the Channel's logic for keeping its channel monitor up to date, as well as the `update_monitor_ooo` logic in `ChannelMonitor`. 

~Partially resolves~ Closes #657

### DIFF
--- a/lightning/src/ln/channel.rs
+++ b/lightning/src/ln/channel.rs
@@ -386,9 +386,6 @@ pub(super) struct Channel<ChanSigner: ChannelKeys> {
 
 	their_shutdown_scriptpubkey: Option<Script>,
 
-	/// Used exclusively to broadcast the latest local state, mostly a historical quirk that this
-	/// is here:
-	channel_monitor: Option<ChannelMonitor<ChanSigner>>,
 	commitment_secrets: CounterpartyCommitmentSecrets,
 
 	network_sync: UpdateStatus,
@@ -557,7 +554,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 
 			their_shutdown_scriptpubkey: None,
 
-			channel_monitor: None,
 			commitment_secrets: CounterpartyCommitmentSecrets::new(),
 
 			network_sync: UpdateStatus::Fresh,
@@ -786,7 +782,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 
 			their_shutdown_scriptpubkey,
 
-			channel_monitor: None,
 			commitment_secrets: CounterpartyCommitmentSecrets::new(),
 
 			network_sync: UpdateStatus::Fresh,
@@ -1222,7 +1217,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 				payment_preimage: payment_preimage_arg.clone(),
 			}],
 		};
-		self.channel_monitor.as_mut().unwrap().update_monitor_ooo(monitor_update.clone(), logger).unwrap();
 
 		if (self.channel_state & (ChannelState::AwaitingRemoteRevoke as u32 | ChannelState::PeerDisconnected as u32 | ChannelState::MonitorUpdateFailed as u32)) != 0 {
 			for pending_update in self.holding_cell_htlc_updates.iter() {
@@ -1552,7 +1546,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 			} }
 		}
 
-		self.channel_monitor = Some(create_monitor!());
 		let channel_monitor = create_monitor!();
 
 		self.channel_state = ChannelState::FundingSent as u32;
@@ -1618,7 +1611,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 			} }
 		}
 
-		self.channel_monitor = Some(create_monitor!());
 		let channel_monitor = create_monitor!();
 
 		assert_eq!(self.channel_state & (ChannelState::MonitorUpdateFailed as u32), 0); // We have no had any monitor(s) yet to fail update!
@@ -2060,7 +2052,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 				htlc_outputs: htlcs_and_sigs
 			}]
 		};
-		self.channel_monitor.as_mut().unwrap().update_monitor_ooo(monitor_update.clone(), logger).unwrap();
 
 		for htlc in self.pending_inbound_htlcs.iter_mut() {
 			let new_forward = if let &InboundHTLCState::RemoteAnnounced(ref forward_info) = &htlc.state {
@@ -2280,7 +2271,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 				secret: msg.per_commitment_secret,
 			}],
 		};
-		self.channel_monitor.as_mut().unwrap().update_monitor_ooo(monitor_update.clone(), logger).unwrap();
 
 		// Update state now that we've passed all the can-fail calls...
 		// (note that we may still fail to generate the new commitment_signed message, but that's
@@ -3115,14 +3105,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 		self.user_id
 	}
 
-	/// May only be called after funding has been initiated (ie is_funding_initiated() is true)
-	pub fn channel_monitor(&mut self) -> &mut ChannelMonitor<ChanSigner> {
-		if self.channel_state < ChannelState::FundingSent as u32 {
-			panic!("Can't get a channel monitor until funding has been created");
-		}
-		self.channel_monitor.as_mut().unwrap()
-	}
-
 	/// Guaranteed to be Some after both FundingLocked messages have been exchanged (and, thus,
 	/// is_usable() returns true).
 	/// Allowed in any state (including after shutdown)
@@ -3397,9 +3379,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 		if header.bitcoin_hash() != self.last_block_connected {
 			self.last_block_connected = header.bitcoin_hash();
 			self.update_time_counter = cmp::max(self.update_time_counter, header.time);
-			if let Some(channel_monitor) = self.channel_monitor.as_mut() {
-				channel_monitor.last_block_hash = self.last_block_connected;
-			}
 			if self.funding_tx_confirmations > 0 {
 				if self.funding_tx_confirmations == self.minimum_depth as u64 {
 					let need_commitment_update = if non_shutdown_state == ChannelState::FundingSent as u32 {
@@ -3458,9 +3437,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 			self.funding_tx_confirmations = self.minimum_depth as u64 - 1;
 		}
 		self.last_block_connected = header.bitcoin_hash();
-		if let Some(channel_monitor) = self.channel_monitor.as_mut() {
-			channel_monitor.last_block_hash = self.last_block_connected;
-		}
 		false
 	}
 
@@ -3871,7 +3847,6 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 				their_revocation_point: self.their_cur_commitment_point.unwrap()
 			}]
 		};
-		self.channel_monitor.as_mut().unwrap().update_monitor_ooo(monitor_update.clone(), logger).unwrap();
 		self.channel_state |= ChannelState::AwaitingRemoteRevoke as u32;
 		Ok((res, monitor_update))
 	}
@@ -4242,8 +4217,6 @@ impl<ChanSigner: ChannelKeys + Writeable> Writeable for Channel<ChanSigner> {
 		self.their_shutdown_scriptpubkey.write(writer)?;
 
 		self.commitment_secrets.write(writer)?;
-
-		self.channel_monitor.as_ref().unwrap().write_for_disk(writer)?;
 		Ok(())
 	}
 }
@@ -4398,13 +4371,6 @@ impl<ChanSigner: ChannelKeys + Readable> Readable for Channel<ChanSigner> {
 		let their_shutdown_scriptpubkey = Readable::read(reader)?;
 		let commitment_secrets = Readable::read(reader)?;
 
-		let (monitor_last_block, channel_monitor) = Readable::read(reader)?;
-		// We drop the ChannelMonitor's last block connected hash cause we don't actually bother
-		// doing full block connection operations on the internal ChannelMonitor copies
-		if monitor_last_block != last_block_connected {
-			return Err(DecodeError::InvalidValue);
-		}
-
 		Ok(Channel {
 			user_id,
 
@@ -4476,7 +4442,6 @@ impl<ChanSigner: ChannelKeys + Readable> Readable for Channel<ChanSigner> {
 
 			their_shutdown_scriptpubkey,
 
-			channel_monitor: Some(channel_monitor),
 			commitment_secrets,
 
 			network_sync: UpdateStatus::Fresh,
--- a/lightning/src/ln/channelmanager.rs
+++ b/lightning/src/ln/channelmanager.rs
@@ -38,7 +38,7 @@ use bitcoin::secp256k1;
 use chain::chaininterface::{BroadcasterInterface,ChainListener,FeeEstimator};
 use chain::transaction::OutPoint;
 use ln::channel::{Channel, ChannelError};
-use ln::channelmonitor::{ChannelMonitor, ChannelMonitorUpdate, ChannelMonitorUpdateErr, ManyChannelMonitor, HTLC_FAIL_BACK_BUFFER, CLTV_CLAIM_BUFFER, LATENCY_GRACE_PERIOD_BLOCKS, ANTI_REORG_DELAY};
+use ln::channelmonitor::{ChannelMonitor, ChannelMonitorUpdate, ChannelMonitorUpdateErr, ManyChannelMonitor, HTLC_FAIL_BACK_BUFFER, CLTV_CLAIM_BUFFER, LATENCY_GRACE_PERIOD_BLOCKS, ANTI_REORG_DELAY, MonitorEvent};
 use ln::features::{InitFeatures, NodeFeatures};
 use routing::router::{Route, RouteHop};
 use ln::msgs;
@@ -2966,6 +2966,48 @@ impl<ChanSigner: ChannelKeys, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> 
 			Err(e) => { Err(APIError::APIMisuseError { err: e.err })}
 		}
 	}
+
+	/// Process pending events from the ManyChannelMonitor.
+	fn process_pending_monitor_events(&self) {
+		let mut failed_channels = Vec::new();
+		{
+			for monitor_event in self.monitor.get_and_clear_pending_monitor_events() {
+				match monitor_event {
+					MonitorEvent::HTLCEvent(htlc_update) => {
+						if let Some(preimage) = htlc_update.payment_preimage {
+							log_trace!(self.logger, "Claiming HTLC with preimage {} from our monitor", log_bytes!(preimage.0));
+							self.claim_funds_internal(self.channel_state.lock().unwrap(), htlc_update.source, preimage);
+						} else {
+							log_trace!(self.logger, "Failing HTLC with hash {} from our monitor", log_bytes!(htlc_update.payment_hash.0));
+							self.fail_htlc_backwards_internal(self.channel_state.lock().unwrap(), htlc_update.source, &htlc_update.payment_hash, HTLCFailReason::Reason { failure_code: 0x4000 | 8, data: Vec::new() });
+						}
+					},
+					MonitorEvent::CommitmentTxBroadcasted(funding_outpoint) => {
+						let mut channel_lock = self.channel_state.lock().unwrap();
+						let channel_state = &mut *channel_lock;
+						let by_id = &mut channel_state.by_id;
+						let short_to_id = &mut channel_state.short_to_id;
+						let pending_msg_events = &mut channel_state.pending_msg_events;
+						if let Some(mut chan) = by_id.remove(&funding_outpoint.to_channel_id()) {
+							if let Some(short_id) = chan.get_short_channel_id() {
+								short_to_id.remove(&short_id);
+							}
+							failed_channels.push(chan.force_shutdown(false));
+							if let Ok(update) = self.get_channel_update(&chan) {
+								pending_msg_events.push(events::MessageSendEvent::BroadcastChannelUpdate {
+									msg: update
+								});
+							}
+						}
+					},
+				}
+			}
+		}
+
+		for failure in failed_channels.drain(..) {
+			self.finish_force_close_channel(failure);
+		}
+	}
 }
 
 impl<ChanSigner: ChannelKeys, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> events::MessageSendEventsProvider for ChannelManager<ChanSigner, M, T, K, F, L>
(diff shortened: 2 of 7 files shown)


valentinewallace:
Oops, broke build -- will fix in a bit

valentinewallace:
I looked into extracting some common HTLC selection functionality between `monitor_would_broadcast` and `build_commitment_transaction` but they're pretty different, they want different HTLCs, and monitor_would_broadcast needs local _and_ remote HTLCs, and idk. Let me know if I'm missing something and it's definitely a good idea.

valentinewallace:
Hmm, fuzzing seems slow still. Looking into it.

lightning/src/ln/channel.rs
```diff
@@ -293,6 +291,7 @@ pub(super) struct Channel<ChanSigner: ChannelKeys> {
 	pending_inbound_htlcs: Vec<InboundHTLCOutput>,
 	pending_outbound_htlcs: Vec<OutboundHTLCOutput>,
 	holding_cell_htlc_updates: Vec<HTLCUpdateAwaitingACK>,
+	payment_preimages: HashMap<PaymentHash, PaymentPreimage>,
```
TheBlueMatt, line 294: I don't see where this is pruned from? In ChannelMonitor we remove entries when we're sure we no longer need them (ie all non-revoked commitment txn don't have any HTLC with the given hash), and otherwise this grows unbounded with every HTLC we relay.

lightning/src/ln/channel.rs
```diff
@@ -795,6 +791,71 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
...
+					transaction_output_index: None
+				});
+			}
+		}
+
+		let cap = self.pending_inbound_htlcs.len() + self.pending_outbound_htlcs.len();
+		let mut remote_outputs = Vec::with_capacity(cap);
+		let mut local_outputs = Vec::with_capacity(cap);
+		let awaiting_raa = (self.channel_state & ChannelState::AwaitingRemoteRevoke as u32) != 0;
+		for ref htlc in self.pending_inbound_htlcs.iter() {
```
TheBlueMatt, line 811: I figure you should be able to do this with an `iter().map(..)` to avoid putting every HTLC into a Vec here, no?

lightning/src/ln/channelmanager.rs
```diff
@@ -3104,7 +3104,17 @@ impl<ChanSigner: ChannelKeys, M: Deref + Sync + Send, T: Deref + Sync + Send, K:
...
 				}
-				if channel.is_funding_initiated() && channel.channel_monitor().would_broadcast_at_height(height, &self.logger) {
+				#[cfg(any(test, feature = "fuzztarget"))]
+				// In testing/fuzzing, check that the channel's would_broadcast_at_height
+				// implementation is always identical to the ChannelMonitor equivalent.
+				let funding_initiated = channel.is_funding_initiated();
+				#[cfg(any(test, feature = "fuzztarget"))]
+				for i in height..height + 10 {
+					assert_eq!(funding_initiated && channel.monitor_would_broadcast_at_height(i, &self.logger),
+										 funding_initiated && self.monitor.get_monitor_would_broadcast(&channel.get_funding_txo().unwrap(), i));
```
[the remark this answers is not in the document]
valentinewallace: Ok, sg -- FYI, I lowered it to 10 because `144` was making the fuzz tests run ~80 minutes on GH Actions (up from ~15 minutes), and about an hour locally (up from ~5 minutes, on my computer).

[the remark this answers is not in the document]
TheBlueMatt: Hmm, would be good to figure out why its so slow - it may be that reducing allocations inside the Channel call would improve it.

lightning/src/ln/channel.rs
```diff
@@ -109,11 +110,8 @@ enum InboundHTLCState {
 	/// Note that we have to keep an eye on the HTLC until we've received a broadcastable
 	/// commitment transaction without it as otherwise we'll have to force-close the channel to
 	/// claim it before the timeout (obviously doesn't apply to revoked HTLCs that we can't claim
-	/// anyway). That said, ChannelMonitor does this for us (see
-	/// ChannelMonitor::would_broadcast_at_height) so we actually remove the HTLC from our own
-	/// local state before then, once we're sure that the next commitment_signed and
-	/// ChannelMonitor::provide_latest_local_commitment_tx_info will not include this HTLC.
-	LocalRemoved(InboundHTLCRemovalReason),
+	/// anyway).
+	LocalRemoved(bool, InboundHTLCRemovalReason),
```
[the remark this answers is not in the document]
arik-so: I also think a random `bool`'s meaning is hard to extract without a corresponding comment. Type names simplify that a bit.

lightning/src/ln/channel.rs
valentinewallace: Oh, I thought that in lightning you want to hold on to (at least some) preimages in the case where we're paying an invoice, as "proof of payment."

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
valentinewallace: Unfortunately the fuzz tests still seem to be over an hour on GH actions. From my testing, the time increase seems correlated to the `for i in height..height + X`'s `X` variable here.

Do you think there'd be significant speedup if, say, the HTLCs were checked in one pass instead of two? Otherwise, maybe we could compromise with a number in the middle? Though FWIW maybe it's not such a big deal to have slow CI runs.

lightning/src/ln/channel.rs
TheBlueMatt: Right, as a sender you may (though we push that logic onto our users), but an intermediate node shouldn't care.

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
TheBlueMatt: Hmm, right, may just be amount of work - probably the most immediate thing to try is figure out which fuzzer it is (chanmon_consistency, probably) and reduce the number of block_connected calls (L326) that are made no matter the input.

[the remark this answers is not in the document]
valentinewallace: Ok cool, I halved the number of `block_connected` calls as you say, and that halved the fuzz runtime. Do you think ~40 minutes is OK for GH actions, or should we go even lower?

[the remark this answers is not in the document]
TheBlueMatt: Less about the runtime in GH, more about making sure fuzzing is quick enough to find issues - I think we can set the block_connected calls there to whatever is required to lock in a channel, instead of an arbitrary (and high) number like 50.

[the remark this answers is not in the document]
valentinewallace: Cool, the runtime is totally back to normal range now so the fuzzers should be about as fast as they were before.

lightning/src/ln/channelmonitor.rs
```diff
@@ -1993,7 +1983,30 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
 		self.last_block_hash = block_hash.clone();
 	}
 
-	pub(super) fn would_broadcast_at_height<L: Deref>(&self, height: u32, logger: &L) -> bool where L::Target: Logger {
+	fn would_broadcast_at_height<L: Deref>(&self, height: u32, logger: &L) -> bool where L::Target: Logger {
+		let local_outputs = self.current_local_commitment_tx.htlc_outputs.iter().map(|&(ref a, _, _)| a);
+
+		let map_fn: fn(&(HTLCOutputInCommitment, Option<Box<HTLCSource>>)) -> &HTLCOutputInCommitment = |a| &a.0;
```
arik-so, line 1989: I'd rename this to extract_commitment or something more descriptive

lightning/src/ln/channel.rs
```diff
@@ -795,6 +796,57 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
...
+			.map(|htlc| get_htlc_output!(htlc, false));
+
+		let outbound_local_outputs = self.pending_outbound_htlcs.iter()
+			.filter_map(|htlc| {
+				match htlc.state {
+					OutboundHTLCState::AwaitingRemovedRemoteRevoke(_) => None,
+					_ => Some(get_htlc_output!(htlc, true))
+				}
+			});
+		let awaiting_raa = (self.channel_state & ChannelState::AwaitingRemoteRevoke as u32) != 0;
```
arik-so, line 821: `raa` should be expanded for easier legibility

lightning/src/ln/channelmonitor.rs
```diff
@@ -1993,7 +1983,30 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
...
+		let dummy = Vec::new();
+		let curr_remote_outputs = if let Some(ref txid) = self.current_remote_commitment_txid {
+			if let Some(ref htlc_outputs) = self.remote_claimable_outpoints.get(txid) {
+				htlc_outputs.iter().map(map_fn)
+			} else { dummy.iter().map(map_fn) }
+		} else { dummy.iter().map(map_fn) };
+
+		let map_fn: fn(&(HTLCOutputInCommitment, Option<Box<HTLCSource>>)) -> &HTLCOutputInCommitment = |a| &a.0;
```
arik-so, line 1997: what's the difference between the two map_fns? Can't the first one be reused?

TheBlueMatt:
Ooooo, right, so I think that's true - `would_broadcast_at_height` is only used to close a channel and we could happily have the `ChannelMonitor` handle that (though currently `Channel[Manager]` handles that fully). My knee-jerk reaction was that there would be a race condition where we'd continue updating a channel state after its been closed (and the latest local commitment broadcast), but I think we could not even bother notifying the ChannelManager that we're closing the channel and wait for `provide_latest_local_commitment_tx_info` to return an Err because `local_tx_signed` is set to close the channel, so there's no race.

lightning/src/ln/channelmanager.rs
```diff
@@ -3007,22 +3034,49 @@ impl<ChanSigner: ChannelKeys, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
...
 		{
 			//TODO: This behavior should be documented.
-			for htlc_update in self.monitor.get_and_clear_pending_htlcs_updated() {
-				if let Some(preimage) = htlc_update.payment_preimage {
-					log_trace!(self.logger, "Claiming HTLC with preimage {} from our monitor", log_bytes!(preimage.0));
-					self.claim_funds_internal(self.channel_state.lock().unwrap(), htlc_update.source, preimage);
-				} else {
-					log_trace!(self.logger, "Failing HTLC with hash {} from our monitor", log_bytes!(htlc_update.payment_hash.0));
-					self.fail_htlc_backwards_internal(self.channel_state.lock().unwrap(), htlc_update.source, &htlc_update.payment_hash, HTLCFailReason::Reason { failure_code: 0x4000 | 8, data: Vec::new() });
+			for monitor_event in self.monitor.get_and_clear_pending_monitor_events() {
```
TheBlueMatt, line 3043: Maybe take this opportunity to DRY this up with the same code in `get_and_clear_pending_msg_events()`?

lightning/src/ln/channelmonitor.rs
```diff
@@ -1350,30 +1366,10 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
 	{
 		for tx in self.get_latest_local_commitment_txn(logger).iter() {
 			broadcaster.broadcast_transaction(tx);
+			self.pending_monitor_events.push(MonitorEvent::CommitmentTxBroadcasted(self.funding_info.0));
```
TheBlueMatt, line 1369: Should be outside of the loop, no?

TheBlueMatt (commented):
I had to admit I'm somewhat surprised that's the only test changes you need, but looks good.

lightning/src/ln/channelmanager.rs
```diff
@@ -2966,6 +2966,52 @@ impl<ChanSigner: ChannelKeys, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
 			Err(e) => { Err(APIError::APIMisuseError { err: e.err })}
 		}
 	}
+
+	/// Process pending events from the ManyChannelMonitor.
+	fn process_pending_monitor_events(&self) {
+		let mut failed_channels = Vec::new();
+		// TODO: Event release to users and serialization is currently race-y: it's very easy for a
```
TheBlueMatt, line 2973: I think we can drop this since #617 :).

lightning/src/ln/channelmanager.rs
```diff
@@ -2966,6 +2966,52 @@ impl<ChanSigner: ChannelKeys, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
...
+		// user to serialize a ChannelManager with pending events in it and lose those events on
+		// restart. This is doubly true for the fail/fulfill-backs from monitor events!
+		{
+			//TODO: This behavior should be documented.
```
TheBlueMatt, line 2977: I think this should be duplicated across the two get_and_clear* methods - the point is that the behavior where we query channelmonitors when clearing other events is, at the minimum, very surprising.

lightning/src/ln/channelmonitor.rs
```diff
@@ -1966,7 +1962,11 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
 				}
 			}
 		}
-		self.onchain_tx_handler.block_connected(txn_matched, claimable_outpoints, height, &*broadcaster, &*fee_estimator, &*logger);
+
+		let broadcasted_commit_tx = self.onchain_tx_handler.block_connected(txn_matched, claimable_outpoints, height, &*broadcaster, &*fee_estimator, &*logger);
+		if broadcasted_commit_tx {
+			self.pending_monitor_events.push(MonitorEvent::CommitmentTxBroadcasted(self.funding_info.0));
```
TheBlueMatt, line 1968: I think its right, but it'd be a bit simpler to see this patch is correct (at least doesn't change behavior) if you did this in the would_broadcast_at_height check like the following patch (and drop the one in block_disconnected, too):

```
diff --git a/lightning/src/ln/channelmonitor.rs b/lightning/src/ln/channelmonitor.rs
index 3108889a..5c225300 100644
--- a/lightning/src/ln/channelmonitor.rs
+++ b/lightning/src/ln/channelmonitor.rs
@@ -1934,5 +1934,5 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
                        claimable_outpoints.push(ClaimRequest { absolute_timelock: height, aggregable: false, outpoint: BitcoinOutPoint { txid: self.funding_info.0.txid.clone(), vout: self.funding_info.0.index as u32 }, witness_data: InputMaterial::Funding { funding_redeemscript: self.funding_redeemscript.clone() }});
-               }
-               if should_broadcast {
+                       self.pending_monitor_events.push(MonitorEvent::CommitmentTxBroadcasted(self.funding_info.0));
                        if let Some(commitment_tx) = self.onchain_tx_handler.get_fully_signed_local_tx(&self.funding_redeemscript) {
+                               self.local_tx_signed = true;
                                let (mut new_outpoints, new_outputs, _) = self.broadcast_by_local_state(&commitment_tx, &self.current_local_commitment_tx);
@@ -1966,5 +1966,2 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
                let broadcasted_commit_tx = self.onchain_tx_handler.block_connected(txn_matched, claimable_outpoints, height, &*broadcaster, &*fee_estimator, &*logger);
-               if broadcasted_commit_tx {
-                       self.pending_monitor_events.push(MonitorEvent::CommitmentTxBroadcasted(self.funding_info.0));
-               }
 
```

TheBlueMatt (commented):
Aside from the one clarifying change I suggested and the two comment nits, looks good!

lightning/src/ln/channelmonitor.rs
```diff
@@ -148,6 +148,16 @@ pub enum ChannelMonitorUpdateErr {
 #[derive(Debug)]
 pub struct MonitorUpdateError(pub &'static str);
 
+/// An event to be processed by the ChannelManager.
+#[derive(PartialEq)]
+pub enum MonitorEvent {
```
arik-so, line 153: Out of curiosity, how do we decide which events are monitor events? For example, why is HTLCEvent for claiming one, but forwarding offered HTLCs isn't?

lightning/src/ln/onchaintx.rs
```diff
@@ -666,6 +666,7 @@ impl<ChanSigner: ChannelKeys> OnchainTxHandler<ChanSigner> {
 		let mut new_claims = Vec::new();
 		let mut aggregated_claim = HashMap::new();
 		let mut aggregated_soonest = ::std::u32::MAX;
+		let mut broadcasted = false;
```
arik-so, line 669: The more idiomatic past tense of broadcast is broadcast. However, I think a more easily parsable name would be `did_broadcast`.

lightning/src/ln/channelmonitor.rs
```diff
@@ -1966,7 +1962,11 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
 				}
 			}
 		}
-		self.onchain_tx_handler.block_connected(txn_matched, claimable_outpoints, height, &*broadcaster, &*fee_estimator, &*logger);
+
+		let broadcasted_commit_tx = self.onchain_tx_handler.block_connected(txn_matched, claimable_outpoints, height, &*broadcaster, &*fee_estimator, &*logger);
```
arik-so, line 1966: Sorry, but this name makes it sound like block_connected returns a commitment tx that was broadcast, which had me scratching my head. I think renaming it to `did_broadcast_commitment_tx` would be much easier for readers. Or I'm just dense.

lightning/src/ln/channelmonitor.rs
```diff
@@ -1988,12 +1988,15 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
 			//- maturing spendable output has transaction paying us has been disconnected
 		}
 
-		self.onchain_tx_handler.block_disconnected(height, broadcaster, fee_estimator, logger);
+		let broadcasted_commit_tx = self.onchain_tx_handler.block_disconnected(height, broadcaster, fee_estimator, logger);
```
arik-so, line 1991: same as above

lightning/src/ln/channelmonitor.rs
```diff
@@ -2151,22 +2154,24 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
...
 							source,
 							payment_preimage: Some(payment_preimage),
 							payment_hash
-						});
+						}));
 					}
 				} else if offered_preimage_claim {
-					if !self.pending_htlcs_updated.iter().any(|update| update.source == source) {
+					if !self.pending_monitor_events.iter().any(
+						|update| if let &MonitorEvent::HTLCEvent(ref upd) = update { upd.source == source } else { false }) {
```
arik-so, line 2168: I think this is a bit too much crammed into one line

lightning/src/ln/channelmonitor.rs
```diff
@@ -2422,10 +2427,15 @@ impl<ChanSigner: ChannelKeys + Readable> Readable for (BlockHash, ChannelMonitor
...
-		for _ in 0..pending_htlcs_updated_len {
-			pending_htlcs_updated.push(Readable::read(reader)?);
+		let pending_monitor_events_len: u64 = Readable::read(reader)?;
+		let mut pending_monitor_events = Vec::with_capacity(cmp::min(pending_monitor_events_len as usize, MAX_ALLOC_SIZE / (32 + 8*3)));
+		for _ in 0..pending_monitor_events_len {
+			let ev = match <u8 as Readable>::read(reader)? {
+				0 => MonitorEvent::HTLCEvent(Readable::read(reader)?),
+				1 => MonitorEvent::CommitmentTxBroadcasted(funding_info.0),
+				_ => return Err(DecodeError::InvalidValue)
+			};
```
arik-so, line 2437: One day we'll have versioning for reads lol :)

valentinewallace: Indeed lol

lightning/src/ln/channelmonitor.rs
TheBlueMatt: In this context, its basically "a thing which we detected on-chain which has an impact on our understanding of a channel's state" (ie, basically, "channel has been closed on chain" or "transaction on-chain resolved an HTLC which was outstanding when we went on chain").

lightning/src/ln/channel.rs
```diff
@@ -908,7 +969,7 @@ impl<ChanSigner: ChannelKeys> Channel<ChanSigner> {
 				InboundHTLCState::AwaitingRemoteRevokeToAnnounce(_) => (!generated_by_local, "AwaitingRemoteRevokeToAnnounce"),
```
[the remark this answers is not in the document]
valentinewallace: Open to other options :smile:

TheBlueMatt (approved):

lightning/src/ln/channelmanager.rs
ariard: I think we move it out of `get_and_clear_*` into the top-level API ? Did we have a strong reason to place the monitor query other than not complexifying API, can't remember ?

You may want to call monitor only after each block connection as both events `CommitmentTxBroadcasted/HTLCEvent` are only updated after a block which is not the case of `Event/MessageSendEvent`

TheBlueMatt: Yea, there's no good reason for it aside from not complexifying the current API - adding a new method which users have to poll is terrible, and we can't rely on users calling ChannelMonitor::block_connected before they call ChannelManger::block_connected (which would be required to poll it in block_connected). Not sure where else to put it.

lightning/src/ln/channelmonitor.rs
```diff
@@ -1938,7 +1934,9 @@ impl<ChanSigner: ChannelKeys> ChannelMonitor<ChanSigner> {
 			claimable_outpoints.push(ClaimRequest { absolute_timelock: height, aggregable: false, outpoint: BitcoinOutPoint { txid: self.funding_info.0.txid.clone(), vout: self.funding_info.0.index as u32 }, witness_data: InputMaterial::Funding { funding_redeemscript: self.funding_redeemscript.clone() }});
 		}
 		if should_broadcast {
+			self.pending_monitor_events.push(MonitorEvent::CommitmentTxBroadcasted(self.funding_info.0));
 			if let Some(commitment_tx) = self.onchain_tx_handler.get_fully_signed_local_tx(&self.funding_redeemscript) {
+				self.local_tx_signed = true;
```
ariard, line 1939: @valentinewallace 

I think this is breaking one of the property of our distributed watchtower infrastructure as documented here : https://github.com/rust-bitcoin/rust-lightning/blob/3defcc896266f3d67848ce28981a756e971a3f0c/lightning/src/ln/channelmonitor.rs#L1174

AFAICT, by adding logs flag is true beyond updating local view with latest local state. 

I'll fix it and add coverage for this.

valentinewallace: Logs flag? Ok I think I see what you mean though, thanks for catching this!

TheBlueMatt: We're discussing on IRC still, but I don't think this change is correct - we should refuse to accept updates after we've signed a local tx irrespective of why we signed it.

ariard: Sorry, after discussion on IRC and writing a test case I realized that the alleged property was only supported in my mind. In fact #667 makes it to actually implement it, so did so in #679 with test coverage.

PR #671, opened 2020-08-19
Bump MSRV to 1.30.0

valentinewallace:
We wanted to bump to 1.29 to continue to support mrustc bootstrapping, but on 1.29
there's a bug preventing us from compiling the lightning package only, thus parts
of lightning-net-tokio cause a compilation error.

The advantage of bumping the MSRV is an improved borrow checker which should
enable improved code quality, and not having jump through weird hoops sometimes
to get 1.22 working.

Maybe this is a good compromise in the meantime, and we can always bump more later?

It'd make part of #667 a bit nicer (though admittedly it's possible there's a workaround to my current hack for 1.22, and I'm just too tired of 1.22 to find it lol). I'm OK to revisit bumping MSRV later if it's preferred though.

Related to #471

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,8 @@ jobs:
       matrix:
         toolchain: [ stable,
                      beta,
-                     # 1.22.0 is MSRV for rust-lightning in general:
-                     1.22.0,
+                     # 1.30.0 is MSRV for Rust-Lightning
+                     1.30.0,
                      # 1.34.2 is Debian stable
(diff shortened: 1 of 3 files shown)


elichai:
Is this bug documented anywhere? if not how can I reproduce it locally?

valentinewallace:
> Is this bug documented anywhere? if not how can I reproduce it locally?

I can't find it documented, but to repro just run `cargo +1.29.0 build -p lightning` on the master branch

PR #719, opened 2020-09-24
Fix tx weight calculation

### LINKED ISSUE #630: Ensure tx have a mininal relay fee
Right now we have failure applying this test hardening patch, some of our generated transactions don't have a minimal relay fee.

Fixing this while reworking our fee logic for anchor outputs.

See : https://github.com/TheBlueMatt/rust-lightning/commit/34daef5ae560b984d4dea231ae9759da5a9bfc69

valentinewallace:
Closes #630.


ariard:
fyi @valentinewallace  the scope of #630 is a bunch wider :) There is also all the other standardness rules we should check.

PR #763, opened 2020-12-01
Lightning block sync and sample block sources

jkczyz:
Adds a `lightning-block-sync` crate consisting of the following components:
- `MicroSPVClient` capable of syncing chain listeners from a variety of block sources, including fork detection
- `BlockSource` trait for fetching blocks and headers
- Sample REST and RPC HTTP clients implementing the `BlockSource` trait
- `Poll` trait for polling one or more `BlockSource` implementations for new block data
- Chain pollers implementing the `Poll` trait for both one and multiple blocks sources.

This PR largely builds off #614 by accounting for recent refactoring in #649. It refactors much of that code and adds unit testing.

Looking for feedback from reviewers on the overall API and advice on breaking the code into smaller PRs for ease of review. Some code from #614 either remains untouched and/or may be dropped if not desirable to achieve an MVP.


lightning-block-sync/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+					Ok(_) => {
+						self.chain_tip = ($new_hash, $new_header);
+						blocks_connected = true;
+					},
+				}
+				blocks_connected
+			} }
+		}
+
+		match self.chain_poller.poll_chain_tip(self.chain_tip.1).await {
```
TheBlueMatt, line 426: I'm a little confused by the distinction between chain_poller and the block sources - given many block sources almost certainly will be a part of what we expect to poll for the current best chain, it seems like they shouldn't be separated.

Still, I wonder if we can't move a bunch of the source logic into the Poll trait - instead of multiplexing between sources here, make the Poll trait (maybe renamed to Multiplexer or something) implement `BlockSource` given a list of primary and backup `BlockSource`s. This may also be an opportunity to simplify the initial PR, since, assuming its doable, we could implement MicroSPVClient against a single known block source and not have to do a lot of the tracking we do here, just implementing the chain walking.

TheBlueMatt (commented):
Left a few comments that may be useful for cleanup-direction.

lightning-block-sync/src/http_clients.rs
```diff
@@ -0,0 +1,1080 @@
...
+			 Content-Length: {}\r\n\
+			 \r\n\
+			 {}", uri, host, auth, content.len(), content)?;
+
+		let bytes = read_http_resp(&self.stream, MAX_HTTP_RESPONSE_LEN).await?;
+		F::try_from(bytes)
+	}
+}
+
+async fn read_http_resp(mut socket: &TcpStream, max_resp: usize) -> std::io::Result<Vec<u8>> {
```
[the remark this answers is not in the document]
jkczyz: Yeah, I tend to agree thought didn't get around into investigating this fully. Thanks for pointing out some of the alternatives. Will spend a little time looking into this.

lightning-block-sync/src/http_endpoint.rs
```diff
@@ -0,0 +1,178 @@
+use http;
```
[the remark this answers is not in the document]
TheBlueMatt: > And as we move outward to crates like lightning-block-sync we may want to allow for some others, possibly only in sample implementations they provide like the HTTP clients.

Totally agreed here.

> My argument for using this particular crate is that it is fairly lightweight and lets us focus on domain-specific code.

Yea, if it were more lightweight I may agree, but I think that's the wrong metric - I think the right one would be something that captures the differing costs - how many LoC and of what complexity would it take to write our own, and how many extra dependencies are involved (here its, I believe, three crates just to take http, but it may be more). For JSON, the dependency tree is steep, but JSON is also rather difficult to parse correctly, and easy to screw up. For URI parsing, we're just looking for some separators and returning a few parts, it should be rather difficult to screw that up. As for `read_http_resp`, I'd *love* to replace that with a "real" HTTP crate, but last I heard all the options were downright terrible (https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6 has been updated to indicate its rather out-of-date, but I haven't done any research on what the latest state is).

lightning-block-sync/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+	///
+	/// For sources which cannot find headers based on the hash, returning Transient when
+	/// height_hint is None is fine, though get_best_block() should never return a None for height
+	/// on the same source. Such a source should never be used in init_sync_chain_monitor as it
+	/// doesn't have any initial height information.
+	fn get_header<'a>(&'a mut self, header_hash: &'a BlockHash, height_hint: Option<u32>) -> AsyncBlockSourceResult<'a, BlockHeaderData>;
+
+	/// Gets the block for a given hash. BlockSources may be headers-only, in which case they
+	/// should always return Err(BlockSourceError::Transient) here.
+	fn get_block<'a>(&'a mut self, header_hash: &'a BlockHash) -> AsyncBlockSourceResult<'a, Block>;
```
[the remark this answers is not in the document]
TheBlueMatt: Yes. I presume all clients can be polled in the same manner for headers. Block data I could imagine being a broad mix of potential options - I anticipate eventually you may even have headers-only sources that exist purely to give you a sanity check (eg "if the headers-only source proves I'm missing blocks, also go download blocks from http://nsa.gov cause I need the blocks more than I need privacy from the NSA").

lightning-block-sync/src/http_endpoint.rs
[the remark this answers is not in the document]
jkczyz: > Yea, if it were more lightweight I may agree, but I think that's the wrong metric - I think the right one would be something that captures the differing costs - how many LoC and of what complexity would it take to write our own, and how many extra dependencies are involved (here its, I believe, three crates just to take http, but it may be more). For JSON, the dependency tree is steep, but JSON is also rather difficult to parse correctly, and easy to screw up. For URI parsing, we're just looking for some separators and returning a few parts, it should be rather difficult to screw that up. As for `read_http_resp`, I'd _love_ to replace that with a "real" HTTP crate, but last I heard all the options were downright terrible (https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6 has been updated to indicate its rather out-of-date, but I haven't done any research on what the latest state is).

Yeah, it takes on three dependencies:

```
├── http v0.2.1
│   ├── bytes v0.5.5
│   ├── fnv v1.0.7
│   └── itoa v0.4.6
```

For URI parsing, the implementation that I replaced was not compatible with Rust's `std::net::ToSocketAddrs` for some cases. So there's an argument in how much time we want to invest in making sure the implementation is robust and well tested vs grabbing something off the shelf.

lightning-block-sync/src/http_clients.rs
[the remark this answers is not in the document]
devrandom: in case this does stay with HTTP, why not use one of the HTTP client crates that are already out there?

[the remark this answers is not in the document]
jkczyz: This is definitely a consideration as per https://github.com/rust-bitcoin/rust-lightning/pull/763#discussion_r534527616. From offline discussions, the URI parsing can be removed entirely by simply having the user of `HttpEndpoint` pass the components parts as parameters instead of the URI string. I took a stab at this yesterday with good results. But I still plan on investigating whether there is a suitable, off-the-shelf replacement for `read_http_resp`.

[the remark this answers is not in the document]
devrandom: Sorry, I didn't read Matt's comment carefully enough.

lightning-block-sync/src/http_endpoint.rs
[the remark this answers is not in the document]
TheBlueMatt: As for `read_http_resp`, https://crates.io/crates/chunked_transfer ! Has zero dependencies, too.

lightning-block-sync/src/lib.rs
TheBlueMatt: Right, so we chatted about this a bit offline, but I'm still a bit unsure about the separation here. I went back and reminded myself of how it was done in 614, but I think maybe we want to expand, or at least have more flexibility, around disabling block sources. Right now, this separation (keeping state in poller to figure out when to stop polling a block source) feels very tied to the idea that we'd only ever decide to disable a poller based on its response to a get_best_block() call. In the future, we may decide that a block source providing us garbage blocks is *also* ground for disabling it entirely, at which point this abstraction doesn't make sense anymore.

I see that putting all of the multi-source logic behind a single `BlockSource` multiplexer may take a bit more complexity, but I still think it wouldn't be as bad as you're thinking. It probably even makes sense to replace the recent headers cache with a simple limited-to-100k-entries HashMap or so - its only 100KB and covers a week of headers, then you don't even have to consider the complexity of clearing it when things are synced.

lightning-block-sync/src/http_endpoint.rs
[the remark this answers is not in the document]
jkczyz: Will that crate work with `tokio`? It requires sources to implement `std::io::Read`.

lightning-block-sync/src/lib.rs
jkczyz: The larger question is around where responsibility lies. I prefer to keep the responsibilities narrow as a means to separate concerns. So in this example, a `BlockSource` is only concerned with retrieving chain data. And implementations of `Poll` are only concerned with how to query block sources.

To support the use case that you mentioned, the `Poll` trait can be expanded with a `disable_source` function, which still falls under the concern of "how to query block sources". Then `MicroSPVClient` could signal to disable a block source if it provided garbage blocks. Such decisions may not be possible from within a `BlockSource` alone but rather may require the additional context provided by `MicroSPVClient` (e.g., disabling a source that is on a different chain).

lightning-block-sync/src/http_endpoint.rs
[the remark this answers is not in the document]
devrandom: This one seems more suitable for async stream processing: https://docs.rs/uhttp_chunked_bytes/0.5.0/uhttp_chunked_bytes/ - it can defer when data is not yet available.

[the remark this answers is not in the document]
jkczyz: Nice find! I'll take a stab at using it.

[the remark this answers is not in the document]
jkczyz: > This one seems more suitable for async stream processing: https://docs.rs/uhttp_chunked_bytes/0.5.0/uhttp_chunked_bytes/ - it can defer when data is not yet available.

Turns out `uhttp_chunked_bytes` requires an `Iterator` over the body, which is problematic because iterators cannot be async. We'd have to read the entire stream into memory (assuming the server disconnects) then pass an iterator over the read buffer to the library. Let me know if you had something else in mind.

The same approach can be used with `chunked_transfer`, which has an easier interface to work with, IMHO. Need to add some DoS protection, but I think this should be doable using `AsyncReadExt::take`.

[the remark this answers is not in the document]
devrandom: You are right about `uhttp_chunked_bytes` - I misread the code - it doesn't seem like it can defer in the middle of a read.

About reading the body - isn't the issue that with chunked encoding you don't know how long the body is, so you can't read it fully without doing the decoding as you go along?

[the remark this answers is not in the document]
jkczyz: Right, the approach only works if the server closes the stream, as then the client would get an EOF. Not sure if this is a safe assumption though. Otherwise, the only way of knowing if you hit the end is by doing the decoding, as you said.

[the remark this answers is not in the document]
jkczyz: I ended up writing a simple adapter using `futures::executor::block_on`.

```rust
struct ReadAdapter<'a, R: tokio::io::AsyncRead + std::marker::Unpin>(&'a mut R);

impl<'a, R: tokio::io::AsyncRead + std::marker::Unpin> std::io::Read for ReadAdapter<'a, R> {
       fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
               #[cfg(test)]
               std::thread::yield_now();
               futures::executor::block_on(self.0.read(buf))
       }
}
```

It does the trick, but in order to consistently avoid test deadlocks I had to have the `HttpServer` drop its side of the `TcpStream` after writing the response.

lightning-block-sync/src/http_clients.rs
```diff
@@ -0,0 +1,1080 @@
...
+	let mut chunked = false;
+	// We expect the HTTP headers to fit in 8KB, and use resp as a temporary buffer for headers
+	// until we know our real length.
+	resp.extend_from_slice(&[0; 8192]);
+	'read_headers: loop {
+		if bytes_read >= 8192 {
+			return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, "headers too large"));
+		}
+		bytes_read += read_socket!();
+		for line in resp[..bytes_read].split(|c| *c == '\n' as u8 || *c == '\r' as u8) {
```
devrandom, line 125: It seems to me that this inner loop looks at all lines in the response buffer, even if they are beyond the end of the headers.  So if there's something that looks like a header near the top of the body, it will get confused.

lightning-block-sync/src/http_clients.rs
```diff
@@ -0,0 +1,1080 @@
...
+			let transfer_encoding = b"Transfer-Encoding: ";
+			if line.len() > transfer_encoding.len() && line[..transfer_encoding.len()].eq_ignore_ascii_case(transfer_encoding) {
+				match &*String::from_utf8_lossy(&line[transfer_encoding.len()..]).to_ascii_lowercase() {
+					"chunked" => chunked = true,
+					_ => return Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, "unsupported transfer encoding")),
+				}
+			}
+		}
+		for (idx, window) in resp[..bytes_read].windows(4).enumerate() {
+			if window[0..2] == *b"\n\n" || window[0..2] == *b"\r\r" {
```
devrandom, line 149: this code will not be reached if the body is a single byte.  i.e. `\n\n5` is of length 3, so `windows(4)` will not stop the iteration just before we get to the end of the headers.

lightning-block-sync/src/http_clients.rs
```diff
@@ -0,0 +1,1080 @@
...
+		for line in resp[..bytes_read].split(|c| *c == '\n' as u8 || *c == '\r' as u8) {
+			let content_header = b"Content-Length: ";
+			if line.len() > content_header.len() && line[..content_header.len()].eq_ignore_ascii_case(content_header) {
+				actual_len = match match std::str::from_utf8(&line[content_header.len()..]){
+					Ok(s) => s, Err(e) => return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, e)),
+				}.parse() {
+					Ok(len) => len, Err(e) => return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, e)),
+				};
+			}
+			let http_resp_1 = b"HTTP/1.1 200 ";
```
devrandom, line 134: The status line must be the first line

lightning-block-sync/src/http_clients.rs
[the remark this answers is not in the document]
devrandom: Maybe could steal the `hyper` response reader - https://github.com/hyperium/hyper/blob/master/src/proto/h1/decode.rs

jkczyz:
Pushed changes that:
- remove dependency on `http` crate
- fix some code not compiling with `tokio` feature
- replace custom chunk decoding with `chunked_transfer` crate

@devrandom Thanks for the feedback on the status/header parsing. I'm working on replacing that code as I think it can be simplified considerably.

lightning-block-sync/src/lib.rs
jkczyz: Thought about this some more after our offline discussion. I do see value in having a struct like `BlockSoureMultiplexer` in so far as it would allow users to customize policies around retrying, using backup sources, and generally choosing a source as it sees fit. However, such a multiplexer could accomplish this by implementing _both_ the `BlockSource` and `Poll` traits. Returning itself from `Poll::poll_chain_tip` is effectively how it would multiplex `get_block`. It could even decided to return a single source from `poll_chain_tip` in cases were it decides that is best.

As mentioned above, there is still a need to signal back somehow to say some source returned bad data. This is because there are two checks that `MicroSPVClient` performs that either require its context or it should be relied upon performing:
- [found a different chain](https://github.com/rust-bitcoin/rust-lightning/pull/763/files#diff-f457bab978fc8b89ad308d5195f99d7b65a4a6ba1673c5f164104b2dda9a0db6R183) during fork detection
- [sanity checking block data](https://github.com/rust-bitcoin/rust-lightning/pull/763/files#diff-f457bab978fc8b89ad308d5195f99d7b65a4a6ba1673c5f164104b2dda9a0db6R320) received from a source

The former can only be determined in the context of fork detection. The latter presumably should be handled by `MicroSPVClient` rather than relying on a `BlockSource` to implement this.

Agreed that the caching logic can be simplified which would clean up the `Poll` logic a bit. Backup querying would be removed since the multiplexer would handle it. Otherwise, I'm not sure if there is much else that can be removed from `MicroSPVClient`. Let me know what you think and if I missed something.

TheBlueMatt: As for your two checks, can't the former be detected in some other way? Basically just have the block source know about the chain's genesis or even pass to the block source in calls from the spv client. The second is obviously more nuanced - I'd generally think the block source could handle it, but you end up easily double-checking the data there, which isn't ideal. One design could be to make some container which holds the transaction list and does the checking there - the neat part there is it may line up with returning a list of transactions anyway, by allowing us to have a constructor that takes a full block (and does the checks) or one that takes a pre-filtered list of transactions.

jkczyz: Alright, I managed to massage this into a cleaner design now, IMHO.

`Poll`'s interface has been expanded such that it is now a pure adapter of `BlockSource`, meaning neither it nor `MicroSPVClient` expose `BlockSource` as part of their interface any longer. Instead, functionality needed by `MicroSPVClient` is defined by `Poll`.

`ChainPoller` implements `Poll` for a single `BlockSource`, while `Poll`'s interface is defined in terms of `ValidatedBlockHeader` and `ValidatedBlock`. These essentially encapsulate the checks that `MicroSPVClient` had been performing previously. Therefore, all checks now occur directly within `ChainPoller`.

More elaborate implementations of `Poll` can and must be defined in terms of `ChainPoller`. For example, `ChainMultiplexer` is defined in this manner and no longer needs to implement `BlockSource`. 

Regarding the nuances around transaction lists, I think this can be part of a later refactor after we get a sample node working with a full-block source. Something like what you imagine should work well with the current design with minimal refactoring.

Thoughts?

lightning-block-sync/src/lib.rs
```diff
@@ -0,0 +1,993 @@
...
+	/// Check each source for a new best tip and update the chain listener accordingly.
+	/// Returns true if some blocks were [dis]connected, false otherwise.
+	pub async fn poll_best_tip(&mut self) -> BlockSourceResult<(ChainTip, bool)> {
+		let chain_tip = self.chain_poller.poll_chain_tip(self.chain_tip).await?;
+		let blocks_connected = match chain_tip {
+			ChainTip::Common => false,
+			ChainTip::Better(chain_tip) => {
+				debug_assert_ne!(chain_tip.block_hash, self.chain_tip.block_hash);
+				debug_assert!(chain_tip.chainwork > self.chain_tip.chainwork);
+				self.update_chain_tip(chain_tip).await
```
arik-so: Huge love from me for most any macro->function refactor!

lightning-block-sync/src/lib.rs
```diff
@@ -0,0 +1,1001 @@
...
+			// Walk back the chain, finding blocks needed to connect and disconnect. Only walk back the
+			// header with the greater height, or both if equal heights.
+			let current_height = current.height;
+			let previous_height = previous.height;
+			if current_height <= previous_height {
+				steps.push(ForkStep::DisconnectBlock(previous));
+				previous = self.look_up_previous_header(chain_poller, &previous).await?;
+			}
+			if current_height >= previous_height {
+				steps.push(ForkStep::ConnectBlock(current));
```
arik-so, line 337: are the connect block steps guaranteed to be in the correct chronological order? At first glance, working back from the tip, it appears to me as though the connect steps would end up reversed?

jkczyz: `sync_listener` will first examine the events in order to process disconnected blocks. Then it will examine the events in reverse order (using `rev()` to process connected blocks.

lightning-block-sync/src/lib.rs
```diff
@@ -0,0 +1,1047 @@
...
+	/// attack without giving up privacy by querying a privacy-losing block sources.
+	pub fn init(chain_tip: ValidatedBlockHeader, chain_poller: P) -> Self {
+		let header_cache = HeaderCache::new();
+		let chain_notifier = ChainNotifier { header_cache };
+		Self { chain_tip, chain_poller, chain_notifier }
+	}
+
+	/// Check each source for a new best tip and update the chain listener accordingly.
+	/// Returns true if some blocks were [dis]connected, false otherwise.
+	pub async fn poll_best_tip<CL: ChainListener + Sized>(&mut self, chain_listener: &mut CL) ->
```
TheBlueMatt, line 458: Hmm, why move away from having the `MicroSPVClient` hold a reference to the `ChainListener`? It seems like it could make it easier to screw up tracking of best tip - I believe with the new API you'd need one `MicroSPVClient` per `ChainListener` (or we'd end up connecting blocks only on one listener), but the API doesn't make that obvious (naively, I'd implement the client-side by just calling poll on all the listeners in series). May also make sense to have a `add_sync_listener` which calls `init_sync_listener` as a part of adding a listener to the client.

TheBlueMatt (commented):
New API changes looks great, but one question about the poll function specifically.

lightning-block-sync/src/lib.rs
TheBlueMatt: Looks good to me.

lightning-block-sync/src/lib.rs
jkczyz: > Hmm, why move away from having the `MicroSPVClient` hold a reference to the `ChainListener`? It seems like it could make it easier to screw up tracking of best tip - I believe with the new API you'd need one `MicroSPVClient` per `ChainListener` (or we'd end up connecting blocks only on one listener), but the API doesn't make that obvious (naively, I'd implement the client-side by just calling poll on all the listeners in series).

Very good point. I was considering moving it back for that very reason and also because blocks would be needlessly re-fetched. Can't recall if I moved it to ease testing. Anyhow, I'll move it back.

> May also make sense to have a `add_sync_listener` which calls `init_sync_listener` as a part of adding a listener to the client.

This would imply the client holding multiple listeners vs the user implementing `ChainListener` as needed. For example, having in implementation calling out to both `ChannelManager` and `ChainMonitor`. Haven't put much thought into having a more flexible interface as this, but just wanted to call that out.

TheBlueMatt: > This would imply the client holding multiple listeners vs the user implementing ChainListener as needed. For example, having an implementation calling out to both ChannelManager and ChainMonitor. Haven't put much thought into having a more flexible interface as this, but just wanted to call that out.

Right, yea, either could work fine. In either case it may make sense to pull `init_sync_listener` into the `MicroSPVClient`, I think, though.

jkczyz: > Right, yea, either could work fine. In either case it may make sense to pull `init_sync_listener` into the `MicroSPVClient`, I think, though.

I tried adding a member function that accepts a block source in a similar manner as `ChainPoller` (i.e., `<'b, B: DerefMut<Target=dyn BlockSource + 'b> + Sized + Sync + Send>`. However, attempting to use this with a block source already referenced by `ChainPoller` would result in two mutable borrows. I thought there was a workaround for this. Any advice here?

TheBlueMatt: Right, we'd need to (a) accept the `old_block` blockhash as an argument, (b) poll the poller for how to get to our current tip (but not *the* current tip, it cant change between sources) from `old_block`, and then (c) somehow track the sources that are done as a part of `MicroSPVClient`.

jkczyz: So that requires a bit of refactoring but is doable. The problem I'm having is that -- when writing a the test -- I need to initialize `MicroSPVClient` with a `ChainPoller` that has a mutable reference to a block source and then try to pass the same block source to the new method to do the sync. But that leads to a compilation error because there are two mutable references. I thought there may be some trick with `Arc` / `Mutex` to work around this but couldn't figure it out.

TheBlueMatt: Right, I figured that you wouldn't need to pass the block source to the new `init_sync_listener` - it would just make some request through the `ChainPoller` to get a block source/block data.

jkczyz: Yeah, that's easier, but it would mean possibly fetching through multiple sources if a different implementation of `Poll` is used.

TheBlueMatt: I think we're maybe out of sync here, but no matter - in general using different sources shouldn't matter - they are all "trusted" to give us the right transaction data, presumably. As long as we can (a) iteratively add new listeners and sync them up to the same block in a safe way that makes accidentally syncing them to different blocks hard, and (b) then let the spv client run, keeping all the listeners up to date, we should be good.

jkczyz: Gotcha. I was referring to the comment above `init_sync_listener`:

https://github.com/rust-bitcoin/rust-lightning/blob/eaedc0a8a3cce989aa792daff3737b48201dec23/lightning-block-sync/src/lib.rs#L390-L394

TheBlueMatt: Right, I think that comment was just a general note about how that block source could leave us un-sync-able if it drove us onto some fork and then refused to provide previous headers on the chain to the main chain. Yet another reason to move it into the rest of the spvclient logic - we could use the disconnected-headers caching to ensure that no single block source could do that to us.

PR #771, opened 2021-01-08
Add BOLT 3 test vector for CLTV tiebreaker

jkczyz:
The test vector was added in https://github.com/lightningnetwork/lightning-rfc/pull/539.


jkczyz:
Tested by removing the [tiebreaker](https://github.com/jkczyz/rust-lightning/blob/caa16e0fb549db56e432f84cc902cf1835a8e360/lightning/src/ln/chan_utils.rs#L988) and seeing a test failure.

TheBlueMatt (approved):
Why did we only take one of the new tests when it looks like the spec PR has several tests?

jkczyz:
> Why did we only take one of the new tests when it looks like the spec PR has several tests?

The spec PR only adds one new test vector. However, it was updated to include all test vectors for use with `option_static_remotekey` which we already moved to in babf0af30bfe766e6c3f1c3cf76b7b0fa0830dbc.

PR #791, opened 2021-02-05
SPV client utility for syncing a lightning node

jkczyz:
Completes the `lightning-block-sync` crate with the following components:
- `SpvClient` capable of syncing chain listeners from a variety of block sources, including fork detection
- `Poll` trait for polling one or more `BlockSource` implementations for new block data
- `ChainPoller` for implementing the `Poll` trait for one blocks source

This PR is the second half of #763.

### DIFF
--- a/lightning-block-sync/src/init.rs
+++ b/lightning-block-sync/src/init.rs
@@ -1,0 +1,351 @@
+use crate::{BlockSource, BlockSourceResult, Cache, ChainNotifier};
+use crate::poll::{ChainPoller, Validate, ValidatedBlockHeader};
+
+use bitcoin::blockdata::block::{Block, BlockHeader};
+use bitcoin::hash_types::BlockHash;
+use bitcoin::network::constants::Network;
+
+use lightning::chain;
+
+/// Performs a one-time sync of chain listeners using a single *trusted* block source, bringing each
+/// listener's view of the chain from its paired block hash to `block_source`'s best chain tip.
+///
+/// Upon success, the returned header can be used to initialize [`SpvClient`]. In the case of
+/// failure, each listener may be left at a different block hash than the one it was originally
+/// paired with.
+///
+/// Useful during startup to bring the [`ChannelManager`] and each [`ChannelMonitor`] in sync before
+/// switching to [`SpvClient`]. For example:
+///
+/// ```
+/// use bitcoin::hash_types::BlockHash;
+/// use bitcoin::network::constants::Network;
+///
+/// use lightning::chain;
+/// use lightning::chain::Watch;
+/// use lightning::chain::chainmonitor::ChainMonitor;
+/// use lightning::chain::channelmonitor;
+/// use lightning::chain::channelmonitor::ChannelMonitor;
+/// use lightning::chain::chaininterface::BroadcasterInterface;
+/// use lightning::chain::chaininterface::FeeEstimator;
+/// use lightning::chain::keysinterface;
+/// use lightning::chain::keysinterface::KeysInterface;
+/// use lightning::ln::channelmanager::ChannelManager;
+/// use lightning::ln::channelmanager::ChannelManagerReadArgs;
+/// use lightning::util::config::UserConfig;
+/// use lightning::util::logger::Logger;
+/// use lightning::util::ser::ReadableArgs;
+///
+/// use lightning_block_sync::*;
+///
+/// use std::cell::RefCell;
+/// use std::io::Cursor;
+///
+/// async fn init_sync<
+/// 	B: BlockSource,
+/// 	K: KeysInterface<Signer = S>,
+/// 	S: keysinterface::Sign,
+/// 	T: BroadcasterInterface,
+/// 	F: FeeEstimator,
+/// 	L: Logger,
+/// 	C: chain::Filter,
+/// 	P: channelmonitor::Persist<S>,
+/// >(
+/// 	block_source: &mut B,
+/// 	chain_monitor: &ChainMonitor<S, &C, &T, &F, &L, &P>,
+/// 	config: UserConfig,
+/// 	keys_manager: &K,
+/// 	tx_broadcaster: &T,
+/// 	fee_estimator: &F,
+/// 	logger: &L,
+/// 	persister: &P,
+/// ) {
+/// 	// Read a serialized channel monitor paired with the block hash when it was persisted.
+/// 	let serialized_monitor = "...";
+/// 	let (monitor_block_hash, mut monitor) = <(BlockHash, ChannelMonitor<S>)>::read(
+/// 		&mut Cursor::new(&serialized_monitor), keys_manager).unwrap();
+///
+/// 	// Read the channel manager paired with the block hash when it was persisted.
+/// 	let serialized_manager = "...";
+/// 	let (manager_block_hash, mut manager) = {
+/// 		let read_args = ChannelManagerReadArgs::new(
+/// 			keys_manager,
+/// 			fee_estimator,
+/// 			chain_monitor,
+/// 			tx_broadcaster,
+/// 			logger,
+/// 			config,
+/// 			vec![&mut monitor],
+/// 		);
+/// 		<(BlockHash, ChannelManager<S, &ChainMonitor<S, &C, &T, &F, &L, &P>, &T, &K, &F, &L>)>::read(
+/// 			&mut Cursor::new(&serialized_manager), read_args).unwrap()
+/// 	};
+///
+/// 	// Synchronize any channel monitors and the channel manager to be on the best block.
+/// 	let mut cache = UnboundedCache::new();
+/// 	let mut monitor_listener = (RefCell::new(monitor), &*tx_broadcaster, &*fee_estimator, &*logger);
+/// 	let listeners = vec![
+/// 		(monitor_block_hash, &mut monitor_listener as &mut dyn chain::Listen),
+/// 		(manager_block_hash, &mut manager as &mut dyn chain::Listen),
+/// 	];
+/// 	let chain_tip = init::synchronize_listeners(
+/// 		block_source, Network::Bitcoin, &mut cache, listeners).await.unwrap();
+///
+/// 	// Allow the chain monitor to watch any channels.
+/// 	let monitor = monitor_listener.0.into_inner();
+/// 	chain_monitor.watch_channel(monitor.get_funding_txo().0, monitor);
+///
+/// 	// Create an SPV client to notify the chain monitor and channel manager of block events.
+/// 	let chain_poller = poll::ChainPoller::new(block_source, Network::Bitcoin);
+/// 	let mut chain_listener = (chain_monitor, &manager);
+/// 	let spv_client = SpvClient::new(chain_tip, chain_poller, &mut cache, &chain_listener);
+/// }
+/// ```
+///
+/// [`SpvClient`]: ../struct.SpvClient.html
+/// [`ChannelManager`]: ../../lightning/ln/channelmanager/struct.ChannelManager.html
+/// [`ChannelMonitor`]: ../../lightning/chain/channelmonitor/struct.ChannelMonitor.html
+pub async fn synchronize_listeners<B: BlockSource, C: Cache>(
+	block_source: &mut B,
+	network: Network,
+	header_cache: &mut C,
+	mut chain_listeners: Vec<(BlockHash, &mut dyn chain::Listen)>,
+) -> BlockSourceResult<ValidatedBlockHeader> {
+	let (best_block_hash, best_block_height) = block_source.get_best_block().await?;
+	let best_header = block_source
+		.get_header(&best_block_hash, best_block_height).await?
+		.validate(best_block_hash)?;
+
+	// Fetch the header for the block hash paired with each listener.
+	let mut chain_listeners_with_old_headers = Vec::new();
+	for (old_block_hash, chain_listener) in chain_listeners.drain(..) {
+		let old_header = match header_cache.look_up(&old_block_hash) {
+			Some(header) => *header,
+			None => block_source
+				.get_header(&old_block_hash, None).await?
+				.validate(old_block_hash)?
+		};
+		chain_listeners_with_old_headers.push((old_header, chain_listener))
+	}
+
+	// Find differences and disconnect blocks for each listener individually.
+	let mut chain_poller = ChainPoller::new(block_source, network);
+	let mut chain_listeners_at_height = Vec::new();
+	let mut most_common_ancestor = None;
+	let mut most_connected_blocks = Vec::new();
+	for (old_header, chain_listener) in chain_listeners_with_old_headers.drain(..) {
+		// Disconnect any stale blocks, but keep them in the cache for the next iteration.
+		let header_cache = &mut ReadOnlyCache(header_cache);
+		let (common_ancestor, connected_blocks) = {
+			let chain_listener = &DynamicChainListener(chain_listener);
+			let mut chain_notifier = ChainNotifier { header_cache, chain_listener };
+			let difference =
+				chain_notifier.find_difference(best_header, &old_header, &mut chain_poller).await?;
+			chain_notifier.disconnect_blocks(difference.disconnected_blocks);
+			(difference.common_ancestor, difference.connected_blocks)
+		};
+
+		// Keep track of the most common ancestor and all blocks connected across all listeners.
+		chain_listeners_at_height.push((common_ancestor.height, chain_listener));
+		if connected_blocks.len() > most_connected_blocks.len() {
+			most_common_ancestor = Some(common_ancestor);
+			most_connected_blocks = connected_blocks;
+		}
+	}
+
+	// Connect new blocks for all listeners at once to avoid re-fetching blocks.
+	if let Some(common_ancestor) = most_common_ancestor {
+		let chain_listener = &ChainListenerSet(chain_listeners_at_height);
+		let mut chain_notifier = ChainNotifier { header_cache, chain_listener };
+		chain_notifier.connect_blocks(common_ancestor, most_connected_blocks, &mut chain_poller)
+			.await.or_else(|(e, _)| Err(e))?;
+	}
+
+	Ok(best_header)
+}
+
+/// A wrapper to make a cache read-only.
+///
+/// Used to prevent losing headers that may be needed to disconnect blocks common to more than one
+/// listener.
+struct ReadOnlyCache<'a, C: Cache>(&'a mut C);
+
+impl<'a, C: Cache> Cache for ReadOnlyCache<'a, C> {
+	fn look_up(&self, block_hash: &BlockHash) -> Option<&ValidatedBlockHeader> {
+		self.0.look_up(block_hash)
+	}
+
+	fn block_connected(&mut self, _block_hash: BlockHash, _block_header: ValidatedBlockHeader) {
+		unreachable!()
+	}
+
+	fn block_disconnected(&mut self, _block_hash: &BlockHash) -> Option<ValidatedBlockHeader> {
+		None
+	}
+}
+
+/// Wrapper for supporting dynamically sized chain listeners.
+struct DynamicChainListener<'a>(&'a mut dyn chain::Listen);
+
+impl<'a> chain::Listen for DynamicChainListener<'a> {
+	fn block_connected(&self, _block: &Block, _height: u32) {
+		unreachable!()
+	}
+
+	fn block_disconnected(&self, header: &BlockHeader, height: u32) {
+		self.0.block_disconnected(header, height)
+	}
+}
+
+/// A set of dynamically sized chain listeners, each paired with a starting block height.
+struct ChainListenerSet<'a>(Vec<(u32, &'a mut dyn chain::Listen)>);
+
+impl<'a> chain::Listen for ChainListenerSet<'a> {
+	fn block_connected(&self, block: &Block, height: u32) {
+		for (starting_height, chain_listener) in self.0.iter() {
+			if height > *starting_height {
+				chain_listener.block_connected(block, height);
+			}
+		}
+	}
+
+	fn block_disconnected(&self, _header: &BlockHeader, _height: u32) {
+		unreachable!()
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use crate::test_utils::{Blockchain, MockChainListener};
+	use super::*;
+
+	use bitcoin::network::constants::Network;
+
+	#[tokio::test]
+	async fn sync_from_same_chain() {
+		let mut chain = Blockchain::default().with_height(4);
+
+		let mut listener_1 = MockChainListener::new()
+			.expect_block_connected(*chain.at_height(2))
+			.expect_block_connected(*chain.at_height(3))
+			.expect_block_connected(*chain.at_height(4));
+		let mut listener_2 = MockChainListener::new()
+			.expect_block_connected(*chain.at_height(3))
+			.expect_block_connected(*chain.at_height(4));
+		let mut listener_3 = MockChainListener::new()
+			.expect_block_connected(*chain.at_height(4));
+
+		let listeners = vec![
+			(chain.at_height(1).block_hash, &mut listener_1 as &mut dyn chain::Listen),
+			(chain.at_height(2).block_hash, &mut listener_2 as &mut dyn chain::Listen),
+			(chain.at_height(3).block_hash, &mut listener_3 as &mut dyn chain::Listen),
+		];
+		let mut cache = chain.header_cache(0..=4);
+		match synchronize_listeners(&mut chain, Network::Bitcoin, &mut cache, listeners).await {
+			Ok(header) => assert_eq!(header, chain.tip()),
+			Err(e) => panic!("Unexpected error: {:?}", e),
+		}
+	}
+
+	#[tokio::test]
+	async fn sync_from_different_chains() {
+		let mut main_chain = Blockchain::default().with_height(4);
+		let fork_chain_1 = main_chain.fork_at_height(1);
+		let fork_chain_2 = main_chain.fork_at_height(2);
+		let fork_chain_3 = main_chain.fork_at_height(3);
+
+		let mut listener_1 = MockChainListener::new()
+			.expect_block_disconnected(*fork_chain_1.at_height(4))
+			.expect_block_disconnected(*fork_chain_1.at_height(3))
+			.expect_block_disconnected(*fork_chain_1.at_height(2))
+			.expect_block_connected(*main_chain.at_height(2))
+			.expect_block_connected(*main_chain.at_height(3))
+			.expect_block_connected(*main_chain.at_height(4));
+		let mut listener_2 = MockChainListener::new()
+			.expect_block_disconnected(*fork_chain_2.at_height(4))
+			.expect_block_disconnected(*fork_chain_2.at_height(3))
+			.expect_block_connected(*main_chain.at_height(3))
+			.expect_block_connected(*main_chain.at_height(4));
+		let mut listener_3 = MockChainListener::new()
+			.expect_block_disconnected(*fork_chain_3.at_height(4))
+			.expect_block_connected(*main_chain.at_height(4));
+
+		let listeners = vec![
+			(fork_chain_1.tip().block_hash, &mut listener_1 as &mut dyn chain::Listen),
+			(fork_chain_2.tip().block_hash, &mut listener_2 as &mut dyn chain::Listen),
+			(fork_chain_3.tip().block_hash, &mut listener_3 as &mut dyn chain::Listen),
+		];
+		let mut cache = fork_chain_1.header_cache(2..=4);
+		cache.extend(fork_chain_2.header_cache(3..=4));
+		cache.extend(fork_chain_3.header_cache(4..=4));
+		match synchronize_listeners(&mut main_chain, Network::Bitcoin, &mut cache, listeners).await {
+			Ok(header) => assert_eq!(header, main_chain.tip()),
+			Err(e) => panic!("Unexpected error: {:?}", e),
+		}
+	}
+
+	#[tokio::test]
+	async fn sync_from_overlapping_chains() {
+		let mut main_chain = Blockchain::default().with_height(4);
+		let fork_chain_1 = main_chain.fork_at_height(1);
+		let fork_chain_2 = fork_chain_1.fork_at_height(2);
+		let fork_chain_3 = fork_chain_2.fork_at_height(3);
+
+		let mut listener_1 = MockChainListener::new()
+			.expect_block_disconnected(*fork_chain_1.at_height(4))
+			.expect_block_disconnected(*fork_chain_1.at_height(3))
+			.expect_block_disconnected(*fork_chain_1.at_height(2))
+			.expect_block_connected(*main_chain.at_height(2))
+			.expect_block_connected(*main_chain.at_height(3))
+			.expect_block_connected(*main_chain.at_height(4));
+		let mut listener_2 = MockChainListener::new()
+			.expect_block_disconnected(*fork_chain_2.at_height(4))
+			.expect_block_disconnected(*fork_chain_2.at_height(3))
+			.expect_block_disconnected(*fork_chain_2.at_height(2))
+			.expect_block_connected(*main_chain.at_height(2))
+			.expect_block_connected(*main_chain.at_height(3))
+			.expect_block_connected(*main_chain.at_height(4));
+		let mut listener_3 = MockChainListener::new()
+			.expect_block_disconnected(*fork_chain_3.at_height(4))
+			.expect_block_disconnected(*fork_chain_3.at_height(3))
+			.expect_block_disconnected(*fork_chain_3.at_height(2))
+			.expect_block_connected(*main_chain.at_height(2))
+			.expect_block_connected(*main_chain.at_height(3))
+			.expect_block_connected(*main_chain.at_height(4));
+
+		let listeners = vec![
+			(fork_chain_1.tip().block_hash, &mut listener_1 as &mut dyn chain::Listen),
+			(fork_chain_2.tip().block_hash, &mut listener_2 as &mut dyn chain::Listen),
+			(fork_chain_3.tip().block_hash, &mut listener_3 as &mut dyn chain::Listen),
+		];
+		let mut cache = fork_chain_1.header_cache(2..=4);
+		cache.extend(fork_chain_2.header_cache(3..=4));
+		cache.extend(fork_chain_3.header_cache(4..=4));
+		match synchronize_listeners(&mut main_chain, Network::Bitcoin, &mut cache, listeners).await {
+			Ok(header) => assert_eq!(header, main_chain.tip()),
+			Err(e) => panic!("Unexpected error: {:?}", e),
+		}
+	}
+
+	#[tokio::test]
+	async fn cache_connected_and_keep_disconnected_blocks() {
+		let mut main_chain = Blockchain::default().with_height(2);
+		let fork_chain = main_chain.fork_at_height(1);
+		let new_tip = main_chain.tip();
+		let old_tip = fork_chain.tip();
+
+		let mut listener = MockChainListener::new()
+			.expect_block_disconnected(*old_tip)
+			.expect_block_connected(*new_tip);
+
+		let listeners = vec![(old_tip.block_hash, &mut listener as &mut dyn chain::Listen)];
+		let mut cache = fork_chain.header_cache(2..=2);
+		match synchronize_listeners(&mut main_chain, Network::Bitcoin, &mut cache, listeners).await {
+			Ok(_) => {
+				assert!(cache.contains_key(&new_tip.block_hash));
+				assert!(cache.contains_key(&old_tip.block_hash));
+			},
+			Err(e) => panic!("Unexpected error: {:?}", e),
+		}
+	}
+}
--- a/lightning-block-sync/src/lib.rs
+++ b/lightning-block-sync/src/lib.rs
@@ -1,5 +1,9 @@
 //! A lightweight client for keeping in sync with chain activity.
 //!
+//! Defines an [`SpvClient`] utility for polling one or more block sources for the best chain tip.
+//! It is used to notify listeners of blocks connected or disconnected since the last poll. Useful
+//! for keeping a Lightning node in sync with the chain.
+//!
 //! Defines a [`BlockSource`] trait, which is an asynchronous interface for retrieving block headers
 //! and data.
 //!
@@ -9,6 +13,7 @@
 //! Both features support either blocking I/O using `std::net::TcpStream` or, with feature `tokio`,
 //! non-blocking I/O using `tokio::net::TcpStream` from inside a Tokio runtime.
 //!
+//! [`SpvClient`]: struct.SpvClient.html
 //! [`BlockSource`]: trait.BlockSource.html
 
 #[cfg(any(feature = "rest-client", feature = "rpc-client"))]
@@ -31,7 +36,7 @@ mod test_utils;
 #[cfg(any(feature = "rest-client", feature = "rpc-client"))]
 mod utils;
 
-use crate::poll::{Poll, ValidatedBlockHeader};
+use crate::poll::{ChainTip, Poll, ValidatedBlockHeader};
 
 use bitcoin::blockdata::block::{Block, BlockHeader};
 use bitcoin::hash_types::BlockHash;
@@ -54,9 +59,13 @@ pub trait BlockSource : Sync + Send {
 	/// error.
 	fn get_block<'a>(&'a mut self, header_hash: &'a BlockHash) -> AsyncBlockSourceResult<'a, Block>;
 
-	// TODO: Phrase in terms of `Poll` once added.
-	/// Returns the hash of the best block and, optionally, its height. When polling a block source,
-	/// the height is passed to `get_header` to allow for a more efficient lookup.
+	/// Returns the hash of the best block and, optionally, its height.
+	///
+	/// When polling a block source, [`Poll`] implementations may pass the height to [`get_header`]
+	/// to allow for a more efficient lookup.
+	///
+	/// [`Poll`]: poll/trait.Poll.html
+	/// [`get_header`]: #tymethod.get_header
 	fn get_best_block<'a>(&'a mut self) -> AsyncBlockSourceResult<(BlockHash, Option<u32>)>;
 }
 
@@ -133,6 +142,25 @@ pub struct BlockHeaderData {
 	pub chainwork: Uint256,
 }
 
+/// A lightweight client for keeping a listener in sync with the chain, allowing for Simplified
+/// Payment Verification (SPV).
+///
+/// The client is parameterized by a chain poller which is responsible for polling one or more block
+/// sources for the best chain tip. During this process it detects any chain forks, determines which
+/// constitutes the best chain, and updates the listener accordingly with any blocks that were
+/// connected or disconnected since the last poll.
+///
+/// Block headers for the best chain are maintained in the parameterized cache, allowing for a
+/// custom cache eviction policy. This offers flexibility to those sensitive to resource usage.
+/// Hence, there is a trade-off between a lower memory footprint and potentially increased network
+/// I/O as headers are re-fetched during fork detection.
+pub struct SpvClient<P: Poll, C: Cache, L: ChainListener> {
+	chain_tip: ValidatedBlockHeader,
+	chain_poller: P,
+	chain_notifier: ChainNotifier<C>,
+	chain_listener: L,
+}
+
 /// Adaptor used for notifying when blocks have been connected or disconnected from the chain.
 ///
 /// Used when needing to replay chain data upon startup or as new chain events occur.
@@ -183,6 +211,69 @@ impl Cache for UnboundedCache {
 
 	fn block_disconnected(&mut self, block_hash: &BlockHash) -> Option<ValidatedBlockHeader> {
 		self.remove(block_hash)
+	}
+}
+
+impl<P: Poll, C: Cache, L: ChainListener> SpvClient<P, C, L> {
+	/// Creates a new SPV client using `chain_tip` as the best known chain tip.
+	///
+	/// Subsequent calls to [`poll_best_tip`] will poll for the best chain tip using the given chain
+	/// poller, which may be configured with one or more block sources to query. At least one block
+	/// source must provide headers back from the best chain tip to its common ancestor with
+	/// `chain_tip`.
+	/// * `header_cache` is used to look up and store headers on the best chain
+	/// * `chain_listener` is notified of any blocks connected or disconnected
+	///
+	/// [`poll_best_tip`]: struct.SpvClient.html#method.poll_best_tip
+	pub fn new(
+		chain_tip: ValidatedBlockHeader,
+		chain_poller: P,
+		header_cache: C,
+		chain_listener: L,
+	) -> Self {
+		let chain_notifier = ChainNotifier { header_cache };
+		Self { chain_tip, chain_poller, chain_notifier, chain_listener }
+	}
+
+	/// Polls for the best tip and updates the chain listener with any connected or disconnected
+	/// blocks accordingly.
+	///
+	/// Returns the best polled chain tip relative to the previous best known tip and whether any
+	/// blocks were indeed connected or disconnected.
+	pub async fn poll_best_tip(&mut self) -> BlockSourceResult<(ChainTip, bool)> {
+		let chain_tip = self.chain_poller.poll_chain_tip(self.chain_tip).await?;
+		let blocks_connected = match chain_tip {
+			ChainTip::Common => false,
+			ChainTip::Better(chain_tip) => {
+				debug_assert_ne!(chain_tip.block_hash, self.chain_tip.block_hash);
+				debug_assert!(chain_tip.chainwork > self.chain_tip.chainwork);
+				self.update_chain_tip(chain_tip).await
+			},
+			ChainTip::Worse(chain_tip) => {
+				debug_assert_ne!(chain_tip.block_hash, self.chain_tip.block_hash);
+				debug_assert!(chain_tip.chainwork <= self.chain_tip.chainwork);
+				false
+			},
+		};
+		Ok((chain_tip, blocks_connected))
+	}
+
+	/// Updates the chain tip, syncing the chain listener with any connected or disconnected
+	/// blocks. Returns whether there were any such blocks.
+	async fn update_chain_tip(&mut self, best_chain_tip: ValidatedBlockHeader) -> bool {
+		match self.chain_notifier.synchronize_listener(
+			best_chain_tip, &self.chain_tip, &mut self.chain_poller, &mut self.chain_listener).await
+		{
+			Ok(_) => {
+				self.chain_tip = best_chain_tip;
+				true
+			},
+			Err((_, Some(chain_tip))) if chain_tip.block_hash != self.chain_tip.block_hash => {
+				self.chain_tip = chain_tip;
+				true
+			},
+			Err(_) => false,
+		}
 	}
 }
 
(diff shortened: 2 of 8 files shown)


TheBlueMatt (commented):
Big fan of where the traits ended up. All the below comments are simple code-level things. As for `init_sync_listener`, roughly what I was thinking was the below. Its not ideal - passing in a `ValidatedBlockHeader` instead of a `BlockHash` is probably impractical for users, the `add_listener` method really needs to return a Result and be able to indicate that it failed to find some blocks it needs to disconnect couldn't be found, and the call needs to probably not modify the cache while calling `sync_listener`.  That said, I think the safer API around having multiple listeners (most users will have two - channel and chain monitoring) is nice. There may be some other way to capture this, maybe some wrapper listener object that handles it during setup, I'm not sure.

```
diff --git a/lightning-block-sync/src/lib.rs b/lightning-block-sync/src/lib.rs
index 72332b07..beea7c5a 100644
--- a/lightning-block-sync/src/lib.rs
+++ b/lightning-block-sync/src/lib.rs
@@ -151,11 +151,11 @@ pub struct BlockHeaderData {
 /// custom cache eviction policy. This offers flexibility to those sensitive to resource usage.
 /// Hence, there is a trade-off between a lower memory footprint and potentially increased network
 /// I/O as headers are re-fetched during fork detection.
-pub struct SpvClient<P: Poll, C: Cache, L: ChainListener> {
+pub struct SpvClient<'a, P: Poll, C: Cache> {
        chain_tip: ValidatedBlockHeader,
        chain_poller: P,
        chain_notifier: ChainNotifier<C>,
-       chain_listener: L,
+       chain_listeners: Vec<&'a mut dyn ChainListener>,
 }
 
 /// Adaptor used for notifying when blocks have been connected or disconnected from the chain.
@@ -190,9 +190,9 @@ pub async fn init_sync_listener<CL: ChainListener, B: BlockSource>(
        let old_header = block_source
                .get_header(&old_block, None).await.unwrap()
                .validate(old_block).unwrap();
-       let mut chain_poller = poll::ChainPoller::new(block_source as &mut dyn BlockSource, network);
+       let mut chain_poller = poll::ChainPoller::new(block_source, network);
        let mut chain_notifier = ChainNotifier { header_cache: UnboundedCache::new() };
-       chain_notifier.sync_listener(new_header, &old_header, &mut chain_poller, chain_listener).await.unwrap();
+       chain_notifier.sync_listener(new_header, &old_header, &mut chain_poller, &mut [chain_listener]).await.unwrap();
 }
 
 /// The `Cache` trait defines behavior for managing a block header cache, where block headers are
@@ -230,7 +230,7 @@ impl Cache for UnboundedCache {
        }
 }
 
-impl<P: Poll, C: Cache, L: ChainListener> SpvClient<P, C, L> {
+impl<'a, P: Poll, C: Cache> SpvClient<'a, P, C> {
        /// Creates a new SPV client using `chain_tip` as the best known chain tip.
        ///
        /// Subsequent calls to [`poll_best_tip`] will poll for the best chain tip using the given chain
@@ -245,10 +245,17 @@ impl<P: Poll, C: Cache, L: ChainListener> SpvClient<P, C, L> {
                chain_tip: ValidatedBlockHeader,
                chain_poller: P,
                header_cache: C,
-               chain_listener: L,
        ) -> Self {
                let chain_notifier = ChainNotifier { header_cache };
-               Self { chain_tip, chain_poller, chain_notifier, chain_listener }
+               Self { chain_tip, chain_poller, chain_notifier, chain_listeners: Vec::new() }
+       }
+
+       pub async fn add_listener(&mut self, listener: &'a mut dyn ChainListener, old_header: ValidatedBlockHeader) {
+               if &old_header.block_hash[..] == &[0; 32] { return; } // Remove?
+               if old_header.block_hash == self.chain_tip.block_hash { return; }
+
+               let new_header = self.chain_tip;
+               self.chain_notifier.sync_listener(new_header, &old_header, &mut self.chain_poller, &mut [listener]).await.unwrap();
        }
 
        /// Polls for the best tip and updates the chain listener with any connected or disconnected
@@ -277,7 +284,7 @@ impl<P: Poll, C: Cache, L: ChainListener> SpvClient<P, C, L> {
        /// Updates the chain tip, syncing the chain listener with any connected or disconnected
        /// blocks. Returns whether there were any such blocks.
        async fn update_chain_tip(&mut self, best_chain_tip: ValidatedBlockHeader) -> bool {
-               match self.chain_notifier.sync_listener(best_chain_tip, &self.chain_tip, &mut self.chain_poller, &mut self.chain_listener).await {
+               match self.chain_notifier.sync_listener(best_chain_tip, &self.chain_tip, &mut self.chain_poller, &mut self.chain_listeners).await {
                        Ok(_) => {
                                self.chain_tip = best_chain_tip;
                                true
@@ -315,12 +322,12 @@ impl<C: Cache> ChainNotifier<C> {
        /// disconnected to the fork point. Thus, this may return an `Err` that includes where the tip
        /// ended up which may not be `new_header`. Note that iff the returned `Err` contains `Some`
        /// header then the transition from `old_header` to `new_header` is valid.
-       async fn sync_listener<L: ChainListener, P: Poll>(
+       async fn sync_listener<P: Poll>(
                &mut self,
                new_header: ValidatedBlockHeader,
                old_header: &ValidatedBlockHeader,
                chain_poller: &mut P,
-               chain_listener: &mut L,
+               chain_listener: &mut [&mut dyn ChainListener],
        ) -> Result<(), (BlockSourceError, Option<ValidatedBlockHeader>)> {
                let mut events = self.find_fork(new_header, old_header, chain_poller).await.map_err(|e| (e, None))?;
 
@@ -333,7 +340,9 @@ impl<C: Cache> ChainNotifier<C> {
                                        if let Some(cached_header) = self.header_cache.remove(&header.block_hash) {
                                                assert_eq!(cached_header, *header);
                                        }
-                                       chain_listener.block_disconnected(&header.header, header.height);
+                                       for l in chain_listener.iter_mut() {
+                                               l.block_disconnected(&header.header, header.height);
                                        last_disconnect_tip = Some(header.header.prev_blockhash);
                                },
                                &ForkStep::ForkPoint(ref header) => {
@@ -361,7 +370,9 @@ impl<C: Cache> ChainNotifier<C> {
 
                                println!("Connecting block {}", header.block_hash);
                                self.header_cache.insert(header.block_hash, header);
-                               chain_listener.block_connected(&block, header.height);
+                               for l in chain_listener.iter_mut() {
+                                       l.block_connected(&block, header.height);
                                new_tip = Some(header);
                        }
                }
```

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +138,563 @@
...
+		}
+		Ok(())
+	}
+
+	/// Walks backwards from `current_header` and `prev_header`, finding the common ancestor.
+	/// Returns the steps needed to produce the chain with `current_header` as its tip from the
+	/// chain with `prev_header` as its tip. There is no ordering guarantee between different
+	/// `ForkStep` types, but `DisconnectBlock` and `ConnectBlock` are each returned in
+	/// height-descending order.
+	async fn find_fork<P: Poll>(
```
devrandom, line 376: please see if you like this simplification: https://github.com/rust-bitcoin/rust-lightning/commit/9e07f878f9601340ac050f47b62fdd6f908fdc89

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +138,563 @@ pub struct BlockHeaderData {
...
+///
+/// [`ChainNotifier`]: struct.ChainNotifier.html
+pub trait Cache {
+	/// Retrieves the block header keyed by the given block hash.
+	fn get(&self, block_hash: &BlockHash) -> Option<&ValidatedBlockHeader>;
+
+	/// Inserts a block header keyed by the given block hash.
+	fn insert(&mut self, block_hash: BlockHash, block_header: ValidatedBlockHeader);
+
+	/// Removes the block header keyed by the given block hash.
```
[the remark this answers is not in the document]
jkczyz: Updated the documentation, but preferred to keep the naming as is for some symmetry and to be more aligned with cache terminology. This makes the call sites a bit more readable, IMO.

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +138,563 @@ pub struct BlockHeaderData {
...
+			debug_assert_eq!(tip_header.header.block_hash(), *last_disconnect_tip.as_ref().unwrap());
+		} else {
+			new_tip = Some(*old_header);
+		}
+
+		for event in events.drain(..).rev() {
+			if let ForkStep::ConnectBlock(header) = event {
+				let block = chain_poller
+					.fetch_block(&header).await
+					.or_else(|e| Err((e, new_tip)))?;
```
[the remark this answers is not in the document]
jkczyz: I believe this is handled in `update_chain_tip` as it checks against the previous best chain tip. This behavior is demonstrated by the `poll_from_chain_with_better_tip_and_without_any_new_blocks` unit test.

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +138,563 @@ pub struct BlockHeaderData {
...
+	) -> Self {
+		let chain_notifier = ChainNotifier { header_cache };
+		Self { chain_tip, chain_poller, chain_notifier, chain_listener }
+	}
+
+	/// Polls for the best tip and updates the chain listener with any connected or disconnected
+	/// blocks accordingly.
+	///
+	/// Returns the best polled chain tip relative to the previous best known tip and whether any
+	/// blocks were indeed connected or disconnected.
```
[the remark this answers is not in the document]
jkczyz: If we made `self.chain_tip` available, they could compare against that, which I'm doing in the unit tests. But perhaps that wouldn't be the cleanness API. Would it be worth returning an `Err` to the user in that case? Right now we'll return an error when `Poll::poll_chain_tip` fails but not when any failures originate from `update_chain_tip`. Maybe to ask this a different way, does `poll_best_tip` need to return a bool? What is a user expected to do with it?

jkczyz:
Somehow missed @devrandom's comments. Will also address @TheBlueMatt's top-level comment in the next response.

lightning-block-sync/src/poll.rs
```diff
@@ -0,0 +1,344 @@
...
+
+impl std::ops::Deref for ValidatedBlock {
+	type Target = Block;
+
+	fn deref(&self) -> &Self::Target {
+		&self.inner
+	}
+}
+
+pub struct ChainPoller<'b, B: DerefMut<Target=dyn BlockSource + 'b> + Sized + Sync + Send> {
```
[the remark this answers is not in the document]
TheBlueMatt: Mmm, indeed, it would. I figured a ChainMultiplexer would be a different struct that also implements Poll, which could have its own generics that use DerefMut<dyn BlockSource>. Note that the above error is just a syntax error?

lightning-block-sync/src/lib.rs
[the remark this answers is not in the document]
TheBlueMatt: My point was less about readability and more about implementation behavior - it wouldn't be unreasonable to *not* remove from the cache when we disconnect, but the function naming makes it sound like the cache must do so.

lightning-block-sync/src/lib.rs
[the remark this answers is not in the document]
TheBlueMatt: The bool was, at least at some point, used in the rpcclient crate to determine whether it should check for new ChannelManager/ChainMonitor Events. Its not the most interesting thing in the world, but its somewhat useful.

lightning-block-sync/src/lib.rs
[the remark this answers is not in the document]
jkczyz: Fair point! Renamed.

jkczyz:
> As for `init_sync_listener`, roughly what I was thinking was the below. Its not ideal - passing in a `ValidatedBlockHeader` instead of a `BlockHash` is probably impractical for users, the `add_listener` method really needs to return a Result and be able to indicate that it failed to find some blocks it needs to disconnect couldn't be found, and the call needs to probably not modify the cache while calling `sync_listener`. That said, I think the safer API around having multiple listeners (most users will have two - channel and chain monitoring) is nice. There may be some other way to capture this, maybe some wrapper listener object that handles it during setup, I'm not sure.

Agree that a safer API is ideal. Parameterizing by a single listener which may wrap multiple objects seems the safest since setup would occur before `SpvClient` is initialized. It would also users to determine any ordering guarantees.

The more difficult issue is when needing to initialize multiple `ChannelMonitor`s on setup before passing them over to `ChainMonitor`. Here, each monitor may be at a different block hash. Calling `sync_listener` on each would work but leads to duplicate fetches for headers along the common path. Reusing the cache would help here, but re-fetching common blocks couldn't be avoided.

Some sort of wrapper approach that you suggested may work. The issue that I see is determining which block hash across all `ChannelMonitor`s is the earliest in the chain without knowing the height. Hmmm... then again, we should have the height once we initially fetch the header within `init_sync_listener`. So we could get them to a common ancestor and then proceed to sync in unison going forward.

lightning-block-sync/src/poll.rs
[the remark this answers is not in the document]
TheBlueMatt: Ah! Right, somewhat annoying. I think you could work around it with a holder, something like this, and use SourceHolder for your S. Not sure its worth the headache, though.
```
struct SourceHolder<B: DerefMut<Target=dyn BlockSource>> {
    thing: B
}
impl BlockSource for SourceHolder {
    ...
}
```

TheBlueMatt:
Right, @jkczyz, I didnt intend to imply I'm married to the above design, only suggested it as a sample of could be accomplished with some kind of safe "here's a bucket of things, go figure out how to get them on the same chain tip" API :).

jkczyz:
> Right, @jkczyz, I didnt intend to imply I'm married to the above design, only suggested it as a sample of could be accomplished with some kind of safe "here's a bucket of things, go figure out how to get them on the same chain tip" API :).

Let me know what you think of the new API in 99881bd73eb1375bc250a72a69c8e0bf16767fd2. I'm thinking about updating `SpvClient` to take a `ChainNotifier` rather than a `Cache` so the one passed to `sync_listeners` could be re-used. Either way I should unify the APIs to take one or the other.

Another concern that came out of this is being able to disconnect stale blocks upon restart. Seems either the source needs to maintain these or we would need to persist the header cache across restarts.

TheBlueMatt:
> Let me know what you think of the new API in 99881bd.

Hmm, how does one initialize an SpvClient after calling `sync_listeners` - you get the listeners synced to a common ancestor, but does it tell you what the ancestor is, and how do you initialize an SpvClient to start connecting blocks from that ancestor?

>  I'm thinking about updating SpvClient to take a ChainNotifier rather than a Cache so the one passed to sync_listeners could be re-used. Either way I should unify the APIs to take one or the other.

Either is fine to me.

> Another concern that came out of this is being able to disconnect stale blocks upon restart. Seems either the source needs to maintain these or we would need to persist the header cache across restarts.

Right, I think this is more of something we punt to clients - for the Rest and RPC clients it should be no issue as long as you point them to the same bitcoind as was running previously, but this should be documented. No issue in making the cache serializable, but I feel like its up to the user to decide how they want to tackle it.

jkczyz:
> > Let me know what you think of the new API in [99881bd](https://github.com/rust-bitcoin/rust-lightning/commit/99881bd73eb1375bc250a72a69c8e0bf16767fd2).
> 
> Hmm, how does one initialize an SpvClient after calling `sync_listeners` - you get the listeners synced to a common ancestor, but does it tell you what the ancestor is, and how do you initialize an SpvClient to start connecting blocks from that ancestor?

They are all synced to the passed in `new_block` hash, so the user would initialize `SpvClient` with that. The common ancestor is just an intermediary step that allows `sync_listeners` to know how far back it needs to connect blocks.

BTW, given that `SpvClient` takes a `ValidatedBlockHeader` rather than a `BlockHash`, I may settle on using that in both interfaces. Let me know if you have any concerns. Earlier you had said:

> passing in a ValidatedBlockHeader instead of a BlockHash is probably impractical for users

But I'm not sure that I follow why it is impractical given `sync_listeners` does that immediately. Open to using `BlockHash` instead and hiding the look-up in `SpvClient` if you prefer. It would make user code a little simpler I suppose.

TheBlueMatt:
> They are all synced to the passed in new_block hash, so the user would initialize SpvClient with that. The common ancestor is just an intermediary step that allows sync_listeners to know how far back it needs to connect blocks.

Ah, right, it just still feels really awkward given we could create a type-safe API which enforces that a user cannot end up passing in a different common block hash than what they give to SpvClient to start.

> BTW, given that SpvClient takes a ValidatedBlockHeader rather than a BlockHash, I may settle on using that in both interfaces. Let me know if you have any concerns. Earlier you had said:

I find it really awkward because the API they have to deserialize objects returns a blockhash, but then we expect users to call something to fetch the header and then validate it, when we could otherwise do that for them.

jkczyz:
> Ah, right, it just still feels really awkward given we could create a type-safe API which enforces that a user cannot end up passing in a different common block hash than what they give to SpvClient to start.
> 

Yeah, enforcing a common tip -- rather than common ancestor -- through the API is a different sort of problem. What makes this a bit tricky is that there are two sets of listeners.

1. During setup: `ChannelManager` and a `ChannelMonitor` per channel
2. While polling: `ChannelManager` and `ChainMonitor`

Further, (1) requires using a single trusted `BlockSource` while (2) allows multiple source behind a `Poll` implementation. Note also that fetching an arbitrary header can be done through `BlockSource` but not `Poll`.

So my solution was addressing (1). I could add a utility function that glues (1) and (2) together. It may be a common -- albeit very specific -- sort of use case, so I was hoping to leave it out of `lightning-block-sync` in favor of putting it in the sample node. Then there wouldn't be any dependencies on specific listeners in this crate.

> I find it really awkward because the API they have to deserialize objects returns a blockhash, but then we expect users to call something to fetch the header and then validate it, when we could otherwise do that for them.

Gotcha. I'd imagine the utility mentioned above would manage this for them.

TheBlueMatt:
> Yeah, enforcing a common tip -- rather than common ancestor -- through the API is a different sort of problem. What makes this a bit tricky is that there are two sets of listeners.

Oh, right, duh, I'd totally forgotten about this, that changes things. It does seem like an API that doesn't include knowledge of the specific listeners is ideal, but I wonder if we can still get a similar result. What do you think about passing block hashes into the utility method and then having it return a ValidatedBlockHeader, which users then pass into the construction of the SpvClient to tell it where to start.

jkczyz:
Updated as discussed to return the `ValidatedBlockHeader`. Also, cleaned up the interfaces a bit to accept a reference to a `Cache` so that users can decide on whether it should be persisted.

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,276 @@
+use crate::{BlockSource, BlockSourceResult, Cache, ChainListener, ChainNotifier};
+use crate::poll::{ChainPoller, Validate, ValidatedBlockHeader};
+
+use bitcoin::blockdata::block::{Block, BlockHeader};
+use bitcoin::hash_types::BlockHash;
+use bitcoin::network::constants::Network;
+
+/// Performs a one-time sync of chain listeners using a single *trusted* block source, bringing each
```
TheBlueMatt, line 8: Does it make sense to specify what we mean by "trusted" here? (ie that they could get us stuck out on a fork we cannot disconnect from, though maybe thats not the case with the cache now?) We do validate the headers given from the source, though.

TheBlueMatt:
> Updated as discussed to return the ValidatedBlockHeader.

The API is looking good. I'll do a more careful review of the full code over the weekend or next week.

lightning-block-sync/src/poll.rs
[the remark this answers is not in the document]
jkczyz: Managed to get this working with `ChainMultiplexer` (c3fdc7417e8d303ab5992ecfa0c0d54b00a62065) in slightly different way, so I've updated this branch to remove the `dyn` indirection.

TheBlueMatt (commented):
Only doc comments, I'm happy!

lightning-block-sync/src/init.rs
jkczyz: This has indeed shifted since the introduction of the cache and enforced validation. Are there other reasons why a trusted source is needed? I suppose an untrusted source could not provide enough blocks to get all listeners in sync.

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,267 @@
...
+/// listener's view of the chain from its paired block hash to `block_source`'s best chain tip.
```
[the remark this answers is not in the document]
jkczyz: This currently isn't supported by the utility either since it polls the source for the best tip.

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,267 @@
...
+///
+/// Upon success, the returned header can be used to initialize [`SpvClient`]. In the case of
+/// failure, each listener may be left at a different block hash than the one it was originally
+/// paired with.
+///
+/// Useful during startup to bring the [`ChannelManager`] and each [`ChannelMonitor`] in sync before
+/// switching to [`SpvClient`].
+///
```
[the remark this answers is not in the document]
jkczyz: Excellent idea! This may get a little verbose as I'll need to implement `ChainListener` a few times. Do you think it is worth adding these implementations to the crate? I left them off from the draft but I could see them being useful to others since they would be commonly used adapters.

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +142,547 @@ pub struct BlockHeaderData {
...
+	}
+
+	/// Notifies the chain listeners of disconnected blocks.
+	fn disconnect_blocks<L: ChainListener>(
+		&mut self,
+		mut disconnected_blocks: Vec<ValidatedBlockHeader>,
+		chain_listener: &mut L,
+	) {
+		for header in disconnected_blocks.drain(..) {
+			println!("Disconnecting block {}", header.block_hash);
```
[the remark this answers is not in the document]
jkczyz: Good question. I'm wondering if I should drop the prints entirely and leave the onus on the caller to log inside their listeners. We're likely already doing that when connecting and disconnecting blocks, so the additional logging is probably redundant. It would save piping a logger through as well. Thoughts?

lightning-block-sync/src/poll.rs
```diff
@@ -0,0 +1,341 @@
...
+	/// it is internally consistent.
+	fn validate(self, block_hash: BlockHash) -> BlockSourceResult<Self::T>;
+}
+
+impl Validate for BlockHeaderData {
+	type T = ValidatedBlockHeader;
+
+	fn validate(self, block_hash: BlockHash) -> BlockSourceResult<Self::T> {
+		self.header
+			.validate_pow(&self.header.target())
```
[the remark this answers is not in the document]
jkczyz: Opened a PR upstream.

https://github.com/rust-bitcoin/rust-bitcoin/pull/572

Will add a TODO to update this when there is a new upstream release.

lightning-block-sync/src/poll.rs
```diff
@@ -0,0 +1,341 @@
...
+use std::ops::DerefMut;
+
+/// The `Poll` trait defines behavior for polling block sources for a chain tip and retrieving
+/// related chain data. It serves as an adapter for `BlockSource`.
+pub trait Poll {
+	/// Returns a chain tip in terms of its relationship to the provided chain tip.
+	fn poll_chain_tip<'a>(&'a mut self, best_known_chain_tip: ValidatedBlockHeader) ->
+		AsyncBlockSourceResult<'a, ChainTip>;
+
+	/// Returns the header that preceded the given header in the chain.
```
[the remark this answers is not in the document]
jkczyz: Forgot that I can enforce this by making `Validate` have `pub(crate)` visibility now that `sync_listeners` handles validation. User implementations of `Poll` would need to be implemented in terms of `ChainPoller`. Will update the docs accordingly.

lightning-block-sync/src/lib.rs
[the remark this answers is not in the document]
TheBlueMatt: Yea, that seems fine to me.

lightning-block-sync/src/init.rs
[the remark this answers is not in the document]
TheBlueMatt: Yea, I presume there's little point not having them given the crate already depends on `lightning`. Feel free to do it in the next PR and add docs then.

lightning-block-sync/src/init.rs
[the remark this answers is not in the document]
TheBlueMatt: I guess I was just thinking outloud more than anything. Its something we should eventually support, though. Will open an issue.

lightning-block-sync/src/init.rs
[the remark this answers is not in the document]
jkczyz: FWIW, turns out the crate dependency was only needed for these implementations. It was an oversight to include it in this PR.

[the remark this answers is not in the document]
TheBlueMatt: Ah. I mean I guess we could feature-gate it if we want in the future, but for now I see little reason to.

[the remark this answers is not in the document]
jkczyz: Took a stab at writing an example, which turned out to be a good exercise in seeing how suitable the interface works in practice. See 1d2f5b3d637613904f39a490ba1fdfb34e5ba62a.

I ended up changing `ChainListener`'s methods to take `&self` instead of `&mut self`. This was probably not entirely necessary but did make the code a little cleaner as otherwise the mutable borrows needed to be properly scoped. Definitely open to changing this if you have any strong opinions.

This did require implementing `ChainListener` for `RefCell<ChannelMonitor>` since it needs to call methods that take `&mut self`. I don't have a firm grasp of when trait methods should use borrows vs mutable borrows, but this seems to give some advice:

https://doc.rust-lang.org/std/cell/index.html#when-to-choose-interior-mutability

Ideally, we are at least consist in how we decide to design these interfaces that goes beyond "placates the borrow checker". :P

jkczyz:
Rebased

TheBlueMatt (commented):
Hmmm, I don't think we need to fall back to `RefCell`, we can take `&mut self` to a non-mut reference and it should work fine.
```
impl X for &Y { fn (&mut self) { /* Y is a non-mut ref, but we can modify where it points to */ }
impl X for &mut Y { fn (&mut self) { /* Now Y is mut! */ }
```

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+/// 	let mut manager_listener = &manager;
+/// 	let listeners = vec![
+/// 		(monitor_block_hash, &mut monitor_listener as &mut dyn ChainListener),
+/// 		(manager_block_hash, &mut manager_listener as &mut dyn ChainListener),
+/// 	];
+/// 	let chain_tip =
+/// 		init::sync_listeners(block_source, Network::Bitcoin, &mut cache, listeners).await.unwrap();
+///
+/// 	let monitor = monitor_listener.0.into_inner();
+/// 	chain_monitor.watch_channel(monitor.get_funding_txo().0, monitor);
```
ariard, line 103: As a personal feeling, I would over-comment such doc like "We create a set of listeners. It might be unbounded" or "The monitor has been synced up ! It's ready to react to onchain-event"

Some of our readers might not be Rustaceans and even if this code sounds straightforward to us, always consider the cognitive dissonance :)

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+pub async fn sync_listeners<B: BlockSource, C: Cache>(
+	block_source: &mut B,
+	network: Network,
+	header_cache: &mut C,
+	mut chain_listeners: Vec<(BlockHash, &mut dyn ChainListener)>,
+) -> BlockSourceResult<ValidatedBlockHeader> {
+	let (best_block_hash, best_block_height) = block_source.get_best_block().await?;
+	let new_header = block_source
+		.get_header(&best_block_hash, best_block_height).await?
+		.validate(best_block_hash)?;
```
ariard, line 123: You might document this check is to assert block source internal consistency as it's documented in trait method, otherwise you're wondering about the utility of this check. Not a verification ("Is this data from this distrusted source correct ?") but a consistency check ("Is this data consistent as probed from different endpoints ?").

`new_header` maybe `best_header` to match with hash/height ?

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+
+	// Fetch the header for the block hash paired with each listener.
+	let mut chain_listeners_with_old_headers = Vec::new();
+	for (old_block, chain_listener) in chain_listeners.drain(..) {
```
ariard, line 127: Should say `old_block_hash` ?

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+		let txdata: Vec<_> = block.txdata.iter().enumerate().collect();
+		ChannelManager::block_connected(self, &block.header, &txdata, height);
+	}
+
+	fn block_disconnected(&self, header: &BlockHeader, _height: u32) {
+		ChannelManager::block_disconnected(self, header);
+	}
+}
+
+impl<S, C: Deref, T: Deref, F: Deref, L: Deref, P: Deref> ChainListener for &ChainMonitor<S, C, T, F, L, P>
```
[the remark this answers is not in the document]
jkczyz: Implemented for `Z`? When attempting this for these, I get:

```
error[E0119]: conflicting implementations of trait `ChainListener`:`
```

Were you able to get something to compile?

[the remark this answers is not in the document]
jkczyz: More specifically, after temporarily removing the other two implementations used in the example, I get this error:

```term
error[E0119]: conflicting implementations of trait `ChainListener`:
   --> lightning-block-sync/src/init.rs:246:1
    |
226 | / impl<T, S, M: Deref, B: Deref, K: Deref, F: Deref, L: Deref> ChainListener for T
227 | | where
228 | |     T: Deref<Target = ChannelManager<S, M, B, K, F, L>>,
229 | |     S: keysinterface::Sign,
...   |
243 | |     }
244 | | }
    | |_- first implementation here
245 | 
246 | / impl<T, S, C: Deref, B: Deref, F: Deref, L: Deref, P: Deref> ChainListener for T
247 | | where
248 | |     T: Deref<Target = ChainMonitor<S, C, B, F, L, P>>,
249 | |     S: keysinterface::Sign,
...   |
263 | |     }
264 | | }
    | |_^ conflicting implementation
```

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+	// Find differences and disconnect blocks for each listener individually.
+	let mut chain_poller = ChainPoller::new(block_source, network);
+	let mut chain_listeners_at_height = Vec::new();
+	let mut most_common_ancestor = None;
+	let mut most_connected_blocks = Vec::new();
+	for (old_header, chain_listener) in chain_listeners_with_old_headers.drain(..) {
+		// Disconnect any stale blocks, but keep them in the cache for the next iteration.
+		let header_cache = &mut ReadOnlyCache(header_cache);
+		let mut chain_notifier = ChainNotifier { header_cache };
+		let difference =
```
ariard, line 146: IIUC `find_difference` if the old block header isn't on same chain, we'll try to connect blocks anyway leading to an invalid, mixed chain of block connections from the viewpoint of the listener ?

I think this case is catched-up by `get_header` on block source above, but I think it should be documented in top-down documentation, what a user is expected to do if you hit that kind of failure ?

lightning-block-sync/src/init.rs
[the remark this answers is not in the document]
TheBlueMatt: Ah. I didn't try to implement this suggestion but that makes sense - `Trait<A=B>` is the same type as `Trait<A=C>` (but `Trait<A>` is not the same as `Trait<B>`). You could, however, implement it for `ChainMonitor` and `ChannelManager` and then also for `Deref<Target=B: ChainListener>`.

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +142,545 @@ pub struct BlockHeaderData {
...
+///
+/// The client is parameterized by a chain poller which is responsible for polling one or more block
+/// sources for the best chain tip. During this process it detects any chain forks, determines which
+/// constitutes the best chain, and updates the listener accordingly with any blocks that were
+/// connected or disconnected since the last poll.
+///
+/// Block headers for the best chain are maintained in the parameterized cache, allowing for a
+/// custom cache eviction policy. This offers flexibility to those sensitive to resource usage.
+/// Hence, there is a trade-off between a lower memory footprint and potentially increased network
+/// I/O as headers are re-fetched during fork detection.
```
ariard, line 157: "Note that re-fetching headers from a given height is likely to constitute an observable fingerprint and as such a privacy leak"

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +142,545 @@ pub struct BlockHeaderData {
 	/// of equivalent weight.
 	pub chainwork: Uint256,
 }
+
+/// A lightweight client for keeping a listener in sync with the chain, allowing for Simplified
+/// Payment Verification (SPV).
+///
+/// The client is parameterized by a chain poller which is responsible for polling one or more block
+/// sources for the best chain tip. During this process it detects any chain forks, determines which
+/// constitutes the best chain, and updates the listener accordingly with any blocks that were
```
ariard, line 151: Should say "most-height' instead of best. AFAICT, we don't even do stateless block headers verification ?

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +142,545 @@ pub struct BlockHeaderData {
...
+/// Used by [`ChainNotifier`] to store headers along the best chain, which is important for ensuring
+/// that blocks can be disconnected if they are no longer accessible from a block source (e.g., if
+/// the block source does not store stale forks indefinitely).
+///
+/// Implementations may define how long to retain headers such that it's unlikely they will ever be
+/// needed to disconnect a block.  In cases where block sources provide access to headers on stale
+/// forks reliably, caches may be entirely unnecessary.
+///
+/// [`ChainNotifier`]: struct.ChainNotifier.html
+pub trait Cache {
```
ariard, line 188: Should we parameterized cache size by default to a week of blocks ? IIRC our onchain reorg handling stuff as a reorg security of only 6 blocks...

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +142,545 @@ pub struct BlockHeaderData {
...
+	connected_blocks: Vec<ValidatedBlockHeader>,
+}
+
+impl<'a, C: Cache> ChainNotifier<'a, C> {
+	/// Finds the first common ancestor between `new_header` and `old_header`, disconnecting blocks
+	/// from `old_header` to get to that point and then connecting blocks until `new_header`.
+	///
+	/// Validates headers along the transition path, but doesn't fetch blocks until the chain is
+	/// disconnected to the fork point. Thus, this may return an `Err` that includes where the tip
+	/// ended up which may not be `new_header`. Note that iff the returned `Err` contains `Some`
```
ariard, line 311: typo: iff

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +142,545 @@ pub struct BlockHeaderData {
...
+
+	/// Blocks that were connected to the chain since the last poll.
+	connected_blocks: Vec<ValidatedBlockHeader>,
+}
+
+impl<'a, C: Cache> ChainNotifier<'a, C> {
+	/// Finds the first common ancestor between `new_header` and `old_header`, disconnecting blocks
+	/// from `old_header` to get to that point and then connecting blocks until `new_header`.
+	///
+	/// Validates headers along the transition path, but doesn't fetch blocks until the chain is
```
ariard, line 309: Should say consistency-check to dissociate from valid-headers according to a full node verification logic ?

ariard (commented):
Just a first parse to get in mind the changes, still have to parse `poll.rs`
Great work, I'll try to integrate it to the sample soon :)

lightning-block-sync/src/init.rs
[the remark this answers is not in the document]
jkczyz: Nice! That avoids the conflict but note that when implementing for `Deref` I get this warning:

```
warning: trait objects without an explicit `dyn` are deprecated
   --> lightning-block-sync/src/init.rs:281:42
    |
281 | impl<T: ChainListener> ChainListener for Deref<Target = T> {
    |                                          ^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn Deref<Target = T>`
    |
    = note: `#[warn(bare_trait_objects)]` on by default
```

Which I think we want to avoid doing? If I introduce another type parameter for `Deref` that brings me back to having conflicting implementations.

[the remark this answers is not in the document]
TheBlueMatt: I think you just need another parameter: `impl<D: Deref, T: ChainListener> ChainListener for D where D::Target = T {`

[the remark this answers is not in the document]
jkczyz: Yeah, that's what I was referring to though when I said it brings me back to having a conflicting implementation.

```
error[E0119]: conflicting implementations of trait `ChainListener` for type `lightning::ln::channelmanager::ChannelManager<_, _, _, _, _, _>`:
   --> lightning-block-sync/src/init.rs:281:1
    |
243 | / impl<S, M: Deref, B: Deref, K: Deref, F: Deref, L: Deref> ChainListener for ChannelManager<S, M, B, K, F, L>
244 | | where
245 | |     S: keysinterface::Sign,
246 | |     M::Target: chain::Watch<S>,
...   |
259 | |     }
260 | | }
    | |_- first implementation here
...
281 |   impl<T: ChainListener, U: Deref<Target = T>> ChainListener for U {
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `lightning::ln::channelmanager::ChannelManager<_, _, _, _, _, _>`
    |
    = note: upstream crates may add a new impl of trait `std::ops::Deref` for type `lightning::ln::channelmanager::ChannelManager<_, _, _, _, _, _>` in future versions

error: aborting due to previous error
```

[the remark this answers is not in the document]
jkczyz: Hmmm... so if I'm reading the error correctly, I think we can move `ChainListener` to the `lightning` crate and define the relevant listeners there.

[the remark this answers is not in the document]
TheBlueMatt: Huh. Yes, I think that would fix the issue.

lightning-block-sync/src/lib.rs
arik-so: I think given that these examples are meant to serve as minimum viable demo clients, that is something that can be documented and annotated with a warning somewhere – as long as we remember to do it

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+/// 	let chain_poller = poll::ChainPoller::new(block_source, Network::Bitcoin);
+/// 	let chain_listener = (&chain_monitor, &manager_listener);
+/// 	let spv_client = SpvClient::new(chain_tip, chain_poller, &mut cache, chain_listener);
+/// }
+/// ```
+///
+/// [`SpvClient`]: ../struct.SpvClient.html
+/// [`ChannelManager`]: ../../lightning/ln/channelmanager/struct.ChannelManager.html
+/// [`ChannelMonitor`]: ../../lightning/chain/channelmonitor/struct.ChannelMonitor.html
+pub async fn sync_listeners<B: BlockSource, C: Cache>(
```
arik-so, line 114: one thing about this PR is that there is a lot of async action, and the naming makes things like `sync_listeners` rather ambiguous, because it sounds like the method is returning listeners that aren't async, as opposed to `synchronize_listeners`. I think in some instances, spelling it out might avoid confusion

lightning-block-sync/src/init.rs
arik-so: are you saying if the new chain and the stale chain have no common ancestor? Is that not a rather ridiculously unlikely scenario?

lightning-block-sync/src/poll.rs
```diff
@@ -0,0 +1,356 @@
...
+			return Err(BlockSourceError::persistent("invalid block height"));
+		}
+
+		let work = self.header.work();
+		if self.chainwork != previous_header.chainwork + work {
+			return Err(BlockSourceError::persistent("invalid chainwork"));
+		}
+
+		if let Network::Bitcoin = network {
+			if self.height % 2016 == 0 {
```
arik-so, line 129: I'm kinda surprised difficulty adjustment calculations are part of the sample module. This is certainly something I wouldn't expect third parties to replicate or copy and paste. Ought we create a utility for this? Also, third parties might be much more likely to use external dependencies, and those might have this covered.

arik-so (approved):
Only a couple nits. Looks great to me, all ongoing discussions notwithstanding, especially given it's a sample module.

TheBlueMatt:
> I didn't quite follow what you meant by this. The functions that I need to implement take &self, not &mut self. So there is a compilation error with something like this:

Right, I was suggesting swapping the functions back to take `&mut self`. If you do that, its fine to implement the trait on a `&Struct`, which means the `&mut self` is now a mutable reference to a regular reference, so multiple access of the `&Struct` is allowed.

jkczyz:
> > I didn't quite follow what you meant by this. The functions that I need to implement take &self, not &mut self. So there is a compilation error with something like this:
> 
> Right, I was suggesting swapping the functions back to take `&mut self`. If you do that, its fine to implement the trait on a `&Struct`, which means the `&mut self` is now a mutable reference to a regular reference, so multiple access of the `&Struct` is allowed.

Gotcha. My larger question was around interface design. That is, when should a trait's methods take `&mut self`. I find the answer "whenever an implementation needs to mutate" lacking as presumably there may always be some implementation that requires mutation.

As a different example, a look-up method for a cache shouldn't take `&mut self` even though an LRU cache implementation would require it. Thus, it would need to rely on internal mutability to achieve its goals.

For listening, it seems like a grey area at first. A listener may be passive and simply log events or may need to modify some state based on events. For the passive logging example, the logger may itself need to modify its own internal state. But would a logger interface take `&mut self`? I'd wager not. Further, more generally, whatever is notifying the listeners would need to either own or hold a reference to them, making the interface dictate which type of reference to hold.

So I tend towards choosing `&self` here, but would be interested in any counterpoints to this specific example and to the larger question.

(Note that both `ChannelManager` and `ChainMonitor` use internal mutability which makes it possible to take `&self` for them.)

TheBlueMatt:
Right, I don't think Rust has a really good answer for this - interior mutability means you lose one of rustc's greatest safety features (not to mention runtime checking costs), while, I agree, making everything mut and wrapping in a `Deref` to avoid needing mut is another layer of indirection that makes the API hard to read (not to mention runtime cost of a pointer indirection in some cases). You're sort of damned if you do, damned if you don't here, in general.

In this specific case, I'd vote for the `&mut` option and implementing it on `Deref` here because we expect users to primarily work with `Deref` things anyway - `Arc`s to manager and chainmonitors in the multi-threaded case, and in the init case we enforce single ownership of the `ChannelMonitor` directly in the compiler, which is nice.

jkczyz:
> In this specific case, I'd vote for the `&mut` option and implementing it on `Deref` here because we expect users to primarily work with `Deref` things anyway - `Arc`s to manager and chainmonitors in the multi-threaded case, and in the init case we enforce single ownership of the `ChannelMonitor` directly in the compiler, which is nice.

Ok, in order to get everything working with `Deref`, I needed to parameterize `SpvClient` with `L: Deref where L::Target: ChainListener`. However, IIUC, this means I must us `RefCell` for `ChannelMonitor` because making `ChainListener` methods take `&mut self` would require changing these to use `DerefMut`, which `Arc` does not implement.

I think I touched on most of this offline, but let me know if I'm off on anything.

Also,  I ended up pushing the change through to `ChainNotifier` as it made it a bit cleaner. Renamed `ChainListener` to `Listen` now that it is under `lightning::chain`, too.

TheBlueMatt:
> Ok, in order to get everything working with Deref, I needed to parameterize SpvClient with L: Deref where L::Target: ChainListener

Wait, why? Can't we switch to SpvClient using just T: ChainListener and then implement Chainlistener for &TheThings?

jkczyz:
> > Ok, in order to get everything working with Deref, I needed to parameterize SpvClient with L: Deref where L::Target: ChainListener
> 
> Wait, why? Can't we switch to SpvClient using just T: ChainListener and then implement Chainlistener for &TheThings?

That will work for references but breaks on any other types that implement `Deref` like `Arc`. I could implement it for `Arc`, too, but why not parameterize `SpvClient` with `Deref` just as we do for types like `ChannelManager`? Then it would support any type that implements `Deref`.

lightning-block-sync/src/init.rs
jkczyz: Good call! I meant to do this when cleaning things up but it slipped my mind. Leaving unresolved as a reminder.

lightning-block-sync/src/init.rs
jkczyz: Good call on naming. Regarding `validate`, without calling it you don't have a type that is compatible with the return type.

lightning-block-sync/src/poll.rs
jkczyz: `ChainPoller` is such a utility. In fact, the only we to get validated data is to use it because the `Validate` trait is `pub(crate)`. As such, anyone implementing `Poll` must do so in terms of `ChainPoller`.

jkczyz (commented):
Will get to the remaining comments in the morning.

lightning-block-sync/src/init.rs
jkczyz: `find_difference` will fail in this case as covered by the `sync_from_different_chains` unit test.

lightning-block-sync/src/init.rs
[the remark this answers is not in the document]
jkczyz: More precisely, implementing `ChainListener` for types in the `lightning` crate (e.g., `ChannelManager`).

TheBlueMatt:
> That will work for references but breaks on any other types that implement Deref like Arc. I could implement it for Arc, too, but why not parameterize SpvClient with Deref just as we do for types like ChannelManager? Then it would support any type that implements Deref.

Oops, sorry, I meant implement Listener for Deref types.

jkczyz:
> Oops, sorry, I meant implement Listener for Deref types.

I think I'm already doing so here:

https://github.com/rust-bitcoin/rust-lightning/pull/791/commits/e012c4c6ab8540e3918cb780277c4f7483d75f5a#diff-6dfc691076e612ec16f5f292503bf11fe71edf95700ee7e9a8a67b19ce6719e3R139

But that gives me compilation errors when passing references. If I also implement for `&T`, the example fails to compile when modifying it to use `Rc`. I might be missing something, so if you want to play around with it I made a branch with the change: https://github.com/jkczyz/rust-lightning/commit/1bbfe122e602096e3251e67cafd498430edce88f.

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,427 @@
...
+			}
+		}
+	}
+
+	fn block_disconnected(&self, _header: &BlockHeader, _height: u32) {
+		unreachable!()
+	}
+}
+
+impl<S, B: Deref, F: Deref, L: Deref> ChainListener for (RefCell<ChannelMonitor<S>>, B, F, L)
```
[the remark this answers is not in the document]
jkczyz: Naming suggestions? :)

jkczyz:
> Only a couple nits. Looks great to me, all ongoing discussions notwithstanding, especially given it's a sample module.

Note that this is intended to be used outside of a sample (but not required), though the various traits would likely be customized to something more suitable for the use case.

lightning-block-sync/src/init.rs
ariard: @arik-so a belt-and-suspender if someone screws up between regtest/testnet/mainet ? But like I say already covered, good to me

lightning-block-sync/src/lib.rs
ariard: Let's say you're a lightclient shuting down at block 100, sheltered behind Tor addr XYZ. I'm your BIP157 server and logs the fact you're not downloading anymore headers. Few hours, later you're back with a new tor address ABC, but requests back headers 99, 98 to handle the reorg which did happen meantime. I think it's a hint XYZ == ABC ?

Bookmarking in #565, we already have an issue for this.

lightning-block-sync/src/lib.rs
ariard: "Best chain tip" is really unclear and doesn't say if it's most-work or not, semi-validated or not. If we verify chainwork, I would say "most-work chain tip".

lightning-block-sync/src/lib.rs
ariard: Yes and I'm not sure if it's really worthy for a lightclient to keep the entire header chain. We might perf recommendations later when we know better.

lightning-block-sync/src/lib.rs
jkczyz: The word "best" is used many places throughout this PR and including in identifiers. It would better to define what we mean by "best" in one place if there is ambiguity, IMHO.

lightning-block-sync/src/init.rs
```diff
@@ -0,0 +1,351 @@
...
+/// 	let serialized_manager = "...";
+/// 	let (manager_block_hash, mut manager) = {
+/// 		let read_args = ChannelManagerReadArgs::new(
+/// 			keys_manager,
+/// 			fee_estimator,
+/// 			chain_monitor,
+/// 			tx_broadcaster,
+/// 			logger,
+/// 			config,
+/// 			vec![&mut monitor],
```
valentinewallace, line 78: Hmm @TheBlueMatt is it the case that `ChannelMonitor`s need to be synced to chain *before* being handed to the `ChannelManager` on deserialization? I thought it was but maybe wrong...

TheBlueMatt: Hmm, so the docs for `ChannelManagerReadArgs` seem to say you should reconnect after deserializing the `ChannelManager` (and I tend to trust docs written at the time the code was written more than my recollection). If I had to guess, I'd think it would work ok either way, you just end up with events already ready to be handled by the manager before its even ready to go, but the only thing its supposed to do that requires mut is broadcast using the monitor iff the monitor is out of sync with the updates the manager sent, and the monitor deciding to close itself based on blockchain data is not such an update.

lightning-block-sync/src/lib.rs
```diff
@@ -124,3 +146,562 @@ pub struct BlockHeaderData {
...
+	connected_blocks: Vec<ValidatedBlockHeader>,
+}
+
+impl<'a, C: Cache, L: Deref> ChainNotifier<'a, C, L> where L::Target: chain::Listen {
+	/// Finds the first common ancestor between `new_header` and `old_header`, disconnecting blocks
+	/// from `old_header` to get to that point and then connecting blocks until `new_header`.
+	///
+	/// Validates headers along the transition path, but doesn't fetch blocks until the chain is
+	/// disconnected to the fork point. Thus, this may return an `Err` that includes where the tip
+	/// ended up which may not be `new_header`. Note that if the returned `Err` contains `Some`
```
TheBlueMatt, line 309: the previous use of iff instead of if was not a typo - the inverse is also true, and important to note.

TheBlueMatt (approved):

lightning-block-sync/src/lib.rs
jkczyz: s/inverse/converse :)

Right, I mistakenly misinterpreted this to include the `Ok` case. Switching back.

lightning-block-sync/src/poll.rs
```diff
@@ -0,0 +1,356 @@
...
+use std::ops::DerefMut;
+
+/// The `Poll` trait defines behavior for polling block sources for a chain tip and retrieving
+/// related chain data. It serves as an adapter for `BlockSource`.
+///
+/// [`ChainPoller`] adapts a single `BlockSource`, while any other implementations of `Poll` are
+/// required to be built in terms of it to ensure chain data validity.
+///
+/// [`ChainPoller`]: ../struct.ChainPoller.html
+pub trait Poll {
```
ariard, line 16: Do you have other ideas for trait name ? It's conflicting with return type of `poll_recv` used in the sample, see https://docs.rs/tokio/1.2.0/tokio/sync/mpsc/struct.Receiver.html. 

Or we can just alias the type ?

arik-so (approved):

lightning-block-sync/src/poll.rs
jkczyz: Could you fully-qualify it as `std::task::Poll` instead? It's a pretty short name even when fully qualified. I'd imagine you could break the code up reasonably where they wouldn't need to be used in the same file, as well.

jkczyz:
Squashed!

PR #793, opened 2021-02-05
Unknown bits feature only needs to be validated once

### LINKED ISSUE #792: Unnecessary double validation of unknown bits feature
Browsing the code I found this in `peer_handlers.rs`:

```rust
if msg.features.requires_unknown_bits() {
	log_info!(self.logger, "Peer global features required unknown version bits");
	return Err(PeerHandleError{ no_connection_possible: true }.into());
}
if msg.features.requires_unknown_bits() {
	log_info!(self.logger, "Peer local features required unknown version bits");
	return Err(PeerHandleError{ no_connection_possible: true }.into());
}
```

It seems the same thing is validated twice. Might be leftover from the changes in 31cc243e6f04e82f7754d26df7a2bb0275d7ebd1. Before that features were split in global and local features, but since they're now merged and hence 1 check should be enough.

I'll send a PR shortly.

galderz:
Closes #792.

Removed the redundant validation logic.

### DIFF
--- a/lightning/src/ln/peer_handler.rs
+++ b/lightning/src/ln/peer_handler.rs
@@ -678,11 +678,7 @@ impl<Descriptor: SocketDescriptor, CM: Deref, RM: Deref, L: Deref> PeerManager<D
 			// Setup and Control messages:
 			wire::Message::Init(msg) => {
 				if msg.features.requires_unknown_bits() {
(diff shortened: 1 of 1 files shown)


ariard (commented):
Code Review ACK 048e862

Thanks for cleaning up this!

PR #818, opened 2021-03-02
Add validate_best_block_header utility

jkczyz:
Refactor synchronize_listeners by pulling out a function returning the
validated block header of a BlockSource's best chain tip. This is needed
when a node is started from scratch and has no listeners to sync.


TheBlueMatt (approved):

ariard (commented):
Code Review ACK 60f4daf

PR #836, opened 2021-03-09
Add methods to set invoice features in Features objects.

valentinewallace:
I was struggling to get the `Vec<u5>` from `lightning-invoice` into an our-`features`-compatible `Vec<u8>`. This was much easier. Lmk if it makes sense.

### DIFF
--- a/lightning/src/ln/features.rs
+++ b/lightning/src/ln/features.rs
@@ -32,6 +32,8 @@ use ln::msgs::DecodeError;
 use util::ser::{Readable, Writeable, Writer};
 
 mod sealed {
+	use ln::features::Features;
+
 	/// The context in which [`Features`] are applicable. Defines which features are required and
 	/// which are optional for the context.
 	///
@@ -157,7 +159,8 @@ mod sealed {
 	///
 	/// [`Context`]: trait.Context.html
 	macro_rules! define_feature {
-		($odd_bit: expr, $feature: ident, [$($context: ty),+], $doc: expr) => {
+		($odd_bit: expr, $feature: ident, [$($context: ty),+], $doc: expr, $optional_setter: ident,
+		 $required_setter: ident) => {
 			#[doc = $doc]
 			///
 			/// See [BOLT #9] for details.
@@ -242,6 +245,20 @@ mod sealed {
 				}
 			}
 
+			impl <T: $feature> Features<T> {
+				/// Set this feature as optional.
+				pub fn $optional_setter(mut self) -> Self {
+					<T as $feature>::set_optional_bit(&mut self.flags);
+					self
+				}
+
+				/// Set this feature as required.
+				pub fn $required_setter(mut self) -> Self {
+					<T as $feature>::set_required_bit(&mut self.flags);
+					self
+				}
+			}
+
 			$(
 				impl $feature for $context {
 					// EVEN_BIT % 2 == 0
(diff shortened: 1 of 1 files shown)


jkczyz:
I don't have a problem with this approach. @TheBlueMatt Any reason not to expose these?

TheBlueMatt:
Hmm, our features historically have always been "don't export the manipulation bits, just allows creation via `known` or `empty`. I'm open to changing it, though it would be nice to do something more full-featured than piecemeal.

jkczyz:
> Hmm, our features historically have always been "don't export the manipulation bits, just allows creation via `known` or `empty`. I'm open to changing it, though it would be nice to do something more full-featured than piecemeal.

I tend to like a fluent-style API for this (consuming rather than mutating) similar to what I defined in a test:

https://github.com/rust-bitcoin/rust-lightning/blob/ad09a2f00413b50481975a08aa1315f8ca57291b/lightning/src/ln/features.rs#L673

TheBlueMatt:
> I tend to like a fluent-style API for this (consuming rather than mutating) similar to what I defined in a test:

Yea, definitely makes for cleaner rust code. I need to figure out how to properly map such things in the bindings, though, presumably via a constructor that takes a parameter for each possible function call. Gonna be hard with lightning-invoice.

lightning/src/ln/features.rs
```diff
@@ -508,6 +508,16 @@ impl<T: sealed::VariableLengthOnion> Features<T> {
...
 		<T as sealed::VariableLengthOnion>::supports_feature(&self.flags)
 	}
+
+	/// Set these features as requiring the var_onion_optin feature.
+	pub fn set_supports_var_onion_optin(&mut self) {
+		<T as sealed::VariableLengthOnion>::set_optional_bit(&mut self.flags);
+	}
+
+	/// Set these features as supporting the var_onion_optin feature.
+	pub fn set_requires_var_onion_optin(&mut self) {
```
arik-so, line 518: nit: option

lightning/src/ln/features.rs
```diff
@@ -508,6 +508,16 @@ impl<T: sealed::VariableLengthOnion> Features<T> {
 	pub(crate) fn supports_variable_length_onion(&self) -> bool {
 		<T as sealed::VariableLengthOnion>::supports_feature(&self.flags)
 	}
+
+	/// Set these features as requiring the var_onion_optin feature.
+	pub fn set_supports_var_onion_optin(&mut self) {
```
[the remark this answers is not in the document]
valentinewallace: It's `optin` :) https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md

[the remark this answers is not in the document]
arik-so: Oh lol, nvm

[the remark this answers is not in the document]
jkczyz: Note, in this particular case, we should be consistent in naming these across trait and method names. Here, we deviated a bit from the spec by using `VariableLengthOnion` and `{requires|supports}_variable_length_onion`.

valentinewallace:
Wtf. Why does this keep closing.

lightning/src/ln/features.rs
```diff
@@ -231,6 +233,16 @@ mod sealed {
 				}
 			}
 
+			impl <T: $feature> Features<T> {
+				pub fn $supports_setter(&mut self) {
```
jkczyz, line 249: I think we want to avoid overloading the word "supports" which is used elsewhere to mean has either the optional or required bit set whereas you are using it here to only set the optional bit.

lightning/src/ln/features.rs
```diff
@@ -285,7 +303,8 @@ mod sealed {
 
 	#[cfg(test)]
 	define_feature!(23, UnknownFeature, [TestingContext],
-		"Feature flags for an unknown feature used in testing.");
+		"Feature flags for an unknown feature used in testing.", set_supports_unknown_feature,
+		set_requires_unknown_feature);
```
jkczyz, line 320: Note that elsewhere we already use `set_required_` and `set_optional_` (though enabled only in testing). I'd prefer to standardize on that naming and remove the one-off definitions for this feature.

lightning/src/ln/features.rs
```diff
@@ -242,6 +244,18 @@ mod sealed {
...
+			impl <T: $feature> Features<T> {
+				/// Set this feature as supported.
+				pub fn $supports_setter(&mut self) {
+					<T as $feature>::set_optional_bit(&mut self.flags);
+				}
+
+				/// Set this feature as required.
+				pub fn $requires_setter(&mut self) {
+					<T as $feature>::set_required_bit(&mut self.flags);
+				}
```
jkczyz, line 256: Could you use the fluent-style signature (i.e., consume `mut self` rather than take `&mut self`) as is done in `clear_upfront_shutdown_script`? This allows for stringing these together.

lightning/src/ln/features.rs
```diff
@@ -251,28 +265,34 @@ mod sealed {
...
 				}
 			)*
+
 		}
 	}
 
 	define_feature!(1, DataLossProtect, [InitContext, NodeContext],
-		"Feature flags for `option_data_loss_protect`.");
+		"Feature flags for `option_data_loss_protect`.", set_supports_data_loss_protect,
+		set_requires_data_loss_protect);
```
jkczyz, line 274: It kinda sucks that we have to pass separate names for these. Especially considering we'd have to pass two more if we want to be consistent and define the existing `supports_` and `requires_` predicates using macros, too. Not sure if there is a better solution, though.

jkczyz, line 274: As per my earlier comment, I'd prefer `set_required_` and `set_optional_` prefixes. Or maybe `set_<feature>_required` and `set_<feature>_optional` would read better?

lightning/src/ln/features.rs
valentinewallace: @TheBlueMatt Is this fine for the bindings, or do you wanna put this off for now?

lightning/src/ln/features.rs
valentinewallace: Hm wait, is this fixed? Or did you want something special for this feature?

lightning/src/ln/features.rs
```diff
@@ -242,6 +244,18 @@ mod sealed {
 				}
 			}
 
+			impl <T: $feature> Features<T> {
+				/// Set this feature as supported.
```
jkczyz, line 248: s/supported/optional

lightning/src/ln/features.rs
jkczyz: Ah, never mind. I had thought these may be duplicating `set_required_unknown_bits` and `set_optional_unknown_bits`, but those apply to all contexts not just `TestingContext`.

lightning/src/ln/features.rs
TheBlueMatt: I'd say its fine, it'll result in a bunch of needless `clone`s in Java, but hopefully if we take invoice upstream it'll be rare that people need to use it much anyway.

lightning/src/ln/features.rs
```diff
@@ -32,6 +32,7 @@ use ln::msgs::DecodeError;
 use util::ser::{Readable, Writeable, Writer};
 
 mod sealed {
+	use ln::features::Features;
```
jkczyz, line 35: nit: add new line before docs

lightning/src/ln/features.rs
```diff
@@ -718,4 +741,13 @@ mod tests {
...
 		assert!(!init_features.supports_gossip_queries());
 	}
+
+	#[test]
+	fn set_feature_bits() {
+		let features = InvoiceFeatures::empty()
+			.set_basic_mpp_optional()
+			.set_payment_secret_required();
+		assert!(features.supports_basic_mpp());
+		assert!(features.requires_payment_secret());
```
jkczyz, line 751: Add assertions:

```rust
assert!(!features.requires_basic_mpp());
assert!(features.supports_payment_secret());
```

jkczyz (approved):

lightning/src/ln/features.rs
TheBlueMatt: I think technically you could do a procedural macro for this, but I don't know how to use them and I don't really want to find out.

TheBlueMatt (approved):

PR #861, opened 2021-04-01
De-generify Sign methods

devrandom:
This allows all methods of `Sign` (herein renamed to `BaseSign`) to be non-generic, which allows it to have a vtable and be a `dyn` object.

In general, the following can prevent a trait for being used with `dyn`:

- any of the trait methods is generic
- the trait requires `Sized`, directly or indirectly (e.g. via `Clone`)


devrandom:
Hmm... not sure why the router benchmarks are failing.  They are also failing for me in master.

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,49 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
 }
 
-pub(crate) struct WriterWriteAdaptor<'a, W: Writer + 'a>(pub &'a mut W);
-impl<'a, W: Writer + 'a> Write for WriterWriteAdaptor<'a, W> {
-	fn write_all(&mut self, buf: &[u8]) -> Result<(), ::std::io::Error> {
-		self.0.write_all(buf)
-	}
+pub(crate) struct WriterWriteAdaptor<'a>(pub &'a mut Writer);
+
+impl<'a> Write for WriterWriteAdaptor<'a> {
```
jkczyz, line 69: Is this adaptor still needed?

devrandom: it's used in `impl_consensus_ser!` for serializing bitcoin structs

devrandom: that said, it's better if we implement `Write` directly on `Writer` - will do that now

TheBlueMatt (commented):
Hmmmmm, is there a way to avoid the dropping of `Sign: Clone`? I'm not sure there is a way to avoid requiring `ChannelMonitor: Clone` in the Java bindings (as there is no other way to give java a reference to something in Rust that is only valid for the scope of a function call, which we require for ChannelMonitor), I just haven't spent the time to figure out if I want to upstream that or not :/.

If its really required, maybe there's some other clever way around it - obviously the "Rust-y" way of doing the dyn would be to enum a wrapper, though it requires a few layers of wrappers so may be more challenging here. Or maybe there's a way to have a dynamic clone-but-return-the-boxed-one thing that I'm not aware of (we could always define our own trait for that).

```
enum DynamicSigner {
    A(SignerImplA), B(SignerImplB)
}
impl Sign for DynamicSigner {
  fn f(&self) { match self { A(a) => a.f(), B(b) -> b.f() } }
}
```

devrandom:
> Hmmmmm, is there a way to avoid the dropping of `Sign: Clone`? I'm not sure there is a way to avoid requiring `ChannelMonitor: Clone` in the Java bindings (as there is no other way to give java a reference to something in Rust that is only valid for the scope of a function call, which we require for ChannelMonitor), I just haven't spent the time to figure out if I want to upstream that or not :/.

Might be able to `impl Clone for Box<Signer>`.  Would it be acceptable to Box it in `ChannelMonitor`?  I assume there's no memory allocation concerns because these are somewhat long lived.

TheBlueMatt:
> Might be able to impl Clone for Box<Signer>. Would it be acceptable to Box it in ChannelMonitor? I assume there's no memory allocation concerns because these are somewhat long lived.

Correct, if you are using a `Box<Signer>` that work work, but I think that would be hard, too, but I haven't ever dug into that too deeply.

lightning/src/util/ser_macros.rs
```diff
@@ -116,7 +108,7 @@ macro_rules! decode_tlv {
 macro_rules! impl_writeable {
 	($st:ident, $len: expr, {$($field:ident),*}) => {
 		impl ::util::ser::Writeable for $st {
-			fn write<W: ::util::ser::Writer>(&self, w: &mut W) -> Result<(), ::std::io::Error> {
+			fn write(&self, w: &mut Writer) -> Result<(), ::std::io::Error> {
```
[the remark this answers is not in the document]
devrandom: done

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,46 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
 /// serialization buffer size
 pub const MAX_BUF_SIZE: usize = 64 * 1024;
 
-/// A trait that is similar to std::io::Write but has one extra function which can be used to size
+/// A struct that is similar to std::io::Write but has one extra function which can be used to size
```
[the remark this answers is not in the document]
devrandom: reworded

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,46 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
-pub trait Writer {
+pub struct Writer(pub Vec<u8>);
+
+impl Writer {
+	/// Construct an empty Writer
+	pub fn new() -> Self {
+		Writer(Vec::new())
+	}
+
 	/// Writes the given buf out. See std::io::Write::write_all for more
```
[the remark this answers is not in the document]
devrandom: reworded.

the change to the API should probably be in a followup

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,46 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
-	fn write_all(&mut self, buf: &[u8]) -> Result<(), ::std::io::Error>;
 	/// Hints that data of the given size is about the be written. This may not always be called
 	/// prior to data being written and may be safely ignored.
```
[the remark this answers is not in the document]
devrandom: reorganized

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,46 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
 }
 
-pub(crate) struct WriterWriteAdaptor<'a, W: Writer + 'a>(pub &'a mut W);
-impl<'a, W: Writer + 'a> Write for WriterWriteAdaptor<'a, W> {
-	fn write_all(&mut self, buf: &[u8]) -> Result<(), ::std::io::Error> {
-		self.0.write_all(buf)
-	}
+impl Write for Writer {
 	fn write(&mut self, buf: &[u8]) -> Result<usize, ::std::io::Error> {
 		self.0.write_all(buf)?;
```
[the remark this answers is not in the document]
devrandom: done

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,46 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
 
-/// Writer that only tracks the amount of data written - useful if you need to calculate the length
-/// of some data when serialized but don't yet need the full data.
-pub(crate) struct LengthCalculatingWriter(pub usize);
-impl Writer for LengthCalculatingWriter {
-	#[inline]
 	fn write_all(&mut self, buf: &[u8]) -> Result<(), ::std::io::Error> {
-		self.0 += buf.len();
-		Ok(())
+		self.0.write_all(buf)
```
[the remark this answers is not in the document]
devrandom: done

lightning/src/util/ser.rs
```diff
@@ -625,13 +606,13 @@ impl Readable for PaymentSecret {
 }
 
 impl<T: Writeable> Writeable for Option<T> {
-	fn write<W: Writer>(&self, w: &mut W) -> Result<(), ::std::io::Error> {
+	fn write(&self, w: &mut Writer) -> Result<(), ::std::io::Error> {
 		match *self {
 			None => 0u8.write(w)?,
 			Some(ref data) => {
-				let mut len_calc = LengthCalculatingWriter(0);
+				let mut len_calc = Writer::new();
```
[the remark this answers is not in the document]
devrandom: Implemented using an insert.  Please take a look - we can optimize the 1 byte case, but not sure it's worth it.

BTW, puzzled that in one case the length has 1 added to it.

lightning/src/ln/onion_utils.rs
```diff
@@ -183,10 +183,10 @@ fn shift_arr_right(arr: &mut [u8; ONION_DATA_LEN], amt: usize) {
 pub(super) fn route_size_insane(payloads: &Vec<msgs::OnionHopData>) -> bool {
 	let mut len = 0;
 	for payload in payloads.iter() {
-		let mut payload_len = LengthCalculatingWriter(0);
+		let mut payload_len = Writer::new();
```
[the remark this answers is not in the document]
devrandom: I thought only the legacy onion hop is fixed length?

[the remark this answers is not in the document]
TheBlueMatt: Correct, I was referring there to the idea of adding a function to `Writeable` (or maybe adding a new trait) which allows a caller to ask for the exact size of the object once serialized. We could implement it for a few network messages and use it here.

TheBlueMatt:
Heh, sometimes rust is just really stupid, `BaseSign` looks good, though, can you add docs for it? I think this PR otherwise looks basically good, mod https://github.com/rust-bitcoin/rust-lightning/pull/861#discussion_r607343406

lightning/src/util/ser.rs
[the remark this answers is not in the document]
TheBlueMatt: I figure we should optimize for the 1 byte case, there's no reason not to - its one extra byte insertion that will almost always avoid a vec-shift. As to the 1 - we're really serializing an `Option` - we either write 0 or we write the length, but we may have a `Some` with an object of length zero, so we have to offset to differentiate the two cases.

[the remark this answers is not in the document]
devrandom: Done

lightning/src/ln/onion_utils.rs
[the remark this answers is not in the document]
devrandom: Added to `Writeable` and implemented for the `OnionHopData`.

Anywhere else we could implement it where it would make a significant difference?  I would guess many network messages don't happen often enough to matter.

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,71 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
+	}
+
+	/// Return the number of bytes written
+	pub fn len(&self) -> usize {
+		self.0.len()
+	}
+
+	/// Write the data with a varint length prefix.
+	/// Use length_adjust to add to the length for special cases, such as writing an Option.
+	pub fn write_all_with_length<T: Writeable>(&mut self, data: &T, length_adjust: usize) -> Result<(), ::std::io::Error> {
```
TheBlueMatt, line 68: Hmm, I kinda prefer if we not make this public, I'm not sure if users really should ever be calling this.

lightning/src/util/ser.rs
```diff
@@ -166,18 +172,24 @@ impl<R: Read> Read for ReadTrackingReader<R> {
...
 	/// Writes self out to a Vec<u8>
 	fn encode(&self) -> Vec<u8> {
-		let mut msg = VecWriter(Vec::new());
+		let mut msg = Writer::new();
 		self.write(&mut msg).unwrap();
 		msg.0
 	}
 
+	/// Calculates the encoded length of this object.
+	/// This may be overridden to optimize for specific structs.
```
TheBlueMatt, line 185: Hmmmm, this feels like having two APIs for the same thing - size_hint and encoding_length. size_hint is a "push" and this is a "pull", but given we only ever have one `Writer` interface, I'd think size_hint is the way to go. I'd rather just expose this directly on OnionHopData and maybe add a new (non-public) trait for the test harness stuff.

lightning/src/util/ser.rs
devrandom: done

lightning/src/util/ser.rs
devrandom: done

lightning/src/util/ser.rs
```diff
@@ -747,8 +765,61 @@ impl<A: Readable, B: Readable> Readable for (A, B) {
...
+			// A Vec serializes its own 2 bytes length, so adjust this
+			data.resize(size-3, 0x11u8);
+			// Writing an Option is convenient for round-trip checking of length-prefixed writes
+			Some(data.clone()).write(&mut writer).unwrap();
+			let mut reader = Cursor::new(&writer.0);
+			let read_data: Option<Vec<u8>> = Readable::read(&mut reader).unwrap();
+            assert_eq!(data, read_data.unwrap());
+			// Ensure everything was consumed
+            assert_eq!(reader.position() as usize, writer.len());
+			println!("{}", writer.len())
```
TheBlueMatt, line 822: drop print?

lightning/src/util/ser_macros.rs
```diff
@@ -9,35 +9,23 @@
...
-		}
-
-		BigSize(len.0 as u64).write($stream)?;
-		encode_tlv!($stream, {
+		use util::ser::{BigSize, Writer};
+		let mut buf = Writer::new();
+		encode_tlv!(&mut buf, {
 			$(($type, $field)),*
 		});
+		BigSize(buf.len() as u64).write($stream)?;
```
TheBlueMatt, line 27: Can we not do the same optimization as `encode_tlv` here? At least it would drop one more `memcpy`.

lightning-persister/src/lib.rs
```diff
@@ -143,7 +143,7 @@ impl FilesystemPersister {
 		}
 }
 
-impl<ChannelSigner: Sign + Send + Sync> channelmonitor::Persist<ChannelSigner> for FilesystemPersister {
+impl<ChannelSigner: Sign+Clone> channelmonitor::Persist<ChannelSigner> for FilesystemPersister {
```
TheBlueMatt, line 146: I think you can entirely drop the last commit - the prev commit already removed the `send + Sync` requirement and the last commit just re-adds Clone which is already implied.

TheBlueMatt (commented):
Looks good, just tiny stuff, really. In general, best to word-wrap your commit messages to ~80 chars.

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,71 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
+		let start = self.len();
+		data.write(self).expect("No in-memory data may fail to serialize");
+		self.insert_length(start, length_adjust)
+	}
+
+	fn insert_length(&mut self, slice_start: usize, length_adjust: usize) -> Result<(), ::std::io::Error> {
+		let slice_end = self.len();
+		let length = slice_end - slice_start + length_adjust;
+		let big_size = BigSize(length as u64);
+		let len_len = big_size.expected_length();
```
ariard, line 80: nit: what do you think about "big_size_len" ?

ariard (commented):
Sounds good to me, minor comments

lightning/src/chain/keysinterface.rs
```diff
@@ -738,7 +738,7 @@ impl Readable for InMemorySigner {
 /// Cooperative closes may use seed/2'
 /// The two close keys may be needed to claim on-chain funds!
 pub struct KeysManager {
-	secp_ctx: Secp256k1<secp256k1::SignOnly>,
+	secp_ctx: Secp256k1<secp256k1::All>,
```
jkczyz, line 751: Was this change deliberate? It doesn't seem to go with the commit message.

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,45 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
+		Ok(())
+	}
+
 	/// Hints that data of the given size is about the be written. This may not always be called
 	/// prior to data being written and may be safely ignored.
-	fn size_hint(&mut self, size: usize);
-}
+	pub fn size_hint(&mut self, size: usize) {
+		self.0.reserve(size)
+	}
```
jkczyz, line 59: I was looking into how `reserve` is implemented. It seems it increases the capacity to the _max_ of either the desired capacity or double the current capacity. This makes me wonder how useful `size_hint` really is. 

I suppose we may save some allocations if the writer has 0 capacity, but I'm wondering if the same thing can be accomplished while defining our public interfaces in terms of `std::io::Write` instead of `Writer`. For instance, if a message knows it size, we could do something like:

```rust
Writer::with_capacity(message.size_hint());
```

Rather than having the message call `size_hint` on `Writer`. Thoughts?

lightning/src/chain/keysinterface.rs
```diff
@@ -226,7 +226,7 @@ impl Readable for SpendableOutputDescriptor {
 /// of LN security model, orthogonal of key management issues.
 // TODO: We should remove Clone by instead requesting a new Sign copy when we create
 // ChannelMonitors instead of expecting to clone the one out of the Channel into the monitors.
-pub trait Sign : Send+Clone + Writeable {
+pub trait BaseSign : Send + Writeable {
```
jkczyz, line 229: `BaseSign` seems like a code smell to me. Can't we define `Sign` like this and then use `Sign + Clone` where needed?

lightning/src/ln/msgs.rs
```diff
@@ -887,6 +888,39 @@ mod fuzzy_internal_msgs {
...
+			fn high_zero_bytes_dropped_length(data: u64) -> usize {
+				8 - ((data.leading_zeros()/8) as usize)
+			}
+			match self.format {
+				OnionHopDataFormat::Legacy { .. } => 1 + 8 + 8 + 4 + 12, // 33
+				OnionHopDataFormat::NonFinalNode { .. } => {
+					1 +
+						1 + 1 + high_zero_bytes_dropped_length(self.amt_to_forward) +
+						1 + 1 + high_zero_bytes_dropped_length(self.outgoing_cltv_value as u64) +
+						1 + 1 + 8 // short_channel_id
```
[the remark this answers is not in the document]
devrandom: Theoretically, yes, but in practice `outgoing_cltv_value` needs four bytes on mainnet.

lightning/src/ln/msgs.rs
```diff
@@ -887,6 +888,39 @@ mod fuzzy_internal_msgs {
...
+				}
+				OnionHopDataFormat::FinalNode { ref payment_data } => {
+					1 +
+						1 + 1 + high_zero_bytes_dropped_length(self.amt_to_forward) +
+						1 + 1 + high_zero_bytes_dropped_length(self.outgoing_cltv_value as u64) +
+						match payment_data {
+							None => 0,
+							Some(payment_data) =>
+								1 + 1 + high_zero_bytes_dropped_length(payment_data.total_msat) +
+									32 // payment_secret
```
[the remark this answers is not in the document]
devrandom: Well, `payment_data` would need an `unwrap` in the else branch, which I usually try to avoid if I need to do something in both some/none cases.  Just aesthetics, so happy to change it if LDK is standardizing on the if/else form.

lightning/src/chain/keysinterface.rs
devrandom: Yes.  I'll add to the commit message.

It just doesn't seem worth it to deal with both kinds of contexts separately, given that the caller might as well have an `All` handy to be able to supply both for different calls.

I think it's also not possible to pass in an `All` when `Signing` is expected without generics, which we are trying to get away from here.

lightning/src/chain/keysinterface.rs
devrandom: We had it that way, but @TheBlueMatt had some concerns about complexity / changes to bindings.  It's about 20 spots in the code if I remember correctly.

lightning-persister/src/lib.rs
devrandom: Rebase error.  Done.

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,71 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
+
+impl Writer {
+	/// Construct an empty Writer
+	pub fn new() -> Self {
+		Writer(Vec::new())
+	}
+
 	/// Writes the given buf out. See std::io::Write::write_all for more
-	fn write_all(&mut self, buf: &[u8]) -> Result<(), ::std::io::Error>;
+	pub fn write_all(&mut self, buf: &[u8]) -> Result<(), ::std::io::Error> {
```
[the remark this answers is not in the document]
devrandom: `#[inline]` seems desirable given https://stackoverflow.com/questions/37639276/when-should-inline-be-used-in-rust, so done that.  But it seems strange to force the compiler with `always` rather than give a hint - I doubt the compiler will fail to take advantage unless there's a good reason given the rest of the caller's code.

lightning/src/util/ser.rs
devrandom: We could do `reserve_exact`.

But I am guessing that the caller almost always calls these with a zero sized writer, otherwise it's kind of ineffective.  If this is called with a non-zero sized writer (e.g. if this object is a field in a larger struct), then, yes, the size_hint mechanism won't be very effective and we should think about moving to something like `with_capacity`.

Maybe @TheBlueMatt has more insight about how `size_hint` is used.

lightning/src/util/ser.rs
devrandom: Done

lightning/src/util/ser.rs
devrandom: Done

lightning/src/util/ser.rs
```diff
@@ -747,8 +765,61 @@ impl<A: Readable, B: Readable> Readable for (A, B) {
...
+		for size in vec![3, 0xfc, 0xfd, 0xffff, 0x10000] {
+			let mut writer = Writer::new();
+			let mut data = Vec::new();
+			// A Vec serializes its own 2 bytes length, so adjust this
+			data.resize(size-3, 0x11u8);
+			// Writing an Option is convenient for round-trip checking of length-prefixed writes
+			Some(data.clone()).write(&mut writer).unwrap();
+			let mut reader = Cursor::new(&writer.0);
+			let read_data: Option<Vec<u8>> = Readable::read(&mut reader).unwrap();
+            assert_eq!(data, read_data.unwrap());
```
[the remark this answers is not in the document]
devrandom: Done

lightning/src/util/ser.rs
```diff
@@ -34,65 +34,71 @@ use util::byte_utils::{be64_to_array, be48_to_array, be32_to_array, be16_to_arra
...
-/// A trait that is similar to std::io::Write but has one extra function which can be used to size
-/// buffers being written into.
-/// An impl is provided for any type that also impls std::io::Write which simply ignores size
-/// hints.
+/// A struct that implements a serialization buffer.  A size hint can be given and the length
+/// of the result can be retrieved.
 ///
-/// (C-not exported) as we only export serialization to/from byte arrays instead
-pub trait Writer {
+/// std::io::Write is implemented for serializing Bitcoin objects with impl_consensus_ser! below.
```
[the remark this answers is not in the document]
devrandom: Done

lightning/src/util/ser_macros.rs
devrandom: Ah, good point.  Done with slight improvement to the crate-local interface of `Writer`.

devrandom:
This is ready for another look.

lightning/src/util/ser.rs
TheBlueMatt: Hmm, in  general it probably doesn't make sense to use it for sub-structs. It's mostly intended for where we know the exact size, which we generally do in all our P2P message sending, avoiding copying the data a few times. Luckily, the max Jeff referenced means we can't really do worse than the "normal" behavior with it, though.

TheBlueMatt: As for putting it in messages vs calling it on a Writer, I certainly don't care which we end to with, but it should only be one or the other, and this PR is already nontrivial in size.

devrandom: > Hmm, in general it probably doesn't make sense to use it for sub-structs.

It seems to be getting called for sub-structs, given that it's called on non-empty `Writer` instances during the tests.  Should I open an issue to look into that?

TheBlueMatt: > It seems to be getting called for sub-structs, given that it's called on non-empty Writer instances during the tests. Should I open an issue to look into that?

I mean, I guess its not strictly a bug, though it would be nice to create a sensible estimate at the top of any write methods and call it there, then we could drop it for most sub-structs.

TheBlueMatt (approved):

lightning/src/chain/keysinterface.rs
jkczyz: Gotcha. Figured there might something related to binding. We can remove it easily in the future, so no objections keeping it as is.

lightning/src/util/ser.rs
jkczyz: My main gripe is that we are restricting our interface since any `std::io::Write` can no longer be used, though we can always re-expand later, I suppose.

For size hints, the nice properties of the message returning the size are:

(1) We can use `std::io::Write` in our interfaces without needing another trait.
(2) Size calculations can be encapsulated in sub-structs rather than enclosing structs needing to know how to calculate the sizes of their parts.

TheBlueMatt: > My main gripe is that we are restricting our interface since any std::io::Write can no longer be used, though we can always re-expand later, I suppose.

I'm a little confused, is your question about not wanting to do the concrete-ization in this PR at all, or just that the sizing API isn't great? I think I agree with you on the sizing API, we should change it, but this PR doesn't touch it, so I figure that can come in a followup.

jkczyz: Yeah, the former. The sizing API change can come later. I had thought they may be related if any method taking a `Writer` was also using `size_hint`. But maybe only direct instantiators of `Writer` (`VecWriter` before) use it?

So in this PR, I had figured `Writer` would implement `std::io::Write` so the interface changes would therefore be limited to changing traits and not break any potential users. But maybe there is something related to generifying `Sign` that I'm missing. It's not immediately clear to me how/if these two things are connected. Could use some clarification in the PR description if so.

devrandom: I updated the PR description.

The relevant method in `Writeable` is `write(&self, writer: &mut Writer)`.  If we want to use a trait (e.g. `Write`) instead of the concrete `Writer`, that would have to change to either:

- `write<W: Write>(&self, writer: &mut W)` - this is basically back to the previous code, which would cause `Sign` to have a generic method, and therefore not be able to be `dyn`
- `write(&self, writer: &mut dyn Write)` - this is possible, with a slight performance impact

jkczyz: Ah, is this because `ChannelMonitor` needs to be `Writeable` and thus so does `Sign`? Couldn't we parameterize as `ChannelMonitor<Signer: Sign + Writeable>`?

TheBlueMatt: > Couldn't we parameterize as ChannelMonitor<Signer: Sign + Writeable>?

Yes, but ultimately, as far as I understand, that doesn't solve the issue - as soon as you `Box<dyn Trait>` you can't have generics on the individual method definitions. In theory we could do `Sign<Writer>`, but of course then you're stuck with one writer at compile-time and can't have multiple. It may be possible to do that and then pick `dyn Writer` at compile time, but that would drop a *ton* of optimization we're basically relying on for users who want multiple-different-writer support.

Another approach which doesn't "solve" the problem, but may help work around it, is we could move the `write` on `Sign` to `KeysManager` (ie have a `KeysManager::write_sign()` method), which would allow for a `dyn Sign`, but not a `dyn KeysManager`.

devrandom: > Yes, but ultimately, as far as I understand, that doesn't solve the issue - as soon as you Box<dyn Trait> you can't have generics on the individual method definitions

You could do `Box<dyn Sign>` but not `Box<dyn Sign+Writeable>`.  So that would still work for having multiple `Sign` implementations at runtime (e.g. dispatched via a `DynSigner` as discussed elsewhere), but they wouldn't be implementing their own serialization (`Writeable`).

TheBlueMatt: > but they wouldn't be implementing their own serialization (Writeable).

Right, so we'd have to do my second suggestion, above, and move the serialization to the KeysManager in some way?

TheBlueMatt: Right, presumably via KeysManager or some `Sign`-specific write API that does what this PR does but only for `Sign`. Thoughts, @jkczyz ?

jkczyz: @TheBlueMatt Not sure I understand how that would look with regards to a `ChannelMonitor` serializing its parameterized `Sign`.

@devrandom Could `DynSigner` also have a field storing something like `FnOnce` that downcasts the boxed signer and calls write. Then `DynSigner` could implement `Writeable` by calling this. The instantiator of `DynSigner` would need to give this function, but presumably that is possible since they are also giving boxed signer and thus know what type to downcast to.

TheBlueMatt: > @TheBlueMatt Not sure I understand how that would look with regards to a ChannelMonitor serializing its parameterized Sign.

Right, I guess there's a few options - take a reference to the `KeysManager` and call `KeysManager::serialize_signer(sign: &Sign)`. Or we could go with a straightforward `Sign::serialize() -> Vec<u8>` only.

devrandom: > @devrandom Could `DynSigner` also have a field storing something like `FnOnce` that downcasts the boxed signer and calls write? Then `DynSigner` could implement `Writeable` by calling this. The instantiator of `DynSigner` would need to give this function, but presumably that is possible since they are also giving boxed signer and thus know what type to downcast to.

Yes, that works.  There are actually more options, since a generic `Write` can be converted to a `dyn Write` inside `DynSigner`.

But in any case, moving `Writeable` from `BaseSign` to `Sign` is definitely a valid approach.  I am not worried about the sunk cost of having rewritten the `Writeable` trait, happy to back that out.  Should decide whether or not to do that based on whether we actually expect some advantage from not going via a Vec intermediary serialization.

devrandom: Here's a branch with the above implemented: https://github.com/rust-bitcoin/rust-lightning/compare/main...lightning-signer:degenerify1?expand=1

TheBlueMatt: > Yes, that works.

Huh! I guess when the limits the language is applying are arbitrary, the workarounds are arbitrary too :).

That looks good to me, too, what do you think @jkczyz?

jkczyz: > Huh! I guess when the limits the language is applying are arbitrary, the workarounds are arbitrary too :).
> 
> That looks good to me, too, what do you think @jkczyz?

Works for me! Would be nice to move `Writeable` off to where it is needed rather than making it a sub-trait, if that's possible. Wasn't quite clear if the workaround allowed for that. But either way seems clean enough.

> I am not worried about the sunk cost of having rewritten the `Writeable` trait, happy to back that out.

@devrandom Regarding `Writer` trait, I do like making that a struct and using `std::io::Write`, but that is no longer applicable to this PR and can come at some later time, as you said. I think you meant `Writer` here, at least.

devrandom:
OK, the changes to `Writer` / `Writeable` have been taken out, and Writeable is now part of `Sign` instead of `BaseSign`.  This was as a result of the discussion in https://github.com/rust-bitcoin/rust-lightning/pull/861#discussion_r610472700.

This is ready for another review.

ariard:
Code Review ACK 0821fb3

lightning/src/chain/keysinterface.rs
```diff
@@ -1048,3 +1048,9 @@ impl KeysInterface for KeysManager {
 		InMemorySigner::read(&mut std::io::Cursor::new(reader))
 	}
 }
+
+// Ensure that BaseSign can have a vtable
+#[test]
+pub fn dyn_sign() {
+	let _signer: Box<dyn BaseSign>;
+}
```
jkczyz, line 1056: nit: add newline

jkczyz (approved):

PR #863, opened 2021-04-02
persister: Expose method to read ChannelMonitors from disk

valentinewallace:
Tested w/ the sample.

### DIFF
--- a/lightning-persister/src/lib.rs
+++ b/lightning-persister/src/lib.rs
@@ -12,7 +12,8 @@ extern crate lightning;
 extern crate bitcoin;
 extern crate libc;
 
-use bitcoin::hashes::hex::ToHex;
+use bitcoin::{BlockHash, Txid};
+use bitcoin::hashes::hex::{FromHex, ToHex};
 use crate::util::DiskWriteable;
 use lightning::chain;
 use lightning::chain::chaininterface::{BroadcasterInterface, FeeEstimator};
@@ -22,20 +23,13 @@ use lightning::chain::keysinterface::{Sign, KeysInterface};
 use lightning::chain::transaction::OutPoint;
 use lightning::ln::channelmanager::ChannelManager;
 use lightning::util::logger::Logger;
-use lightning::util::ser::Writeable;
+use lightning::util::ser::{ReadableArgs, Writeable};
+use std::collections::HashMap;
 use std::fs;
-use std::io::Error;
-use std::path::PathBuf;
+use std::io::{Cursor, Error};
+use std::ops::Deref;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
-
-#[cfg(test)]
-use {
-	lightning::util::ser::ReadableArgs,
-	bitcoin::{BlockHash, Txid},
-	bitcoin::hashes::hex::FromHex,
-	std::collections::HashMap,
-	std::io::Cursor
-};
 
 /// FilesystemPersister persists channel data on disk, where each channel's
 /// data is stored in a file named after its funding outpoint.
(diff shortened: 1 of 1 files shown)


lightning-persister/src/lib.rs
```diff
@@ -108,39 +102,64 @@ impl FilesystemPersister {
...
+				));
+			}
 
-				if let Ok((_, loaded_monitor)) =
-					<(BlockHash, ChannelMonitor<Keys::Signer>)>::read(&mut Cursor::new(&contents.unwrap()), keys) {
-						res.insert(OutPoint { txid: txid.unwrap(), index: index.unwrap() }, loaded_monitor);
-					} else {
-						return Err(ChannelMonitorUpdateErr::PermanentFailure);
-					}
+			let index = filename.unwrap().split_at(65).1.split('.').next().unwrap().parse();
```
TheBlueMatt, line 135: nit: Wait, we don't actually generate files with `.*` at the end, do we? I guess we could, but don't need the split, I guess?

TheBlueMatt (approved):

lightning-persister/src/lib.rs
```diff
@@ -108,39 +102,64 @@ impl FilesystemPersister {
...
+					outpoint_to_channelmonitor.insert(
+						OutPoint { txid: txid.unwrap(), index: index.unwrap() },
+						(blockhash, channel_monitor),
+					);
+				} else {
+					return Err(std::io::Error::new(
+						std::io::ErrorKind::Other,
+						"Failed to deserialize ChannelMonitor",
+					));
+				}
```
jkczyz, line 159: Consider using `match` along with a local variable as it is a bit more compact:

```rust
let buffer = Cursor::new(&contents);
match <(BlockHash, ChannelMonitor<Signer>)>::read(&mut buffer, &*keys_manager) {
}
```

lightning-persister/src/lib.rs
```diff
@@ -108,39 +102,64 @@ impl FilesystemPersister {
...
+			return Ok(HashMap::new());
+		}
+		let mut outpoint_to_channelmonitor = HashMap::new();
+		for file_option in fs::read_dir(path).unwrap() {
+			let file = file_option.unwrap();
+			let owned_file_name = file.file_name();
+			let filename = owned_file_name.to_str();
+			if !filename.is_some() || !filename.unwrap().is_ascii() || filename.unwrap().len() < 65 {
+				return Err(std::io::Error::new(
+					std::io::ErrorKind::Other,
```
jkczyz, line 122: `InvalidData` would be a more suitable error kind here and elsewhere.

lightning-persister/src/lib.rs
```diff
@@ -108,39 +102,64 @@ impl FilesystemPersister {
...
-				return Err(ChannelMonitorUpdateErr::PermanentFailure);
+	/// Read `ChannelMonitor`s from disk.
+	pub fn read_channelmonitors<Signer: Sign, K: Deref> (
+		&self, keys_manager: K
+	) -> Result<HashMap<OutPoint, (BlockHash, ChannelMonitor<Signer>)>, std::io::Error>
+		where K::Target: KeysInterface<Signer=Signer> + Sized
+	{
+		let path = self.path_to_monitor_data();
+		if !Path::new(&path).exists() {
+			return Ok(HashMap::new());
```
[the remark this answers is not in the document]
TheBlueMatt: Hmm, isn't it more likely its just the first startup and thus there's no monitors?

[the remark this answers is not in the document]
jkczyz: I was thinking that the caller would handle creating the directory. But I guess that `FilesystemPersister` creates it on demand, so probably ok then. Wasn't sure if we want to give the user insight into this (e.g., they instantiate a persister expecting to find data but don't because the path was wrong).

lightning-persister/src/lib.rs
valentinewallace: Hmm, I think the custom error messages are more helpful?

jkczyz: Oh, you can keep the custom error message. I'm just saying to replace `std::io::ErrorKind::Other` with `std::io::ErrorKind::InvalidData`.

TheBlueMatt (approved):

jkczyz (approved):

PR #876, opened 2021-04-13
Invoice features

valentinewallace:
Tested w/ sample

### DIFF
--- a/lightning-invoice/src/lib.rs
+++ b/lightning-invoice/src/lib.rs
@@ -992,10 +992,10 @@ impl Invoice {
 		self.signed_invoice.payee_pub_key().map(|x| &x.0)
 	}
 
-    /// Get the payment secret if one was included in the invoice
-    pub fn payment_secret(&self) -> Option<&PaymentSecret> {
-        self.signed_invoice.payment_secret()
-    }
+	/// Get the payment secret if one was included in the invoice
+	pub fn payment_secret(&self) -> Option<&PaymentSecret> {
+		self.signed_invoice.payment_secret()
+	}
 
 	/// Recover the payee's public key (only to be used if none was included in the invoice)
 	pub fn recover_payee_pub_key(&self) -> PublicKey {
--- a/lightning-invoice/src/lib.rs
+++ b/lightning-invoice/src/lib.rs
@@ -24,6 +24,7 @@ extern crate secp256k1;
 use bech32::u5;
 use bitcoin_hashes::Hash;
 use bitcoin_hashes::sha256;
+use lightning::ln::features::InvoiceFeatures;
 #[cfg(any(doc, test))]
 use lightning::routing::network_graph::RoutingFees;
 use lightning::routing::router::RouteHintHop;
@@ -329,6 +330,7 @@ pub enum TaggedField {
 	Fallback(Fallback),
 	Route(RouteHint),
 	PaymentSecret(PaymentSecret),
+	Features(InvoiceFeatures),
 }
 
 /// SHA-256 hash
@@ -401,6 +403,7 @@ pub mod constants {
 	pub const TAG_FALLBACK: u8 = 9;
 	pub const TAG_ROUTE: u8 = 3;
 	pub const TAG_PAYMENT_SECRET: u8 = 16;
+	pub const TAG_FEATURES: u8 = 5;
 }
 
 impl InvoiceBuilder<tb::False, tb::False, tb::False> {
@@ -489,6 +492,13 @@ impl<D: tb::Bool, H: tb::Bool, T: tb::Bool> InvoiceBuilder<D, H, T> {
 			Ok(r) => self.tagged_fields.push(TaggedField::Route(r)),
 			Err(e) => self.error = Some(e),
 		}
+		self
+	}
+
+	/// Adds a features field which indicates the set of supported protocol extensions which the
+	/// origin node supports.
+	pub fn features(mut self, features: InvoiceFeatures) -> Self {
+		self.tagged_fields.push(TaggedField::Features(features));
 		self
 	}
 }
@@ -810,6 +820,10 @@ impl RawInvoice {
 		find_extract!(self.known_tagged_fields(), TaggedField::PaymentSecret(ref x), x)
 	}
 
+	pub fn features(&self) -> Option<&InvoiceFeatures> {
+		find_extract!(self.known_tagged_fields(), TaggedField::Features(ref x), x)
+	}
+
 	pub fn fallbacks(&self) -> Vec<&Fallback> {
 		self.known_tagged_fields().filter_map(|tf| match tf {
 			&TaggedField::Fallback(ref f) => Some(f),
@@ -997,6 +1011,11 @@ impl Invoice {
 		self.signed_invoice.payment_secret()
 	}
 
+	/// Get the invoice features if they were included in the invoice
+	pub fn features(&self) -> Option<&InvoiceFeatures> {
+		self.signed_invoice.features()
+	}
+
 	/// Recover the payee's public key (only to be used if none was included in the invoice)
 	pub fn recover_payee_pub_key(&self) -> PublicKey {
 		self.signed_invoice.recover_payee_pub_key().expect("was checked by constructor").0
@@ -1054,6 +1073,7 @@ impl TaggedField {
 			TaggedField::Fallback(_) => constants::TAG_FALLBACK,
 			TaggedField::Route(_) => constants::TAG_ROUTE,
 			TaggedField::PaymentSecret(_) => constants::TAG_PAYMENT_SECRET,
+			TaggedField::Features(_) => constants::TAG_FEATURES,
 		};
 
 		u5::try_from_u8(tag).expect("all tags defined are <32")
--- a/lightning/src/ln/features.rs
+++ b/lightning/src/ln/features.rs
@@ -427,7 +427,6 @@ impl<T: sealed::Context> Features<T> {
 		Features::<C> { flags, mark: PhantomData, }
 	}
 
-	#[cfg(test)]
 	/// Create a Features given a set of flags, in LE.
 	pub fn from_le_bytes(flags: Vec<u8>) -> Features<T> {
 		Features {
--- a/lightning/src/ln/features.rs
+++ b/lightning/src/ln/features.rs
@@ -25,6 +25,8 @@
 use std::{cmp, fmt};
 use std::marker::PhantomData;
 
+use bitcoin::bech32;
+use bitcoin::bech32::{Base32Len, FromBase32, ToBase32, u5, WriteBase32};
 use ln::msgs::DecodeError;
 use util::ser::{Readable, Writeable, Writer};
 
@@ -51,6 +53,7 @@ mod sealed {
 			required_features: [$( $( $required_feature: ident )|*, )*],
 			optional_features: [$( $( $optional_feature: ident )|*, )*],
 		}) => {
+			#[derive(Eq, PartialEq)]
 			pub struct $context {}
 
 			impl Context for $context {
@@ -318,6 +321,7 @@ mod sealed {
 /// appears.
 ///
 /// (C-not exported) as we map the concrete feature types below directly instead
+#[derive(Eq)]
 pub struct Features<T: sealed::Context> {
 	/// Note that, for convenience, flags is LITTLE endian (despite being big-endian on the wire)
 	flags: Vec<u8>,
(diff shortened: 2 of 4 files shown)


lightning-invoice/src/lib.rs
```diff
@@ -506,6 +510,12 @@ impl<D: tb::Bool, H: tb::Bool, T: tb::Bool> InvoiceBuilder<D, H, T> {
 		}
 		self
 	}
+
+	/// Adds feature bits.
```
TheBlueMatt, line 514: This should have more details - "Adds a features field which indicates the set of supported protocol extensions which the invoice generator supports".

lightning/src/ln/features.rs
```diff
@@ -387,6 +391,17 @@ impl InitFeatures {
 	}
 }
 
+/// Calculates the base32 encoded size of a byte slice
+fn bytes_size_to_base32_size(byte_size: usize) -> usize {
```
TheBlueMatt, line 395: Can we use the provided implementation for `[u8]` instead of copying this function https://docs.rs/bech32/0.8.0/bech32/trait.Base32Len.html#impl-Base32Len

lightning/src/ln/features.rs
```diff
@@ -395,6 +410,77 @@ impl InvoiceFeatures {
 	}
 }
 
+impl ToBase32 for InvoiceFeatures {
+	fn write_base32<W: WriteBase32>(&self, writer: &mut W) -> Result<(), <W as WriteBase32>::Err> {
+		(&self.flags[..]).write_base32(writer)
```
TheBlueMatt, line 415: Don't we have to flip the endianness here, too?

lightning/src/ln/features.rs
```diff
@@ -395,6 +410,77 @@ impl InvoiceFeatures {
...
+	fn base32_len(&self) -> usize {
+		bytes_size_to_base32_size(self.flags.len())
+	}
+}
+
+impl FromBase32 for InvoiceFeatures {
+	type Err = bech32::Error;
+
+	fn from_base32(field_data: &[u5]) -> Result<InvoiceFeatures, bech32::Error> {
+		let mut features = InvoiceFeatures::empty();
```
TheBlueMatt, line 429: Features can be of arbitrary length, so we really need to just read the bytes, I think, not check individual bits and copy them.

valentinewallace: Hmm so sadly, `Vec::<u8>::from_base32(vec_of_u5s)` isn't working for me. E.g., if a bit is in the 9th to rightmost position in the vec(u5), it won't still be in the 9th-to-rightmost position after the conversion. 

It's making me think I need to write a custom converter like this Stackoverflow answer? https://stackoverflow.com/a/7135008 Does that sound right?

TheBlueMatt: Hmm, that's definitely what `Vec::<u8>::from_base32` should already be doing - https://docs.rs/bech32/0.8.0/src/bech32/lib.rs.html#238-240

valentinewallace: Yeah, related to offline discussion, it turns out I just need to bit shift the results :)

lightning/src/ln/features.rs
valentinewallace: That didn't work sadly... Neither did the original copied&pasted function. I did something super inefficient :grimacing: still need to think through a better way

lightning/src/ln/features.rs
valentinewallace: So, shifting the bits ended up being very edge casey and I spent too much time on it, and it ended up being much easier to just copy over the bits manually, and not use the [given Vec<u8> implementations](https://docs.rs/bech32/0.8.0/bech32/trait.ToBase32.html#impl-ToBase32)

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,64 @@ impl InvoiceFeatures {
 	}
 }
 
+impl ToBase32 for InvoiceFeatures {
+	fn write_base32<W: WriteBase32>(&self, writer: &mut W) -> Result<(), <W as WriteBase32>::Err> {
+		let total_u5s = ((self.flags.len() as f64 * 8 as f64) / 5 as f64).ceil() as usize;
+		let mut be_u5s: Vec<u5> = vec![u5::try_from_u8(0).unwrap(); total_u5s];
```
TheBlueMatt, line 405: Note I was able to get this to work with the bech32 native stuff, but its not as succinct as I'd like, take it or leave it:

```
let mut vec_read = bech32::convert_bits(field_data, 5, 8, true)?;
let mut good_bytes: Vec<u8>;
let bitshift = (field_data.len() * 5) % 8;

if bitshift == 0 {
	// If we don't need to shift, just byte-swap vec_read
	for i in 0..vec_read.len() / 2 {
		let tmp = vec_read[i];
		let swap_idx = vec_read.len() - i - 1;
		vec_read[i] = vec_read[swap_idx];
		vec_read[swap_idx] = tmp;
	}
	good_bytes = vec_read;
} else {
	// If we didn't read the right bit count, we need to drop bitshift bits at the end while byte swapping
	good_bytes = vec![0; total_bytes];
	for i in 0..total_bytes {
		let read_pos = vec_read.len() - i - 1;
		if i == total_bytes - 1 {
			good_bytes[i] = vec_read[read_pos] >> (8-bitshift);
		} else {
			good_bytes[i] = (vec_read[read_pos] >> (8-bitshift)) | (vec_read[read_pos - 1] << bitshift);
		}
	}
}
while !good_bytes.is_empty() && good_bytes[good_bytes.len() - 1] == 0 {
	good_bytes.pop();
}
Ok(InvoiceFeatures::from_le_bytes(good_bytes))
```

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,64 @@ impl InvoiceFeatures {
 	}
 }
 
+impl ToBase32 for InvoiceFeatures {
+	fn write_base32<W: WriteBase32>(&self, writer: &mut W) -> Result<(), <W as WriteBase32>::Err> {
+		let total_u5s = ((self.flags.len() as f64 * 8 as f64) / 5 as f64).ceil() as usize;
```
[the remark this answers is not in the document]
valentinewallace: Ok, let me know if there's a similar rule for the rounding down that I'm missing

lightning/src/ln/features.rs
valentinewallace: Ah, cool. Probably going to leave it but lmk if there's a strong reason to take it

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,68 @@ impl InvoiceFeatures {
...
+	fn write_base32<W: WriteBase32>(&self, writer: &mut W) -> Result<(), <W as WriteBase32>::Err> {
+		// Explanation for the "4": the normal way to round up when dividing is to add the divisor
+		// minus one before dividing
+		let total_u5s = (self.flags.len() * 8 + 4) / 5 as usize;
+		let mut be_u5s: Vec<u5> = vec![u5::try_from_u8(0).unwrap(); total_u5s];
+		for (u8_idx, byte) in self.flags.iter().enumerate() {
+			for bit_pos in 0..8 {
+				if 1 << bit_pos & byte != 0 {
+					let bit_pos_from_left_0_indexed = u8_idx * 8 + (bit_pos);
+					let new_u5_idx = total_u5s - (bit_pos_from_left_0_indexed as f64 / 5 as f64).floor() as usize - 1;
```
jkczyz, line 412: Is the `f64` cast and `floor` needed? Won't integer division suffice?

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,68 @@ impl InvoiceFeatures {
 	}
 }
 
+impl ToBase32 for InvoiceFeatures {
+	fn write_base32<W: WriteBase32>(&self, writer: &mut W) -> Result<(), <W as WriteBase32>::Err> {
```
jkczyz, line 403: Could you add some tests for these conversions?

lightning/src/ln/features.rs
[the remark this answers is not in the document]
TheBlueMatt: Just divide :) it rounds down by default.

lightning/src/ln/features.rs
TheBlueMatt: No, up to you, I was just annoyed that you said it wasn't working so had to fix it :).

lightning/src/ln/features.rs
valentinewallace: There is `test_payment_secret_and_features_de_and_ser` which uses those methods, and invoice fuzzing. Added more unit tests though.

lightning/src/ln/features.rs
```diff
@@ -427,7 +493,6 @@ impl<T: sealed::Context> Features<T> {
 		Features::<C> { flags, mark: PhantomData, }
 	}
 
-	#[cfg(test)]
```
TheBlueMatt, line 430: Can we expand the documentation to say "little endian" and note that "this is byte-swapped from most on-the-wire encodings"?

lightning/src/ln/features.rs
TheBlueMatt: Actually, I take that back, reading each bit one by one is definitely not the way to do this, especially if we branch on each bit. the above code works, but we can definitely write our own reader too, just as long as we aren't doing it bit-by-bit.

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,68 @@ impl InvoiceFeatures {
...
+	type Err = bech32::Error;
+
+	fn from_base32(field_data: &[u5]) -> Result<InvoiceFeatures, bech32::Error> {
+		// Explanation for the "7": the normal way to round up when dividing is to add the divisor
+		// minus one before dividing
+		let total_bytes = (field_data.len() * 5 + 7) / 8 as usize;
+		let mut le_bytes: Vec<u8> = vec![0; total_bytes];
+		for (u5_idx, chunk) in field_data.iter().enumerate() {
+			let chunk_as_u8 = chunk.to_u8();
+			for bit_pos in 0..5 {
```
TheBlueMatt, line 444: Same here - I don't think its acceptable to branch on each bit in the feature while reading.

lightning/src/ln/features.rs
valentinewallace: Ok, I guess in my mind there's not that many feature bits. Is this specifically for performance or also for a DoS vector reasons? Fine to switch to your impl^ above

TheBlueMatt: Right, my proposal above sucks, but there's not really a reason to do per-bit iteration, this works fine on the read end


```
		let mut res_bytes: Vec<u8> = vec![0; total_bytes];
		for (u5_idx, chunk) in field_data.iter().enumerate() {
			let bit_pos_from_right_0_indexed = (field_data.len() - u5_idx - 1) * 5;
			let new_byte_idx = (bit_pos_from_right_0_indexed / 8) as usize;
			let new_bit_pos = bit_pos_from_right_0_indexed % 8;
			let chunk_u16 = chunk.to_u8() as u16;
			res_bytes[new_byte_idx] |= ((chunk_u16 << new_bit_pos) & 0xff) as u8;
			if new_byte_idx != total_bytes - 1 {
				res_bytes[new_byte_idx + 1] |= ((chunk_u16 >> (8-new_bit_pos)) & 0xff) as u8;
			}
		}
```

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,68 @@ impl InvoiceFeatures {
...
+			let chunk_u16 = chunk.to_u8() as u16;
+			res_bytes[new_byte_idx] |= ((chunk_u16 << new_bit_pos) & 0xff) as u8;
+			if new_byte_idx != total_bytes - 1 {
+				res_bytes[new_byte_idx + 1] |= ((chunk_u16 >> (8-new_bit_pos)) & 0xff) as u8;
+			}
+		}
+		// Trim the highest feature bits.
+		while !res_bytes.is_empty() && res_bytes[res_bytes.len() - 1] == 0 {
+			res_bytes.pop();
+}
```
TheBlueMatt, line 459: nit: indent.

TheBlueMatt (approved):
Looks good, trivial comments and one doc comment.

lightning/src/ln/features.rs
```diff
@@ -427,8 +493,8 @@ impl<T: sealed::Context> Features<T> {
 		Features::<C> { flags, mark: PhantomData, }
 	}
 
-	#[cfg(test)]
-	/// Create a Features given a set of flags, in LE.
+	/// Create a Features given a set of flags, in little-endian. This is in reverse bit order from
```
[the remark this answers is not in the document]
valentinewallace: Maybe dumb q, but the bits are literally in reverse order from what they were, right?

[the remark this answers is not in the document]
TheBlueMatt: I don't believe so? The highest bit in the last position is ultimately the highest bit - not the lowest bit in the last position (which would be true if we swapped the bits as well).

lightning-invoice/src/lib.rs
```diff
@@ -491,6 +494,13 @@ impl<D: tb::Bool, H: tb::Bool, T: tb::Bool> InvoiceBuilder<D, H, T> {
 		}
 		self
 	}
+
+	/// Adds a features field which indicates the set of supported protocol extensions which the
+	/// invoice generator supports.
```
ariard, line 499: nit: spec says "origin node", i like it too :)

lightning/src/ln/features.rs
```diff
@@ -395,6 +399,68 @@ impl InvoiceFeatures {
 	}
 }
 
+impl ToBase32 for InvoiceFeatures {
+	fn write_base32<W: WriteBase32>(&self, writer: &mut W) -> Result<(), <W as WriteBase32>::Err> {
+		// Explanation for the "4": the normal way to round up when dividing is to add the divisor
+		// minus one before dividing
+		let total_u5s = (self.flags.len() * 8 + 4) / 5 as usize;
```
ariard, line 406: nit: `size_u5s`, `length_u5s` ?

ariard (commented):
utACK 9e82f39

TheBlueMatt:
Test locally, CI is just hella broken right now. See #895.

PR #885, opened 2021-04-14
Fix size check in Vec<Signature> serialization + use consts

Tibo-lg:
I noticed that the constant `33` was used in the deserialization of `Vec<Signature>` while I think it should be `64` since signatures are serialized in compact format. This PR fixes this and replaced the numerical `33` and `64` constants with the ones defined in rust-secp256k1 as I believe this could help prevent future mistakes.

(I don't think this issue is a big deal but thought it was worth fixing)


TheBlueMatt (approved):
Good catch!

valentinewallace (approved):

PR #895, opened 2021-04-23
Invoice chanman utility

valentinewallace:
~TODO: tests~

Based on #893


lightning/src/ln/channelmanager.rs
```diff
@@ -2397,6 +2460,11 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		self.our_network_pubkey.clone()
 	}
 
+	/// Gets the node secret held by this ChannelManager
+	pub fn get_our_node_secret(&self) -> SecretKey {
```
TheBlueMatt, line 2464: Because its a secret key, we should probably return a reference to it instead of copying the actual secret bytes in memory. Likely to be optimized into the same thing, but its still "best practice".

lightning-invoice/src/lib.rs
```diff
@@ -890,6 +892,76 @@ impl Deref for PositiveTimestamp {
...
+					proportional_millionths: forwarding_info.fee_proportional_millionths,
+				},
+				cltv_expiry_delta: forwarding_info.cltv_expiry_delta,
+				htlc_minimum_msat: None,
+				htlc_maximum_msat: None,
+			}]);
+		}
+
+		let (payment_hash, payment_secret) = channelmanager.get_payment_secret_preimage(
+			if amt_msat == 0 { None } else { Some(amt_msat) },
```
TheBlueMatt, line 937: nit: because its a minimum, providing `Some(0)` is the same as `None`. Should we just make it non-`Option`al?

lightning-invoice/src/lib.rs
```diff
@@ -890,6 +892,76 @@ impl Deref for PositiveTimestamp {
...
+			min_final_cltv_expiry.into(),
+			0,
+		);
+		let secp_ctx = Secp256k1::new();
+		let our_node_pubkey = channelmanager.get_our_node_id();
+		let mut invoice = InvoiceBuilder::new(match network {
```
TheBlueMatt, line 943: Needs features, right, or is just waiting on #876 to land?

TheBlueMatt (commented):
Looks good.

lightning-invoice/src/lib.rs
valentinewallace: Yeah, locally it's based on features. Now getting the sample based on this PR so I can add a test with an invoice I actually know is good...

lightning-invoice/src/lib.rs
valentinewallace: I'm fine with `Option`al, switched the param

lightning/src/ln/channelmanager.rs
```diff
@@ -210,6 +212,33 @@ pub struct PaymentPreimage(pub [u8;32]);
 #[derive(Hash, Copy, Clone, PartialEq, Eq, Debug)]
 pub struct PaymentSecret(pub [u8;32]);
```
jkczyz, line 213: Perhaps we can move `PaymentSecret` into a more low-level module like `ln`. We can reexport it in `channelmanager` to avoid making any cascading changes, if needed.

lightning-invoice/src/lib.rs
```diff
@@ -901,6 +906,79 @@ impl Deref for PositiveTimestamp {
...
+		.min_final_cltv_expiry(min_final_cltv_expiry.into());
+		if let Some(amt) = amt_msat {
+			invoice = invoice.amount_pico_btc(amt * 10);
+		}
+		for hint in route_hints.drain(..) {
+			invoice = invoice.route(hint);
+		}
+
+		invoice.build_signed(|msg_hash| {
+			secp_ctx.sign_recoverable(msg_hash, &channelmanager.get_our_node_secret())
```
jkczyz, line 978: Rather than exposing the secret, maybe `ChannelManager` could provide an interface for signing?

lightning-invoice/src/lib.rs
```diff
@@ -901,6 +906,79 @@ impl Deref for PositiveTimestamp {
 }
 
 impl Invoice {
+	/// Utility to construct an invoice.
+	pub fn from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
+		channelmanager: &ChannelManager<Signer, M, T, K, F, L>, amt_msat: Option<u64>, description: String,
+		network: Network,
+	) -> Result<Invoice, CreationError>
```
jkczyz, line 913: Could you move this into a separate utility module akin to `lightning_block_sync::init`? I'd prefer to separate higher-level lightning data structures from lower-level invoice code when possible.

lightning-invoice/Cargo.toml
```diff
@@ -10,11 +10,19 @@ readme = "README.md"
...
 bech32 = "0.7"
+bitcoin = "0.26"
 lightning = { version = "0.0.13", path = "../lightning" }
 secp256k1 = { version = "0.20", features = ["recovery"] }
 num-traits = "0.2.8"
 bitcoin_hashes = "0.9.4"
 
+[dev-dependencies.bitcoin]
+version = "0.26"
+features = ["bitcoinconsensus"]
```
TheBlueMatt, line 21: Not that it matters, but I think the solution here is to make rust-lightning's _test_utils feature depend on "bitcoin/bitcoinconsensus" instead.

lightning-invoice/src/lib.rs
```diff
@@ -904,6 +906,79 @@ impl Deref for PositiveTimestamp {
 }
 
 impl Invoice {
+	/// Utility to construct an invoice.
+	pub fn from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
+		channelmanager: &ChannelManager<Signer, M, T, K, F, L>, amt_msat: Option<u64>, description: String,
+		network: Network,
```
TheBlueMatt, line 912: Looks like we're taking the dependency on rust-bitcoin specifically to take a network here and then convert it to a Currency, can we just take a Currency?

lightning-invoice/src/lib.rs
```diff
@@ -904,6 +906,79 @@ impl Deref for PositiveTimestamp {
...
+		for channel in our_channels {
+			let short_channel_id = match channel.short_channel_id {
+				Some(id) => id,
+				None => continue,
+			};
+			let forwarding_info = match channel.counterparty_forwarding_info {
+				Some(info) => info,
+				None => continue,
+			};
+			if forwarding_info.cltv_expiry_delta > min_final_cltv_expiry {
```
TheBlueMatt, line 934: I dont think we need to check this? They're additive, doesn't matter if one hop is lower than our final CLTV.

lightning-invoice/src/lib.rs
```diff
@@ -904,6 +906,79 @@ impl Deref for PositiveTimestamp {
...
+		M::Target: chain::Watch<Signer>,
+		T::Target: BroadcasterInterface,
+		K::Target: KeysInterface<Signer = Signer>,
+		F::Target: FeeEstimator,
+		L::Target: Logger,
+	{
+		// Marshall route hints.
+		let our_channels = channelmanager.list_usable_channels();
+		let mut route_hints = vec![];
+		let mut min_final_cltv_expiry = 18;
```
TheBlueMatt, line 924: Maybe use the new constant in one of the later updates of the dependent PRs https://github.com/rust-bitcoin/rust-lightning/pull/893/commits/615ef7d6f88464ed8478a8c8cc0e7582ba66d773

lightning-invoice/src/lib.rs
valentinewallace: Hm, what do you think the module should be called? `utils` perhaps?

TheBlueMatt:
Can now be rebased after the merge of #893.

lightning/src/ln/channelmanager.rs
valentinewallace: Feels a bit odd to separate out the PaymentSecret from the PaymentPreimage/PaymentHash. Maybe they should all be moved in follow-up?

TheBlueMatt: I mean its a few extra lines of diff, might as well move them all in one go.

lightning-invoice/src/lib.rs
TheBlueMatt: If we sign anything blindly, it feels like basically the same thing - we'd have to actually have logic knowing what we're signing to avoid it, which would be great, except then we get into lightning depending on lightning-invoice and circular dependencies :/.

lightning-invoice/src/lib.rs
jkczyz: Yeah, `utils` would be fine.

lightning/Cargo.toml
```diff
@@ -32,10 +32,6 @@ bitcoin = "0.26"
 hex = { version = "0.3", optional = true }
 regex = { version = "0.1.80", optional = true }
 
-[dev-dependencies.bitcoin]
-version = "0.26"
```
TheBlueMatt, line 36: Oops, we still need this block for `cargo test -p lightning` because that isn't `_test_util` just `test`

valentinewallace: Ah, thanks that's what I'm looking for

lightning-invoice/src/lib.rs
```diff
@@ -1295,10 +1290,100 @@ impl<S> Display for SignOrCreationError<S> {
...
+	) -> Result<Invoice, CreationError>
+	where
+		M::Target: chain::Watch<Signer>,
+		T::Target: BroadcasterInterface,
+		K::Target: KeysInterface<Signer = Signer>,
+		F::Target: FeeEstimator,
+		L::Target: Logger,
+	{
+		// Marshall route hints.
+		let our_channels = channelmanager.list_usable_channels();
```
ariard, line 1321: IIUC this utility returns as routing hints the complete set of our channels, including the non-announced subset one ? 

Sounds pretty harmful for privacy, maybe the utility should have a boolean `all` or otherwise a Vec<short_channel_id> to precise which ones to disclose ?

lightning-invoice/src/lib.rs
ariard: Pretty agree with Jeff here, does isolating the utility in its own module resolve the circular dependency ?

Extracting the node secret key from the node will let any malicious entity signs a bunch of random invoices with known hash/preimages. If the victim node is a merchant it opens the way to further social attacks, like contacting the support to claim that a service/good is owed and has not been delivered ?

ariard (commented):
Otherwise sounds good to me :)

lightning-invoice/src/lib.rs
```diff
@@ -1292,10 +1290,100 @@ impl<S> Display for SignOrCreationError<S> {
 	}
 }
 
+mod utils {
```
jkczyz, line 1293: I was thinking this could be in its own file as well for code cleanliness.

fuzz/src/full_stack.rs
```diff
@@ -14,918 +14,1220 @@
...
-	v[0] = ((u >> 8*7) & 0xff) as u8;
-	v[1] = ((u >> 8*6) & 0xff) as u8;
-	v[2] = ((u >> 8*5) & 0xff) as u8;
-	v[3] = ((u >> 8*4) & 0xff) as u8;
-	v[4] = ((u >> 8*3) & 0xff) as u8;
-	v[5] = ((u >> 8*2) & 0xff) as u8;
-	v[6] = ((u >> 8*1) & 0xff) as u8;
-	v[7] = ((u >> 8*0) & 0xff) as u8;
-	v
+    let mut v = [0; 8];
```
jkczyz, line 79: Formatting changes in `full_stack.rs` look undesirable. See commit 5a478c769dc106ea0a0a0747973ef2d0f67288fe.

lightning-invoice/src/lib.rs
TheBlueMatt: > does isolating the utility in its own module resolve the circular dependency

No, the only solution we could do would be to pull at least `Features` (and probably a few other tiny things) into a new crate which both rust-lightning and lightning-invoice depend on. I'm not super excited about that, but its possible.

> Extracting the node secret key from the node will let any malicious entity signs a bunch of random invoices with known hash/preimages

Yea, we don't really have a concrete story for the node secret's security model right now. We do require it in ChannelManager, so we have to assume anyone with access to the `ChannelManager` can, in principle, do anything they want with the node secret (at least its on the same host/in the same memory space). Its unclear if there's a lot we can do about that, either, we need it to read onions. We could make reading onions a part of the keys interface, but that does seem like an API that is going to change a *ton*.

In general, I think we really need a spec update to let us sign invoices with a different key than onion decryption (etc). cc @devrandom.

devrandom: It would be good not to go further in the direction of passing around the private key.

I think we can have a function `sign_invoice` on `KeysInterface` with the following arguments:

- hrp
- data

and they could be either `Vec<u5>` or `Vec<u8>`.

If the developer actually wants to validate inside `KeysInterface`, they will have to re-parse the invoice.  I think that's fine, since you likely go across process boundaries when signing in the general case (hardware wallet / HSM / hardened signing server).

As to the node secret in other contexts - I think it boils down to signing things and computing shared secrets.  I think both of these can be behind `KeysInterface` (or a smaller trait that can be gotten from it).  I'm happy to try implementing there's doubt.

lightning-invoice/src/lib.rs
TheBlueMatt: I think this is the right default to this - in most cases if we're a "client" node we only have one or two channels anyway (and in non-"client" cases we have several channels that are public), and the "real" fix is to generate random short channel ids.

valentinewallace:
Oops, need to switch `sign_invoice` to return a `Result`. Otherwise I think this is roughly what it's gonna look like.

lightning-invoice/src/lib.rs
TheBlueMatt: Probably also should be pub :).

TheBlueMatt: Probably also should be pub in some way :).

lightning-invoice/src/lib.rs
```diff
@@ -1365,9 +1364,32 @@ pub mod utils {
...
+		}
+
+		invoice_preimage.extend_from_slice(&Vec::<u8>::from_base32(&data_part)
+		  .expect("No padding error may occur due to appended zero above."));
+		let signature = match keys_manager.sign_invoice(invoice_preimage) {
+			Ok(sig) => sig,
+			Err(_) => return Err(SemanticError::InvalidSignature)
+		};
+		let signed_raw_invoice: Result<SignedRawInvoice, SemanticError> = raw_invoice.sign(|_| Ok(signature));
+		Invoice::from_signed(signed_raw_invoice.unwrap())
```
jkczyz, line 1392: The signing API is designed to allow custom errors. So the error type here should be what is returned by `sign_invoice`. Something like:

```rust
let signed_raw_invoice = raw_invoice.sign(|_| keys_manager.sign_invoice(invoice_preimage))?;
```

 But may have to convert to a `SignOrCreationError`.

fuzz/src/full_stack.rs
TheBlueMatt: Looks like rustfmt accidentally got run across this file?

lightning-invoice/src/lib.rs
```diff
@@ -1295,10 +1290,123 @@ impl<S> Display for SignOrCreationError<S> {
 	}
 }
 
+	/// Convenient utilities to create an invoice.
```
TheBlueMatt, line 1293: nit indentation here.

lightning-invoice/src/lib.rs
```diff
@@ -1295,10 +1290,123 @@ impl<S> Display for SignOrCreationError<S> {
...
+	use lightning::chain::keysinterface::{Sign, KeysInterface};
+	use lightning::ln::channelmanager::{ChannelManager, MIN_FINAL_CLTV_EXPIRY};
+	use lightning::ln::features::InvoiceFeatures;
+	use lightning::routing::network_graph::RoutingFees;
+	use lightning::routing::router::RouteHintHop;
+	use lightning::util::logger::Logger;
+	use std::ops::Deref;
+
+	/// Utility to construct an invoice.
+	#[allow(dead_code)]
```
TheBlueMatt, line 1309: This shouldnt be dead code if its publicly-exposed.

lightning-invoice/src/lib.rs
```diff
@@ -1295,10 +1290,123 @@ impl<S> Display for SignOrCreationError<S> {
...
+	use lightning::chain::chaininterface::{BroadcasterInterface, FeeEstimator};
+	use lightning::chain::keysinterface::{Sign, KeysInterface};
+	use lightning::ln::channelmanager::{ChannelManager, MIN_FINAL_CLTV_EXPIRY};
+	use lightning::ln::features::InvoiceFeatures;
+	use lightning::routing::network_graph::RoutingFees;
+	use lightning::routing::router::RouteHintHop;
+	use lightning::util::logger::Logger;
+	use std::ops::Deref;
+
+	/// Utility to construct an invoice.
```
TheBlueMatt, line 1308: Probably useful to describe when you'd want to use this and why.

lightning-invoice/src/lib.rs
```diff
@@ -1295,10 +1290,123 @@ impl<S> Display for SignOrCreationError<S> {
...
+	#[allow(dead_code)]
+	pub fn from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
+		channelmanager: &ChannelManager<Signer, M, T, K, F, L>, amt_msat: Option<u64>, description: String,
+		network: Currency, keys_manager: &dyn KeysInterface<Signer = Signer>
```
TheBlueMatt, line 1312: Should be no reason to need a dyn here - can we just add a template parameter?

lightning-invoice/src/lib.rs
```diff
@@ -1295,10 +1290,123 @@ impl<S> Display for SignOrCreationError<S> {
...
+	use lightning::ln::channelmanager::{ChannelManager, MIN_FINAL_CLTV_EXPIRY};
+	use lightning::ln::features::InvoiceFeatures;
+	use lightning::routing::network_graph::RoutingFees;
+	use lightning::routing::router::RouteHintHop;
+	use lightning::util::logger::Logger;
+	use std::ops::Deref;
+
+	/// Utility to construct an invoice.
+	#[allow(dead_code)]
+	pub fn from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
```
TheBlueMatt, line 1310: May be useful to call this `Invoice_from_...` or so. In context I guess it makes sense but if its just in a module on its own it would be nice to reference the thing being created. Do you agree @jkczyz?

jkczyz: Maybe `create_invoice_from_channelmanager` since using a `from_` prefix tends to imply a conversion.

lightning-invoice/src/lib.rs
```diff
@@ -1301,15 +1302,14 @@ mod utils {
...
 	use std::ops::Deref;
 
 	/// Utility to construct an invoice.
 	#[allow(dead_code)]
 	pub fn from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
 		channelmanager: &ChannelManager<Signer, M, T, K, F, L>, amt_msat: Option<u64>, description: String,
-		network: Currency,
-	) -> Result<Invoice, CreationError>
+		network: Currency, keys_manager: &dyn KeysInterface<Signer = Signer>
+	) -> Result<Invoice, SemanticError>
```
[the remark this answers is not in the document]
valentinewallace: Hm so `Invoice::from_signed` returns ` SemanticError`, which is a bit tricky to translate to a `CreationError`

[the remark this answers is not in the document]
jkczyz: Exactly. This would indicate a bug on our end so we should `unwrap` or `expect` there.

valentinewallace:
Gonna have lunch then come back to this

lightning/src/chain/keysinterface.rs
```diff
@@ -391,6 +392,12 @@ pub trait KeysInterface: Send + Sync {
 	/// contain no versioning scheme. You may wish to include your own version prefix and ensure
 	/// you've read all of the provided bytes to ensure no corruption occurred.
 	fn read_chan_signer(&self, reader: &[u8]) -> Result<Self::Signer, DecodeError>;
+
+	/// Sign an invoice's preimage (note that this is the preimage of the invoice, not the HTLC's
+	/// preimage). By parameterizing by the preimage instead of the hash, we allow implementors of
+	/// this trait to parse the invoice and make sure they're signing what they expect, rather than
+	/// blindly signing the hash.
+	fn sign_invoice(&self, invoice_preimage: Vec<u8>) -> Result<RecoverableSignature, DecodeError>;
```
TheBlueMatt, line 400: `DecodeError` definitely feels pretty strange here - I suppose in principle users may be decoding so that's a part of the possible errors, but users are also signing, and you could imagine something like "hard ware wallet unplugged" as an error. I guess we could do `()` for the error type  here like we do a few others - ultimately we're just gonna pass the user their own error back again.

lightning/src/ln/channelmanager.rs
```diff
@@ -2475,6 +2463,11 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		self.our_network_pubkey.clone()
 	}
 
+	/// Gets the node secret held by this ChannelManager
```
TheBlueMatt, line 2466: This should be dropped now.

lightning-invoice/src/utils.rs
```diff
@@ -0,0 +1,173 @@
...
+		data_part.push(u5::try_from_u8(0).unwrap());
+
+		// if overhang is in (1..3) we need to add u5(0) padding two times
+		if overhang < 3 {
+			data_part.push(u5::try_from_u8(0).unwrap());
+		}
+	}
+
+	invoice_preimage.extend_from_slice(&Vec::<u8>::from_base32(&data_part)
+																		 .expect("No padding error may occur due to appended zero above."));
```
jkczyz, line 100: bad indentation

lightning/src/chain/keysinterface.rs
```diff
@@ -1047,6 +1054,15 @@ impl KeysInterface for KeysManager {
...
 		InMemorySigner::read(&mut std::io::Cursor::new(reader))
 	}
+
+	fn sign_invoice(&self, invoice_preimage: Vec<u8>) -> Result<RecoverableSignature, ()> {
+		let secp_ctx = Secp256k1::new();
+		let mut raw_hash: [u8; 32] = Default::default();
+		raw_hash.copy_from_slice(&Sha256::hash(&invoice_preimage)[..]);
+		let msg_hash = Message::from_slice(&raw_hash[..])
+			.expect("Hash is 32 bytes long, same as MESSAGE_SIZE");
+		Ok(secp_ctx.sign_recoverable(&msg_hash, &self.get_node_secret()))
```
TheBlueMatt, line 1064: You can simplify this a good bit (plus then drop the `Message` import above.
```suggestion
		Ok(self.secp_ctx.sign_recoverable(&hash_to_message!(&Sha256::hash(&invoice_preimage)), &self.get_node_secret()))
```

lightning-invoice/src/utils.rs
```diff
@@ -0,0 +1,173 @@
...
+	let signed_raw_invoice: Result<SignedRawInvoice, DecodeError> = raw_invoice.sign(|_| keys_manager.sign_invoice(invoice_preimage));
```
jkczyz, line 101: Type declaration is probably unnecessary now.

lightning-invoice/src/utils.rs
```diff
@@ -0,0 +1,173 @@
...
+	match signed_raw_invoice {
+		Ok(inv) => return Ok(Invoice::from_signed(inv).unwrap()),
+		Err(e) => return Err(SignOrCreationError::SignError(e))
+	};
```
jkczyz, line 105: nit:

```rust
match signed_raw_invoice {
	Ok(inv) => Ok(Invoice::from_signed(inv).unwrap()),
	Err(e) => Err(SignOrCreationError::SignError(e)),
}
```

lightning-invoice/src/utils.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::ops::Deref;
+
+/// Utility to construct an invoice. Generally, unless you want to do something like a custom
+/// cltv_expiry, this is what you should be using to create an invoice. The reason being, this
+/// method stores the invoice's payment secret and preimage in `ChannelManager`, so (a) the user
+/// doesn't have to store preimage/payment secret information and (b) `ChannelManager` can verify
+/// that the payment secret is valid when the invoice is paid.
+pub fn create_invoice_from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
+	channelmanager: &ChannelManager<Signer, M, T, K, F, L>, keys_manager: K, amt_msat: Option<u64>,
+	description: String, network: Currency
```
jkczyz, line 23: Rename `currency` to `network` and consider moving it before `amt_msat`.

TheBlueMatt (commented):
Aside from Jeff's comments and https://github.com/rust-bitcoin/rust-lightning/pull/895#discussion_r623219246 this LGTM.

lightning-invoice/src/lib.rs
ariard: Agree for the "client" node but for a big hub you will disclose all your private channels with spoke *if* you're also a merchant which is quite the case of a bunch of such nodes. Though opening an issue for now, we'll adjust in function of users feedback.

TheBlueMatt: Right, we ultimately need some heuristic of "we have > 5 public channels, we shouldn't include any route hints at all", probably.

lightning-invoice/src/lib.rs
```diff
@@ -747,8 +747,13 @@ impl RawInvoice {
 		}
 
 		preimage.extend_from_slice(&Vec::<u8>::from_base32(&data_part)
-			.expect("No padding error may occur due to appended zero above."));
+		  .expect("No padding error may occur due to appended zero above."));
```
jkczyz, line 750: nit: leave as tab

TheBlueMatt (approved):

jkczyz (approved):

lightning-invoice/src/utils.rs
```diff
@@ -0,0 +1,158 @@
+/// Convenient utilities to create an invoice.
```
TheBlueMatt, line 1: Lol, oops, needs a `pub mod utils` in lib.rs or this doesn't even compile.

TheBlueMatt (changes requested):

valentinewallace:
Gah my bad, fixed

lightning-invoice/src/utils.rs
TheBlueMatt, line 1: In order for this to be a doc comment it needs to be "//!" not "///"

lightning-invoice/src/utils.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// Utility to construct an invoice. Generally, unless you want to do something like a custom
+/// cltv_expiry, this is what you should be using to create an invoice. The reason being, this
+/// method stores the invoice's payment secret and preimage in `ChannelManager`, so (a) the user
+/// doesn't have to store preimage/payment secret information and (b) `ChannelManager` can verify
+/// that the payment secret is valid when the invoice is paid.
+pub fn create_invoice_from_channelmanager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
+	channelmanager: &ChannelManager<Signer, M, T, K, F, L>, keys_manager: K, network: Currency,
+	amt_msat: Option<u64>, description: String
+) -> Result<Invoice, SignOrCreationError<DecodeError>>
```
TheBlueMatt, line 24: This doesn't compile now.

lightning-invoice/src/utils.rs
valentinewallace: TIL

PR #900, opened 2021-04-28
Send Init message immediately after handshake even if not initiator

devrandom:
We were waiting for the initiator, but the spec doesn't guarantee that they will send Init first, so we might theoretically wait forever.

Also, lnprototest expects this behavior.


TheBlueMatt (approved):

ariard (commented):
Code Review ACK 7113dbd

I read again the BOLT init section and effectively it says :

"
The sending node:
  - MUST send `init` as the first Lightning message for any connection.
 "
 Without precising if the sending maps to inbound or outbound or indifferent. Even if minor, I would invite to clarify the spec on this point.

PR #920, opened 2021-05-12
Background processing of ChannelManager and ChannelMonitor events

jkczyz:
Support processing `Event`s from `ChannelManager` and `ChainMonitor` in the background. Adds an `EventHandler` trait for users to implement and pass to `BackgroundProcessor`. Refactors the `EventsProvider` trait to take such an event handler.

This also addresses a race condition identified in #899 by making use of `PersistenceNotifierGuard` in `ChannelManager` when processing events.


lightning/src/ln/channelmanager.rs
```diff
@@ -3522,22 +3535,27 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
+	M::Target: chain::Watch<Signer>,
+	T::Target: BroadcasterInterface,
+	K::Target: KeysInterface<Signer = Signer>,
+	F::Target: FeeEstimator,
+	L::Target: Logger,
 {
-	fn get_and_clear_pending_events(&self) -> Vec<Event> {
+	fn process_pending_events<H: Deref>(&self, handler: H) -> Result<(), E>
+	where H::Target: EventHandler<E> {
+		let _persistence_guard = PersistenceNotifierGuard::new(&self.total_consistency_lock, &self.persistence_notifier);
```
TheBlueMatt, line 3548: After #918 this probably should figure out if any events were processed and not always persist.

lightning/src/util/events.rs
```diff
@@ -381,8 +382,20 @@ pub trait MessageSendEventsProvider {
...
-	/// Gets the list of pending events which were generated by previous actions, clearing the list
-	/// in the process.
-	fn get_and_clear_pending_events(&self) -> Vec<Event>;
+pub trait EventsProvider<E> {
+	/// Processes any generated events with the given event handler.
+	fn process_pending_events<H: Deref>(&self, handler: H) -> Result<(), E>
+		where H::Target: EventHandler<E>;
+}
+
+/// A trait implemented for objects handling events from [`EventsProvider`].
```
TheBlueMatt, line 391: This or the above needs a discussion of the locking semantics and reentrancy.

lightning/src/ln/channelmanager.rs
```diff
@@ -3522,22 +3535,27 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
 		//TODO: This behavior should be documented. It's non-intuitive that we query
 		// ChannelMonitors when clearing other events.
 		self.process_pending_monitor_events();
 
-		let mut ret = Vec::new();
 		let mut pending_events = self.pending_events.lock().unwrap();
-		mem::swap(&mut ret, &mut *pending_events);
-		ret
+		for event in pending_events.drain(..) {
+			handler.handle_event(event)?;
```
jkczyz, line 3555: The code as written here will pass back the error of the first encountered error and stop processing events. However, any remaining events are lost because of the `drain`. So this leads to some other questions:
- Should event processing continue if handling one event fails?
- Can a user reasonably recover from an error while handling an event?
- Should unprocessed events remain pending for the next call?

The answer to these will determine what the interface will ultimately look like.

lightning-background-processor/src/lib.rs
```diff
@@ -129,10 +138,12 @@ impl BackgroundProcessor {
 			let mut current_time = Instant::now();
 			loop {
 				peer_manager.process_events();
+				channel_manager.process_pending_events(&event_handler).unwrap(); // TODO: Return the error
+				chain_monitor.process_pending_events(&event_handler).unwrap(); // TODO: Return the error
```
jkczyz, line 142: There's an open question around how to handle errors in `BackgroundProcessor`. Currently, it will simply `unwrap` and fail if there is an error. Alternatively, it could return the error in the same manner as it does for the persister, which would necessitate defining an error type covering both.

jkczyz (commented):
Added some comments that I could use some feedback on.

lightning/src/ln/channelmanager.rs
TheBlueMatt: Right, if we take an error from users we definitely shouldn't just drop events, we should keep feeding them to users. That said, I'm a little dubious that we should let users provide Errors either. We shouldn't take an error unless we can do something reasonable with it - maybe try the event again later or something, but short of that I'm not sure what an error even *means*.

jkczyz: Yeah, I'm starting to think not allowing errors is the best approach. The user can always write an `EventHandler` to do whatever they want (including collecting errors or remaining events) and act upon it after `process_pending_events` returns. But, depending on how implemented, it could leave them in a state where `ChannelManager` is persisted without those events but the user hasn't "fully" handled them either. May just need to spell it out in the documentation.

That said, if they are using `BackgroundProcessor` there would be no way for them to examine the handler after processing.

TheBlueMatt: Right, I guess the only case where errors make sense is some kind of transitory issue that we know will resolve very soon, allowing us to handle the error momentarily while holding off persisting the ChannelManager. That said, I think you could just block in the handler in that case, as long as we only hold an event-specific lock?

Maybe we can have a separate lock allowing us to block any other event handlers from entering while still allowing adding new events from other things going on?

jkczyz: > Right, I guess the only case where errors make sense is some kind of transitory issue that we know will resolve very soon, allowing us to handle the error momentarily while holding off persisting the ChannelManager. That said, I think you could just block in the handler in that case, as long as we only hold an event-specific lock?
> 

Will update this to remove the return type from `handle_event`.

> Maybe we can have a separate lock allowing us to block any other event handlers from entering while still allowing adding new events from other things going on?

`pending_events` is already behind a `Mutex`. But I guess if we want to allow other event producers to add more events, we could immediately drop the lock after either draining fully into a `Vec` or swapping as before.

TheBlueMatt: > pending_events is already behind a Mutex. But I guess if we want to allow other event producers to add more events, we could immediately drop the lock after either draining fully into a Vec or swapping as before.

Right, I guess I was thinking that if a user were handling events in parallel it would be a bug. Now that I think about it more maybe its fine, as long as that's what the user wants.

lightning/src/ln/channelmanager.rs
jkczyz: I think this meant to say #916.

lightning/src/ln/channelmanager.rs
jkczyz: FWIW, swapping retains the same behavior as `get_and_clear_pending_events` since there the lock was held for the duration of the method and then the events were handled by the user. I updated the code accordingly.

lightning-background-processor/src/lib.rs
```diff
@@ -263,30 +291,55 @@ mod tests {
...
+			node.best_block = BestBlock::new(header.block_hash(), height);
+			match i {
+				1 => {
+					node.node.transactions_confirmed(&header, &txdata, height);
+					node.chain_monitor.transactions_confirmed(&header, &txdata, height);
+				},
+				ANTI_REORG_DELAY => {
+					node.node.best_block_updated(&header, height);
+					node.chain_monitor.best_block_updated(&header, height);
+				},
```
valentinewallace, line 321: Wondering why we only call this on the last block, rather than for every block in `i..=ANTI_REORG_DELAY`?

valentinewallace:
Could also be useful to document somewhere what we discussed offline re: expected behavior if we crash in the middle of handling a series of events.

jkczyz (commented):
> Could also be useful to document somewhere what we discussed offline re: expected behavior if we crash in the middle of handling a series of events.

Done.

lightning/src/util/events.rs
```diff
@@ -380,9 +381,32 @@ pub trait MessageSendEventsProvider {
 	fn get_and_clear_pending_msg_events(&self) -> Vec<MessageSendEvent>;
 }
 
-/// A trait indicating an object may generate events
+/// A trait indicating an object may generate events.
+///
+/// Events are processed by a handler given to [`process_pending_events`]. Therefore, implementors
+/// should be mindful to avoid lock reentrancy if a handler may call back into the provider.
```
TheBlueMatt, line 387: Should we make it a part of the contract that callbacks into the provider are allowed/don't run locked, except for (possibly) serialization (in the case of a ChannelManager)? Its kinda hard to work with this API without that guarantee, but its kinda annoying to specify it.

lightning/src/chain/channelmonitor.rs
```diff
@@ -226,7 +226,7 @@ pub(crate) const LATENCY_GRACE_PERIOD_BLOCKS: u32 = 3;
 /// It may cause spurrious generation of bumped claim txn but that's allright given the outpoint is already
 /// solved by a previous claim tx. What we want to avoid is reorg evicting our claim tx and us not
 /// keeping bumping another claim tx to solve the outpoint.
-pub(crate) const ANTI_REORG_DELAY: u32 = 6;
+pub const ANTI_REORG_DELAY: u32 = 6;
```
TheBlueMatt, line 229: Maybe make the last four lines a comment instead of documentation? Feels weird to expose them.

lightning/src/util/events.rs
jkczyz: For `ChannelManager`, both `process_pending_events` and `handle_event` take a read lock. The write lock is only taken after notification (i.e., after `process_pending_events` has returned) -- the handler itself doesn't serialize `ChannelManager`.

So I think the general advice is "don't introduce a deadlock". But now that I think of it, does `RwLock` allow reentrant reads? It appears so though the docs say it _might_ panic.

https://doc.rust-lang.org/std/sync/struct.RwLock.html#panics

lightning-net-tokio/src/lib.rs
```diff
@@ -47,12 +47,12 @@
 //!     lightning_net_tokio::connect_outbound(peer_manager, sender, their_node_id, addr).await;
 //!     loop {
 //!         receiver.recv().await;
-//!         for _event in channel_manager.get_and_clear_pending_events().drain(..) {
-//!             // Handle the event!
-//!         }
-//!         for _event in chain_monitor.get_and_clear_pending_events().drain(..) {
-//!             // Handle the event!
-//!         }
+//!         channel_manager.process_pending_events(&|event| {
```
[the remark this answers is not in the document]
jkczyz: Sure, I can remove it. But I don't quite understand why (if?) my change caused this to longer be necessary.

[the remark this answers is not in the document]
TheBlueMatt: The point of the channel here is to have something that you can use to wake up every time an Event may be generated (by hooking network reads). It's a stupid hack, but it works. Since we now do the handling based on the ChannelManager's own notification interface, I don't think there's any need to do it by hooking network reads anymore.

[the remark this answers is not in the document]
jkczyz: Ah, yeah, that's the piece I was missing. Sill need to call that interface here to avoid busy waiting.

lightning/src/util/events.rs
TheBlueMatt: > So I think the general advice is "don't introduce a deadlock".

Right, but the docs don't *really* give you enough information to know what would and what would not cause a deadlock. Reading it naively, I'd think I need to queue all the events and then call back to ChannelManager separately as calling back to it may cause a deadlock. Providing more information would let users do more without fear, though we have to be precise and decide what we're willing to commit to in the API and what not.

> does RwLock allow reentrant reads?

I assumed it would...I mean you're allowed to take as many read locks as you want....

jkczyz: > Right, but the docs don't _really_ give you enough information to know what would and what would not cause a deadlock. Reading it naively, I'd think I need to queue all the events and then call back to ChannelManager separately as calling back to it may cause a deadlock. Providing more information would let users do more without fear, though we have to be precise and decide what we're willing to commit to in the API and what not.

Understood, but these aren't `ChannelManager`'s docs. And note that this line is talking about someone implementing `EventsProvider` not `EventHandler`.  It's really up to an implementor of `EventsProvider` to determine what is safe for a handler to call. We don't need to restrict it at the trait level.

TheBlueMatt: Right, I guess then the docs are somewhat confusing to me - I read them as "user, you should be careful of reentrancy when calling methods here", not "note to RL implementors, be careful of reentrancy". We should have some docs per-implementor that talks about reentrancy, I guess, though sadly Rust really hides the impl-level docs for a trait implementation, making them hard to find :(.

lightning-background-processor/src/lib.rs
```diff
@@ -263,30 +291,55 @@ mod tests {
...
-		let bg_processor = BackgroundProcessor::start(callback, nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
-
 		// Go through the channel creation process until each node should have something persisted.
 		let tx = open_channel!(nodes[0], nodes[1], 100000);
 
+		// Initiate the background processors to watch each node.
+		let data_dir = nodes[0].persister.get_data_dir();
+		let persister = move |node: &ChannelManager<InMemorySigner, Arc<ChainMonitor>, Arc<test_utils::TestBroadcaster>, Arc<KeysManager>, Arc<test_utils::TestFeeEstimator>, Arc<test_utils::TestLogger>>| FilesystemPersister::persist_manager(data_dir.clone(), node);
+		let event_handler = |_| {};
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
```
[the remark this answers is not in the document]
valentinewallace: Ah, I see. The comment above `open_channel!` may need updating since it implies that a ChannelManager will be persisted by the time it finishes. 

> Initially, I didn't move it, but the test failed in CI because open_channel! calls get_and_clear_pending_events, which races with BackgroundProcessor. That is it, the event handler may process the FundingGenerationReady event before open_channel! can do so, causing an assertion to fail in open_channel!.
 
Could we add a comment about this? :) Otherwise the test structure is a bit unintuitive to me

lightning/src/util/events.rs
```diff
@@ -380,9 +381,32 @@ pub trait MessageSendEventsProvider {
...
+	/// duplicate events across process restarts. This may occur if the provider was recovered from
+	/// an old state (i.e., it hadn't been successfully persisted after processing pending events).
+	fn process_pending_events<H: Deref>(&self, handler: H) where H::Target: EventHandler;
+}
+
+/// A trait implemented for objects handling events from [`EventsProvider`].
+pub trait EventHandler {
+	/// Handles the given [`Event`].
+	///
+	/// See [`EventsProvider::process_pending_events`] for details that must be considered when
```
valentinewallace, line 403: Could point to the overall trait docs since those seem to have considerations (like the reentrancy) that could be pertinent as well

lightning-background-processor/src/lib.rs
```diff
@@ -101,25 +104,31 @@ impl BackgroundProcessor {
...
-		M::Target: 'static + chain::Watch<Signer>,
+		CF::Target: 'static + chain::Filter,
+		CW::Target: 'static + chain::Watch<Signer>,
 		T::Target: 'static + BroadcasterInterface,
 		K::Target: 'static + KeysInterface<Signer = Signer>,
 		F::Target: 'static + FeeEstimator,
 		L::Target: 'static + Logger,
+		P::Target: 'static + channelmonitor::Persist<Signer>,
 		CMH::Target: 'static + ChannelMessageHandler,
 		RMH::Target: 'static + RoutingMessageHandler,
```
valentinewallace, line 133: Lol a casual 30-line method signature... Curious what the "best practice" for this would be. Definitely has been an issue since before this PR though.

lightning/src/chain/channelmonitor.rs
```diff
@@ -222,11 +222,11 @@ pub(crate) const CLTV_CLAIM_BUFFER: u32 = 18;
...
-/// We use also this delay to be sure we can remove our in-flight claim txn from bump candidates buffer.
-/// It may cause spurrious generation of bumped claim txn but that's allright given the outpoint is already
-/// solved by a previous claim tx. What we want to avoid is reorg evicting our claim tx and us not
-/// keeping bumping another claim tx to solve the outpoint.
-pub(crate) const ANTI_REORG_DELAY: u32 = 6;
+// We also use this delay to be sure we can remove our in-flight claim txn from bump candidates buffer.
+// It may cause spurious generation of bumped claim txn but that's alright given the outpoint is already
+// solved by a previous claim tx. What we want to avoid is reorg evicting our claim tx and us not
+// keep bumping another claim tx to solve the outpoint.
+pub const ANTI_REORG_DELAY: u32 = 6;
```
valentinewallace, line 229: Could add or expose some public docs for this.

lightning/src/ln/channelmanager.rs
```diff
@@ -3563,21 +3573,29 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
+	F::Target: FeeEstimator,
+	L::Target: Logger,
 {
-	fn get_and_clear_pending_events(&self) -> Vec<Event> {
-		//TODO: This behavior should be documented. It's non-intuitive that we query
-		// ChannelMonitors when clearing other events.
-		self.process_pending_monitor_events();
+	fn process_pending_events<H: Deref>(&self, handler: H) where H::Target: EventHandler {
+		PersistenceNotifierGuard::optionally_notify(&self.total_consistency_lock, &self.persistence_notifier, || {
+			//TODO: This behavior should be documented. It's non-intuitive that we query
```
valentinewallace, line 3585: nit: space before todo + s/non-intuitive/unintuitive

valentinewallace (commented):
Submitted some comments too early. Just some docs nits + the reentrancy clarification, otherwise LGTM

lightning/src/ln/channelmanager.rs
```diff
@@ -3563,21 +3573,29 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
+	K::Target: KeysInterface<Signer = Signer>,
+	F::Target: FeeEstimator,
+	L::Target: Logger,
 {
-	fn get_and_clear_pending_events(&self) -> Vec<Event> {
-		//TODO: This behavior should be documented. It's non-intuitive that we query
-		// ChannelMonitors when clearing other events.
-		self.process_pending_monitor_events();
+	fn process_pending_events<H: Deref>(&self, handler: H) where H::Target: EventHandler {
+		PersistenceNotifierGuard::optionally_notify(&self.total_consistency_lock, &self.persistence_notifier, || {
```
TheBlueMatt, line 3584: While you're at it, can you add a persistence notifier in `get_and_clear_pending_msg_events` too, if it does actual work.

lightning/src/util/events.rs
jkczyz: Re-wrote the docs to separate implementation and uses. Also added docs to `ChannelManager` and `ChainMonitor` implementations. Let me know if anything isn't clear or could be expanded upon.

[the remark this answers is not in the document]
jkczyz: Rewrote as per earlier discussion. Let me know if anything could be cleared up further.

lightning-background-processor/src/lib.rs
jkczyz: Not sure honestly unless we want to use trait objects. :(

lightning-net-tokio/src/lib.rs
```diff
@@ -43,32 +41,30 @@
 //!
 //! // Connect to node with pubkey their_node_id at addr:
 //! async fn connect_to_node(peer_manager: PeerManager, chain_monitor: Arc<ChainMonitor>, channel_manager: ChannelManager, their_node_id: PublicKey, addr: SocketAddr) {
-//!     let (sender, mut receiver) = mpsc::channel(2);
```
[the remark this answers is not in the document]
jkczyz: Good question! I think `Cargo.lock` handles this for us since it will add the revision to the package source.

lightning/src/chain/channelmonitor.rs
valentinewallace: Ah, I missed that, thanks

lightning/src/util/events.rs
TheBlueMatt: Having a whole section dedicated to `Implementations` that contains instructions for how to implement just feels really wrong in public docs here? In practice, I can't imagine we ever expect users to implement this, so it becomes just a list of instructions to ourselves in public docs?

lightning/src/ln/channelmanager.rs
```diff
@@ -3551,33 +3562,59 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
+			} else {
+				result.replace(pending_events);
+				NotifyOption::DoPersist
+			}
+		});
+		result.into_inner()
 	}
 }
 
+/// Provides events that must be periodically handled.
```
TheBlueMatt, line 3586: These docs are collapsed by default on the ChannelManager page, beside four other trait implementations, making them basically invisible IMO. I don't know if there's a way to change rustdoc's behavior here, but I wonder if it makes sense to explicitly link to the known concrete implementations in the `EventsProvider` docs and say, eg, "eg see [\`ChannelManager::process_pending_events\`] and ..." which seems to handle opening the panel.

lightning/src/ln/channelmanager.rs
```diff
@@ -3551,33 +3562,59 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
-		ret
+			let mut pending_events = Vec::new();
+			let mut channel_state = self.channel_state.lock().unwrap();
+			mem::swap(&mut pending_events, &mut channel_state.pending_msg_events);
+
+			if pending_events.is_empty() {
+				NotifyOption::SkipPersist
+			} else {
+				result.replace(pending_events);
+				NotifyOption::DoPersist
```
TheBlueMatt, line 3579: Because message events aren't persisted as a part of ChannelManager (and we never assume that messages will be delivered between when we create them and give them to peers), I dont think this needs a persistence. However, we should be checking if a persist is required as a result of `process_pending_monitor_events` (both here and in `process_pending_events` below).

lightning-net-tokio/src/lib.rs
```diff
@@ -218,11 +202,10 @@ impl Connection {
 		}
 		if let Disconnect::PeerDisconnected = disconnect_type {
 			peer_manager_ref.socket_disconnected(&our_descriptor);
-			Self::event_trigger(&mut us.lock().unwrap());
```
TheBlueMatt, line 221: Oops, I forgot that we're also relying on the event trigger thing here to call `peer_manager.process_events()` (which is also dumb...). Can you add a call to it after the select in the loop and maybe here after the disconnect as well? (while you're at it, we don't need the `peer_manager_ref` clone anymore, it seems?).

TheBlueMatt (commented):
Aside from two functional comments here and the two doc comments, this all looks good.

lightning/src/util/events.rs
jkczyz: Combined and rephrased in terms of requirements.

TheBlueMatt (approved):

lightning/src/chain/chainmonitor.rs
```diff
@@ -306,12 +315,17 @@ impl<ChannelSigner: Sign, C: Deref, T: Deref, F: Deref, L: Deref, P: Deref> even
 	      L::Target: Logger,
 	      P::Target: channelmonitor::Persist<ChannelSigner>,
 {
-	fn get_and_clear_pending_events(&self) -> Vec<Event> {
+	/// Processes [`SpendableOutputs`] events produced from each [`ChannelMonitor`] upon maturity.
+	///
+	/// [`SpendableOutputs`]: events::Event::SpendableOutputs
```
valentinewallace, line 320: Should we note explicitly that calling back into the provider is OK?

valentinewallace (approved):

lightning/src/chain/chainmonitor.rs
jkczyz: Hmm, good catch. Though we don't expect any need to call back, It's actually not safe to call `watch_channel` since it would attempt to take a `write` lock while already holding a `read` lock.

jkczyz: Actually, never mind. We drop the read lock before calling the handler.

TheBlueMatt:
D'oh. Sorry, I missed this conflicted with 851 before I hit merge. Could you do a quick rebase.

jkczyz:
> D'oh. Sorry, I missed this conflicted with 851 before I hit merge. Could you do a quick rebase.

Sure thing. One sec.

valentinewallace (approved):

lightning/src/ln/channelmanager.rs
```diff
@@ -3680,35 +3693,69 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> MessageSend
...
+			let mut result = NotifyOption::SkipPersist;
+
+			// TODO: This behavior should be documented. It's unintuitive that we query
+			// ChannelMonitors when clearing other events.
+			if self.process_pending_monitor_events() {
+				result = NotifyOption::DoPersist;
+			}
 
-		self.check_free_holding_cells();
+			self.check_free_holding_cells();
```
TheBlueMatt, line 3706: I think we need to check `check_free_holding_cells` for updates to persist as well.

jkczyz: Ah, sorry. Didn't look at that closely enough. Figured it was a simple merge conflict. Let me fix it.

jkczyz: To confirm my understand, should we persist whenever `check_free_holding_cells` gets any failed HTLCs back from `maybe_free_holding_cell_htlcs`? Or is the condition more complicated?

https://github.com/rust-bitcoin/rust-lightning/blob/4e8bed20760ee1b53cf2fd15d4a9897780c4fab5/lightning/src/ln/channelmanager.rs#L3524

TheBlueMatt: We should also persist if we get a `ChannelMonitorUpdate` back from `maybe_free_holding_cell_htlcs` (`ChannelMonitorUpdate`s are always a surefire reason to persist - if we're out of sync when we come back up we have to force-close). Reading the code in channel that should suffice, but maybe add a comment in `maybe_free_holding_cell_htlcs` and/or `free_holding_cell_htlcs` that notes that we won't persist unless we get some values back in the return.

jkczyz: Left the comment out since it's really a concern of `ChannelManager`. Made the logic in `check_free_holding_cells` explicit, though, so hopefully this is clear.

TheBlueMatt (approved):

PR #922, opened 2021-05-14
Correct default invoice expiry

valentinewallace:
We previously stated in the codebase that the default invoice expiry
stated in the spec is 2 hours. It's actually 1 hour.

Relevant spec: https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#tagged-fields


lightning-invoice/src/utils.rs
```diff
@@ -54,7 +54,7 @@ where
 
 	let (payment_hash, payment_secret) = channelmanager.create_inbound_payment(
 		amt_msat,
-		7200, // default invoice expiry is 2 hours
+		3600, // default invoice expiry is 1 hours
```
jkczyz, line 57: Could you reuse `DEFAULT_EXPIRY_TIME` here?

TheBlueMatt (approved):

jkczyz (approved):

PR #929, opened 2021-05-26
Parse RPC errors as JSON content

jkczyz:
Bitcoin Core's JSON RPC server returns errors as HTTP error responses with JSON content in the body. Parse this content as JSON to give a more meaningful error. Otherwise, "binary" is given because the content contains ASCII control characters.

Most of the parsing was already in place, but it was only triggered for OK responses.


ariard (commented):
Good to me b2f16ad

TheBlueMatt (approved):

PR #931, opened 2021-05-27
Cache socket address in HttpClient for reconnect

### LINKED ISSUE #930: lightning-block-sync::HttpClient::send_request_with_retry panics if bitcoind crashes
If a query to `bitcoind`'s RPC interface is performed once `bitcoind` is off / has crashed (but it was previously reachable), `send_request_with_retry` does eventually panic with:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }'
```

The issue comes from unwrapping https://github.com/rust-bitcoin/rust-lightning/blob/main/lightning-block-sync/src/http.rs#L165.

Here's a minimal example to reproduce (using BitcoindClient from https://github.com/lightningdevkit/ldk-sample/blob/main/src/bitcoind_client.rs):

```
use crate::bitcoind_client::BitcoindClient;
use std::sync::Arc;
use std::{thread, time};

#[tokio::main]
pub async fn main() {
    let host = String::from("localhost");
    let port = 18443;
    let user = String::from("user");
    let password = String::from("passwd");

    // Initialize our bitcoind client.
    let bitcoin_cli = match BitcoindClient::new(host, port, user, password).await {
        Ok(client) => Arc::new(client),
        Err(e) => {
            println!("Failed to connect to bitcoind client: {}", e);
            return;
        }
    };

    let polling_delta = time::Duration::from_secs(1);

    loop {
        match bitcoin_cli.get_best_block_hash().await {
            Ok(_) => (),
            Err(_) => println!("Connection lost with bitcoind"),
        };

        thread::sleep(polling_delta);
    }
}
```

Steps to reproduce:

- `cargo run` with bitcoind running
- stop bitcoind

Result:

```
Connection lost with bitcoind
Connection lost with bitcoind
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', /Users/sergi/.cargo/git/checkouts/rust-lightning-6850882c83c110d8/499d84c/lightning-block-sync/src/http.rs:164:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jkczyz:
If the `HttpClient` attempts to reconnect to bitcoind that is no longer running, the client fails to get the address from the stream. Cache the address when connecting to prevent this.

Fixes #930

### DIFF
--- a/lightning-block-sync/src/http.rs
+++ b/lightning-block-sync/src/http.rs
@@ -8,7 +8,7 @@ use std::convert::TryFrom;
 use std::fmt;
 #[cfg(not(feature = "tokio"))]
 use std::io::Write;
-use std::net::ToSocketAddrs;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::time::Duration;
 
 #[cfg(feature = "tokio")]
@@ -97,6 +97,7 @@ impl<'a> std::net::ToSocketAddrs for &'a HttpEndpoint {
 
 /// Client for making HTTP requests.
 pub(crate) struct HttpClient {
+	address: SocketAddr,
 	stream: TcpStream,
 }
 
@@ -119,7 +120,7 @@ impl HttpClient {
 			TcpStream::from_std(stream)?
 		};
 
-		Ok(Self { stream })
+		Ok(Self { address, stream })
 	}
 
 	/// Sends a `GET` request for a resource identified by `uri` at the `host`.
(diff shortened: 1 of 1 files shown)


TheBlueMatt (approved):

valentinewallace (approved):

devrandom:
İ'm not sure this is a good idea, since it effectively caches DNS lookups indefinitely.

jkczyz:
> İ'm not sure this is a good idea, since it effectively caches DNS lookups indefinitely.

Technically, this isn't any different behavior from before when using `TcpStream::peer_addr`. We'd have to parameterize `HttpClient` (rather than `connect`) by `ToSocketAddrs` to avoid any caching.

PR #946, opened 2021-06-09
Align get_route's interface with ChannelManager and Invoice

jkczyz:
Update `get_route`'s interface to work better with `ChannelManager` and `Invoice`.
- ~~pass result of `ChannelManager::list_usable_channels` as `first_hops` without converting~~
- accept multi-hop hints for `last_hops` from `Invoice::routes` without copying

This is only the interface change for #945. Only the last hop from each `RouteHint` is considered by `get_route` for now.

### DIFF
--- a/lightning-invoice/src/lib.rs
+++ b/lightning-invoice/src/lib.rs
@@ -30,7 +30,7 @@ use lightning::ln::PaymentSecret;
 use lightning::ln::features::InvoiceFeatures;
 #[cfg(any(doc, test))]
 use lightning::routing::network_graph::RoutingFees;
-use lightning::routing::router::RouteHintHop;
+use lightning::routing::router::RouteHint;
 
 use secp256k1::key::PublicKey;
 use secp256k1::{Message, Secp256k1};
@@ -362,7 +362,7 @@ pub enum TaggedField {
 	ExpiryTime(ExpiryTime),
 	MinFinalCltvExpiry(MinFinalCltvExpiry),
 	Fallback(Fallback),
-	Route(RouteHint),
+	PrivateRoute(PrivateRoute),
 	PaymentSecret(PaymentSecret),
 	Features(InvoiceFeatures),
 }
@@ -419,7 +419,7 @@ pub struct InvoiceSignature(pub RecoverableSignature);
 /// The encoded route has to be <1024 5bit characters long (<=639 bytes or <=12 hops)
 ///
 #[derive(Eq, PartialEq, Debug, Clone)]
-pub struct RouteHint(Vec<RouteHintHop>);
+pub struct PrivateRoute(RouteHint);
 
 /// Tag constants as specified in BOLT11
 #[allow(missing_docs)]
@@ -431,7 +431,7 @@ pub mod constants {
 	pub const TAG_EXPIRY_TIME: u8 = 6;
 	pub const TAG_MIN_FINAL_CLTV_EXPIRY: u8 = 24;
 	pub const TAG_FALLBACK: u8 = 9;
-	pub const TAG_ROUTE: u8 = 3;
+	pub const TAG_PRIVATE_ROUTE: u8 = 3;
 	pub const TAG_PAYMENT_SECRET: u8 = 16;
 	pub const TAG_FEATURES: u8 = 5;
 }
@@ -509,9 +509,9 @@ impl<D: tb::Bool, H: tb::Bool, T: tb::Bool, C: tb::Bool, S: tb::Bool> InvoiceBui
 	}
 
 	/// Adds a private route.
-	pub fn route(mut self, route: Vec<RouteHintHop>) -> Self {
-		match RouteHint::new(route) {
-			Ok(r) => self.tagged_fields.push(TaggedField::Route(r)),
+	pub fn private_route(mut self, hint: RouteHint) -> Self {
+		match PrivateRoute::new(hint) {
+			Ok(r) => self.tagged_fields.push(TaggedField::PrivateRoute(r)),
 			Err(e) => self.error = Some(e),
 		}
 		self
@@ -913,8 +913,8 @@ impl RawInvoice {
 		find_all_extract!(self.known_tagged_fields(), TaggedField::Fallback(ref x), x).collect()
 	}
 
-	pub fn routes(&self) -> Vec<&RouteHint> {
-		find_all_extract!(self.known_tagged_fields(), TaggedField::Route(ref x), x).collect()
+	pub fn private_routes(&self) -> Vec<&PrivateRoute> {
+		find_all_extract!(self.known_tagged_fields(), TaggedField::PrivateRoute(ref x), x).collect()
 	}
 
 	pub fn amount_pico_btc(&self) -> Option<u64> {
@@ -1163,8 +1163,15 @@ impl Invoice {
 	}
 
 	/// Returns a list of all routes included in the invoice
-	pub fn routes(&self) -> Vec<&RouteHint> {
-		self.signed_invoice.routes()
+	pub fn private_routes(&self) -> Vec<&PrivateRoute> {
+		self.signed_invoice.private_routes()
+	}
+
+	/// Returns a list of all routes included in the invoice as the underlying hints
+	pub fn route_hints(&self) -> Vec<&RouteHint> {
+		find_all_extract!(
+			self.signed_invoice.known_tagged_fields(), TaggedField::PrivateRoute(ref x), x
+		).map(|route| &**route).collect()
 	}
 
 	/// Returns the currency for which the invoice was issued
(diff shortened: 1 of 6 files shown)


lightning/src/routing/router.rs
```diff
@@ -343,8 +343,8 @@ fn compute_fees(amount_msat: u64, channel_fees: RoutingFees) -> Option<u64> {
 /// The fees on channels from us to next-hops are ignored (as they are assumed to all be
 /// equal), however the enabled/disabled bit on such channels as well as the
 /// htlc_minimum_msat/htlc_maximum_msat *are* checked as they may change based on the receiving node.
-pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, payee: &PublicKey, payee_features: Option<InvoiceFeatures>, first_hops: Option<&[&ChannelDetails]>,
-	last_hops: &[&RouteHintHop], final_value_msat: u64, final_cltv: u32, logger: L) -> Result<Route, LightningError> where L::Target: Logger {
+pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, payee: &PublicKey, payee_features: Option<InvoiceFeatures>, first_hops: Option<&[ChannelDetails]>,
+	last_hops: Vec<&[RouteHintHop]>, final_value_msat: u64, final_cltv: u32, logger: L) -> Result<Route, LightningError> where L::Target: Logger {
```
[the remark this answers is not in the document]
jkczyz: Primarily because `RouteHint`'s interface uses an error type from `lightning-invoice`, and I didn't want to go down the path of changing that. Also note that `Invoice`'s interface gives `Vec<&RouteHint>` where `RouteHint` wraps `Vec< RouteHintHop>`.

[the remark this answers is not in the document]
TheBlueMatt: Hmm, can we just swap CreationError for ()? We only fail in one case and it's easy enough to map it on the invoice side.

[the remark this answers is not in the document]
jkczyz: Considered doing this, but the check really belongs in `lightning-invoice` as it has to do with keeping the route within a serialization limitation.

https://github.com/rust-bitcoin/rust-lightning/blob/a8038a8234b70bc261f633a9e2b40adebebcd8d4/lightning-invoice/src/lib.rs#L419

Instead, I made a `RouteHint` struct in `lightning` and wrapped it in `lightning-invoice` by `Route` (renaming `RouteHint` there), which is a more consistent name with the other fields. Ended up adding a `route_hints` method to `Invoice` to do the conversion. Was hoping there was a simple way to do this with `Into` but couldn't quite get something working.

lightning-invoice/src/lib.rs
```diff
@@ -419,7 +419,7 @@ pub struct InvoiceSignature(pub RecoverableSignature);
 /// The encoded route has to be <1024 5bit characters long (<=639 bytes or <=12 hops)
 ///
 #[derive(Eq, PartialEq, Debug, Clone)]
-pub struct RouteHint(Vec<RouteHintHop>);
+pub struct Route(RouteHint);
```
TheBlueMatt, line 422: Hmm, I don't think `Route` is right, either, its not a Route in the sense that you can't pay over it, its still a hint. More generally, the bindings will be sad that we have two things called `Route`, even in different scopes. Maybe `InvoiceRouteHint`, if that's not too verbose? I'd maybe vote for changing most of the `route` things in the invoice API to `route_hint` anyway.

TheBlueMatt:
Lets land this for 0.0.98 to clean up the API which will clean up our docs and sample a good chunk.

lightning/src/routing/router.rs
```diff
@@ -343,8 +347,8 @@ fn compute_fees(amount_msat: u64, channel_fees: RoutingFees) -> Option<u64> {
 /// The fees on channels from us to next-hops are ignored (as they are assumed to all be
 /// equal), however the enabled/disabled bit on such channels as well as the
 /// htlc_minimum_msat/htlc_maximum_msat *are* checked as they may change based on the receiving node.
-pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, payee: &PublicKey, payee_features: Option<InvoiceFeatures>, first_hops: Option<&[&ChannelDetails]>,
-	last_hops: &[&RouteHintHop], final_value_msat: u64, final_cltv: u32, logger: L) -> Result<Route, LightningError> where L::Target: Logger {
+pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, payee: &PublicKey, payee_features: Option<InvoiceFeatures>, first_hops: Option<&[ChannelDetails]>,
+	last_hops: Vec<&RouteHint>, final_value_msat: u64, final_cltv: u32, logger: L) -> Result<Route, LightningError> where L::Target: Logger {
```
TheBlueMatt, line 351: Can we take a `&[&RouteHint]`?

TheBlueMatt (commented):
This basically looks good.

lightning/src/routing/router.rs
jkczyz: I was hoping to make it easy for callers by allowing them to pass the result of `Invoice::route_hints` directly. I guess they could just take a slice now that `route_hints` does the conversion, but I'll have to either touch or revert a bunch of code. Any specific reason a slice would be preferred?

jkczyz: Updated to take a slice.

TheBlueMatt (approved):
LGTM, can you squash fixups?

lightning/src/routing/router.rs
```diff
@@ -343,8 +347,8 @@ fn compute_fees(amount_msat: u64, channel_fees: RoutingFees) -> Option<u64> {
 /// The fees on channels from us to next-hops are ignored (as they are assumed to all be
 /// equal), however the enabled/disabled bit on such channels as well as the
 /// htlc_minimum_msat/htlc_maximum_msat *are* checked as they may change based on the receiving node.
-pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, payee: &PublicKey, payee_features: Option<InvoiceFeatures>, first_hops: Option<&[&ChannelDetails]>,
-	last_hops: &[&RouteHintHop], final_value_msat: u64, final_cltv: u32, logger: L) -> Result<Route, LightningError> where L::Target: Logger {
+pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, payee: &PublicKey, payee_features: Option<InvoiceFeatures>, first_hops: Option<&[ChannelDetails]>,
```
TheBlueMatt, line 350: Oops, ugh, the `ChannelDetails` reference is here mostly for the bindings - the bindings take an array of pointers with some additional metadata, so they can't trivially convert them to `[ChannelDetails]` without a bunch of copies.

lightning/src/routing/router.rs
```diff
@@ -359,7 +363,8 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		return Err(LightningError{err: "Cannot send a payment of 0 msat".to_owned(), action: ErrorAction::IgnoreError});
 	}
 
-	for last_hop in last_hops {
+	let last_hops = last_hops.iter().filter_map(|hops| hops.0.last()).collect::<Vec<_>>();
```
TheBlueMatt, line 366: Should we mention in the docs that multi-hop hints are currently ignored and only the last hop is examined?

lightning/src/routing/router.rs
jkczyz: Dropped the commit but it does make the user do more work at the call site. Here is what the guide is looking like:

```rust
let amt_pico_btc = invoice.amount_pico_btc()
    .expect("ERROR: invalid invoice: must contain amount to pay");
let amt_msat = amt_pico_btc / 10;
let payer_pubkey = channel_manager.get_our_node_id();
let network_graph = router.network_graph.read().unwrap();
let payee_pubkey = invoice.recover_payee_pub_key();
let payee_features = invoice.features().cloned();
let first_hops = channel_manager.list_usable_channels();
let last_hops = invoice.route_hints();
let final_cltv = invoice.min_final_cltv_expiry() as u32;

let route = router::get_route(
    &payer_pubkey, &network_graph, &payee_pubkey, payee_features,
    Some(&first_hops.iter().collect::<Vec<_>>()), &last_hops,
    amt_msat, final_cltv, logger.clone(),
).expect("ERROR: failed to find route");

let payment_hash = PaymentHash(invoice.payment_hash().clone().into_inner());
let payment_secret = invoice.payment_secret().cloned();

channel_manager.send_payment(&route, payment_hash, &payment_secret)
    .expect("ERROR: failed to send payment");
```

Later, we may want take a pass through how our interfaces to see how well they mesh together.

TheBlueMatt (approved):
Needs squash.

lightning/src/routing/router.rs
TheBlueMatt: Yea, its very not-ideal, I agree. I'm not really sure how to fix it in the bindings context, though, sadly.

lightning-invoice/src/utils.rs
```diff
@@ -128,7 +123,7 @@ mod test {
 			&invoice.recover_payee_pub_key(),
 			Some(invoice.features().unwrap().clone()),
 			Some(&first_hops.iter().collect::<Vec<_>>()),
-			&last_hops.iter().collect::<Vec<_>>(),
+			&last_hops,
```
valentinewallace, line 126: Much nicer

valentinewallace (approved):

PR #1040, opened 2021-08-09
Multi-Hop Route Hint now considered. Added in unit tests for same.

abhik-99:
This PR solves Issue #945 which deals with using multi-hop route hints from the invoice from payee. Previous behaviour was to only consider the last route hop hint per route. This PR solves the problems with PR #1030 and also includes the unit tests for testing out the Route Hint in a multi-hop setup.

Edit: Commit [914c159](https://github.com/rust-bitcoin/rust-lightning/pull/1040/commits/914c159b3f4e69259157f3e1334c7823bb551712) solves a previously failing test where the channel capacity was calculated to be 0 which skipped certain hints.

Edit: The above has been squashed into fd4d3f

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ lightning-c-bindings/a.out
 **/*.rs.bk
 Cargo.lock
 .idea
-
+lightning/target
--- a/lightning/src/routing/router.rs
+++ b/lightning/src/routing/router.rs
@@ -375,10 +375,11 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		return Err(LightningError{err: "Cannot send a payment of 0 msat".to_owned(), action: ErrorAction::IgnoreError});
 	}
 
-	let last_hops = last_hops.iter().filter_map(|hops| hops.0.last()).collect::<Vec<_>>();
-	for last_hop in last_hops.iter() {
-		if last_hop.src_node_id == *payee {
-			return Err(LightningError{err: "Last hop cannot have a payee as a source.".to_owned(), action: ErrorAction::IgnoreError});
+	for route in last_hops.iter() {
+		for hop in &route.0 {
+			if hop.src_node_id == *payee {
+				return Err(LightningError{err: "Last hop cannot have a payee as a source.".to_owned(), action: ErrorAction::IgnoreError});
+			}
 		}
 	}
 
@@ -876,39 +877,89 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		// If a caller provided us with last hops, add them to routing targets. Since this happens
 		// earlier than general path finding, they will be somewhat prioritized, although currently
 		// it matters only if the fees are exactly the same.
-		for hop in last_hops.iter() {
+		for route in last_hops.iter().filter(|route| !route.0.is_empty()) {
+			let first_hop_in_route = &(route.0)[0];
 			let have_hop_src_in_graph =
-				// Only add the last hop to our candidate set if either we have a direct channel or
-				// they are in the regular network graph.
-				first_hop_targets.get(&hop.src_node_id).is_some() ||
-				network.get_nodes().get(&hop.src_node_id).is_some();
+				// Only add the hops in this route to our candidate set if either
+				// we have a direct channel to the first hop or the first hop is
+				// in the regular network graph.
+				first_hop_targets.get(&first_hop_in_route.src_node_id).is_some() ||
+				network.get_nodes().get(&first_hop_in_route.src_node_id).is_some();
 			if have_hop_src_in_graph {
-				// BOLT 11 doesn't allow inclusion of features for the last hop hints, which
-				// really sucks, cause we're gonna need that eventually.
-				let last_hop_htlc_minimum_msat: u64 = match hop.htlc_minimum_msat {
-					Some(htlc_minimum_msat) => htlc_minimum_msat,
-					None => 0
-				};
-				let directional_info = DummyDirectionalChannelInfo {
-					cltv_expiry_delta: hop.cltv_expiry_delta as u32,
-					htlc_minimum_msat: last_hop_htlc_minimum_msat,
-					htlc_maximum_msat: hop.htlc_maximum_msat,
-					fees: hop.fees,
-				};
-				// We assume that the recipient only included route hints for routes which had
-				// sufficient value to route `final_value_msat`. Note that in the case of "0-value"
-				// invoices where the invoice does not specify value this may not be the case, but
-				// better to include the hints than not.
-				if add_entry!(hop.short_channel_id, hop.src_node_id, payee, directional_info, Some((final_value_msat + 999) / 1000), &empty_channel_features, 0, path_value_msat, 0) {
-					// If this hop connects to a node with which we have a direct channel,
-					// ignore the network graph and, if the last hop was added, add our
-					// direct channel to the candidate set.
-					//
-					// Note that we *must* check if the last hop was added as `add_entry`
-					// always assumes that the third argument is a node to which we have a
-					// path.
-					if let Some(&(ref first_hop, ref features, ref outbound_capacity_msat, _)) = first_hop_targets.get(&hop.src_node_id) {
-						add_entry!(first_hop, *our_node_id , hop.src_node_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, 0, path_value_msat, 0);
+				// We start building the path from reverse, i.e., from payee
+				// to the first RouteHintHop in the path.
+				let hop_iter = route.0.iter().rev();
+				let prev_hop_iter = core::iter::once(payee).chain(
+					route.0.iter().skip(1).rev().map(|hop| &hop.src_node_id));
+				let mut hop_used = true;
+				let mut aggregate_next_hops_fee_msat: u64 = 0;
+				let mut aggregate_next_hops_path_htlc_minimum_msat: u64 = 0;
+
+				for (idx, (hop, prev_hop_id)) in hop_iter.zip(prev_hop_iter).enumerate() {
+					// BOLT 11 doesn't allow inclusion of features for the last hop hints, which
+					// really sucks, cause we're gonna need that eventually.
+					let hop_htlc_minimum_msat: u64 = hop.htlc_minimum_msat.unwrap_or(0);
+
+					let directional_info = DummyDirectionalChannelInfo {
+						cltv_expiry_delta: hop.cltv_expiry_delta as u32,
+						htlc_minimum_msat: hop_htlc_minimum_msat,
+						htlc_maximum_msat: hop.htlc_maximum_msat,
+						fees: hop.fees,
+					};
+
+					let reqd_channel_cap = if let Some (val) = final_value_msat.checked_add(match idx {
+						0 => 999,
+						_ => aggregate_next_hops_fee_msat.checked_add(999).unwrap_or(u64::max_value())
+					}) { Some( val / 1000 ) } else { break; }; // converting from msat or breaking if max ~ infinity
+
+
+					// We assume that the recipient only included route hints for routes which had
+					// sufficient value to route `final_value_msat`. Note that in the case of "0-value"
+					// invoices where the invoice does not specify value this may not be the case, but
+					// better to include the hints than not.
+					if !add_entry!(hop.short_channel_id, hop.src_node_id, prev_hop_id, directional_info, reqd_channel_cap, &empty_channel_features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat) {
+						// If this hop was not used then there is no use checking the preceding hops
+						// in the RouteHint. We can break by just searching for a direct channel between
+						// last checked hop and first_hop_targets
+						hop_used = false;
+					}
+
+					// Searching for a direct channel between last checked hop and first_hop_targets
+					if let Some(&(ref first_hop, ref features, ref outbound_capacity_msat, _)) = first_hop_targets.get(&prev_hop_id) {
+						add_entry!(first_hop, *our_node_id , prev_hop_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat);
+					}
+
+					if !hop_used {
+						break;
+					}
+
+					// In the next values of the iterator, the aggregate fees already reflects
+					// the sum of value sent from payer (final_value_msat) and routing fees
+					// for the last node in the RouteHint. We need to just add the fees to
+					// route through the current node so that the preceeding node (next iteration)
+					// can use it.
+					let hops_fee = compute_fees(aggregate_next_hops_fee_msat + final_value_msat, hop.fees)
+						.map_or(None, |inc| inc.checked_add(aggregate_next_hops_fee_msat));
+					aggregate_next_hops_fee_msat = if let Some(val) = hops_fee { val } else { break; };
+
+					let hop_htlc_minimum_msat_inc = if let Some(val) = compute_fees(aggregate_next_hops_path_htlc_minimum_msat, hop.fees) { val } else { break; };
+					let hops_path_htlc_minimum = aggregate_next_hops_path_htlc_minimum_msat
+						.checked_add(hop_htlc_minimum_msat_inc);
+					aggregate_next_hops_path_htlc_minimum_msat = if let Some(val) = hops_path_htlc_minimum { cmp::max(hop_htlc_minimum_msat, val) } else { break; };
+
+					if idx == route.0.len() - 1 {
+						// The last hop in this iterator is the first hop in
+						// overall RouteHint.
+						// If this hop connects to a node with which we have a direct channel,
+						// ignore the network graph and, if the last hop was added, add our
+						// direct channel to the candidate set.
+						//
+						// Note that we *must* check if the last hop was added as `add_entry`
+						// always assumes that the third argument is a node to which we have a
+						// path.
+						if let Some(&(ref first_hop, ref features, ref outbound_capacity_msat, _)) = first_hop_targets.get(&hop.src_node_id) {
+							add_entry!(first_hop, *our_node_id , hop.src_node_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat);
+						}
 					}
 				}
 			}
@@ -1343,7 +1394,7 @@ mod tests {
 		let logger = Arc::new(test_utils::TestLogger::new());
 		let chain_monitor = Arc::new(test_utils::TestChainSource::new(Network::Testnet));
 		let net_graph_msg_handler = NetGraphMsgHandler::new(genesis_block(Network::Testnet).header.block_hash(), None, Arc::clone(&logger));
-		// Build network from our_id to node7:
+		// Build network from our_id to node6:
 		//
 		//        -1(1)2-  node0  -1(3)2-
 		//       /                       \
@@ -1378,6 +1429,8 @@ mod tests {
 		// node2--1(6)2- node4 -1(9)2--- node6 (not in global route map)
 		//      \                      /
 		//       -1(7)2- node5 -1(10)2-
+		//
+		// Channels 5, 8, 9 and 10 are private channels.
 		//
 		// chan5  1-to-2: enabled, 100 msat fee
 		// chan5  2-to-1: enabled, 0 fee
@@ -2105,7 +2158,51 @@ mod tests {
 			cltv_expiry_delta: (8 << 8) | 1,
 			htlc_minimum_msat: None,
 			htlc_maximum_msat: None,
+		}
+		]), RouteHint(vec![RouteHintHop {
+			src_node_id: nodes[4].clone(),
+			short_channel_id: 9,
+			fees: RoutingFees {
+				base_msat: 1001,
+				proportional_millionths: 0,
+			},
+			cltv_expiry_delta: (9 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
 		}]), RouteHint(vec![RouteHintHop {
+			src_node_id: nodes[5].clone(),
+			short_channel_id: 10,
+			fees: zero_fees,
+			cltv_expiry_delta: (10 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
+		}])]
+	}
+
+	fn last_hops_multi_private_channels(nodes: &Vec<PublicKey>) -> Vec<RouteHint> {
+		let zero_fees = RoutingFees {
+			base_msat: 0,
+			proportional_millionths: 0,
+		};
+		vec![RouteHint(vec![RouteHintHop {
+			src_node_id: nodes[2].clone(),
+			short_channel_id: 5,
+			fees: RoutingFees {
+				base_msat: 100,
+				proportional_millionths: 0,
+			},
+			cltv_expiry_delta: (5 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
+		}, RouteHintHop {
+			src_node_id: nodes[3].clone(),
+			short_channel_id: 8,
+			fees: zero_fees,
+			cltv_expiry_delta: (8 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
+		}
+		]), RouteHint(vec![RouteHintHop {
 			src_node_id: nodes[4].clone(),
 			short_channel_id: 9,
 			fees: RoutingFees {
@@ -2126,11 +2223,13 @@ mod tests {
 	}
 
 	#[test]
-	fn last_hops_test() {
+	fn partial_route_hint_test() {
 		let (secp_ctx, net_graph_msg_handler, _, logger) = build_graph();
 		let (_, our_id, _, nodes) = get_nodes(&secp_ctx);
 
 		// Simple test across 2, 3, 5, and 4 via a last_hop channel
+		// Tests the behaviour when the RouteHint contains a suboptimal hop.
+		// RouteHint may be partially used by the algo to build the best path.
 
 		// First check that last hop can't have its source as the payee.
 		let invalid_last_hop = RouteHint(vec![RouteHintHop {
@@ -2145,7 +2244,7 @@ mod tests {
 			htlc_maximum_msat: None,
 		}]);
 
-		let mut invalid_last_hops = last_hops(&nodes);
+		let mut invalid_last_hops = last_hops_multi_private_channels(&nodes);
 		invalid_last_hops.push(invalid_last_hop);
 		{
 			if let Err(LightningError{err, action: ErrorAction::IgnoreError}) = get_route(&our_id, &net_graph_msg_handler.network_graph.read().unwrap(), &nodes[6], None, None, &invalid_last_hops.iter().collect::<Vec<_>>(), 100, 42, Arc::clone(&logger)) {
@@ -2153,7 +2252,7 @@ mod tests {
 			} else { panic!(); }
 		}
 
-		let route = get_route(&our_id, &net_graph_msg_handler.network_graph.read().unwrap(), &nodes[6], None, None, &last_hops(&nodes).iter().collect::<Vec<_>>(), 100, 42, Arc::clone(&logger)).unwrap();
+		let route = get_route(&our_id, &net_graph_msg_handler.network_graph.read().unwrap(), &nodes[6], None, None, &last_hops_multi_private_channels(&nodes).iter().collect::<Vec<_>>(), 100, 42, Arc::clone(&logger)).unwrap();
 		assert_eq!(route.paths[0].len(), 5);
 
 		assert_eq!(route.paths[0][0].pubkey, nodes[1]);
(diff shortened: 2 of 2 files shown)


lightning/src/routing/router.rs
```diff
@@ -876,40 +878,126 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+						if !add_entry!(hop.short_channel_id, hop.src_node_id, prev_hop, directional_info, reqd_channel_cap, &empty_channel_features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat) {
+							hop_used = false;
+						}
+					}
+
+					if !hop_used {
+						// If this hop was not used then there is no use checking the preceeding hops
+						// in the RouteHint. We can break by just searching for a direct channel between 
+						// last checked hop and first_hop_targets
+						if let Some(&(ref first_hop, ref features, ref outbound_capacity_msat, _)) = first_hop_targets.get(&prev_hop) {
```
TheBlueMatt, line 952: Why is this inside the `hop_used` branch? If we just move it outside of the if we can drop the `idx == route.len() - 1` if as well, no? I mean we could probably also drop this hunk entirely and it wouldn't matter, but its simpler if we just move it out of the if.

lightning/src/routing/router.rs
```diff
@@ -876,40 +878,126 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+							add_entry!(first_hop, *our_node_id , hop.src_node_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat);
+						}
 					}
+
+					aggregate_next_hops_fee_msat = match aggregate_next_hops_fee_msat.checked_add(match idx {
+
+						// When its the Last Hop in the RouteHint, we want the aggregate fees to
+						// be the sum of the final value to route and the fees to route through
+						// the channel between the Last Hop Node and the Payee Node.
+						0 => match compute_fees(aggregate_next_hops_fee_msat, hop.fees) {
```
TheBlueMatt, line 978: This and the `_` block are identical?

lightning/src/routing/router.rs
```diff
@@ -876,40 +878,126 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+
+					prev_hop = match idx {
+						0 => &payee,
+						_ => match prev_route_it.next() {
+							Some(x) => &x.src_node_id,
+							None => &payee,
+						}
+					};
+
+					if idx == 0 {
```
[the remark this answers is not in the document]
abhik-99: Because aggregate_next_hops_fee_msat needs to be zero at the last RouteHintHop which connects to the payee. Since we are building a path from payee then the first value of aggregate_next_hops_fee_msat has to be 0 since no fees will be charged for routing after that hop.

lightning/src/routing/router.rs
abhik-99: Yes, for case 0 I was adding the final_value_msat here but then switched to do it inside the add_entry's if block for idx == 0. New commit reflects this change.

lightning/src/routing/router.rs
abhik-99: The reason is if the current hop being assessed was not used then the direct channel needs to be created from the previously added hop. If the current hop is the last hop then that would mean from payee to the ```first_hop_targets```. However, if ```idx == route.len() - 1``` was triggered then this hop was used and in this case the direct channel will be from this hop to first_hop_targets.

abhik-99: If a particular hop in a ```RouteHint``` is not used then the loop is broken. But ```idx == route.len() - 1``` is only triggered if all the ```RouteHintHop```s are used.

lightning/src/routing/router.rs
```diff
@@ -375,10 +375,12 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		return Err(LightningError{err: "Cannot send a payment of 0 msat".to_owned(), action: ErrorAction::IgnoreError});
 	}
 
-	let last_hops = last_hops.iter().filter_map(|hops| hops.0.last()).collect::<Vec<_>>();
-	for last_hop in last_hops.iter() {
-		if last_hop.src_node_id == *payee {
-			return Err(LightningError{err: "Last hop cannot have a payee as a source.".to_owned(), action: ErrorAction::IgnoreError});
+	let last_hops = last_hops.iter().map(|hops| hops.0.clone()).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
abhik-99: Yeah, It has been removed in commit [396e535](https://github.com/rust-bitcoin/rust-lightning/pull/1040/commits/396e53591427ce6c63a2fe90fd24be16b2e516b8).

lightning/src/routing/router.rs
TheBlueMatt: Hmm, right. Its still a bit simpler if we move the pre-`break` `add_entry!` call out of the `else` block and just always do it, though.

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					if idx == route.0.len() - 1 {
+						// The last hop in this iterator is the first hop in
+						// overall RouteHint.
+						// If this hop connects to a node with which we have a direct channel,
+						// ignore the network graph and, if the last hop was added, add our
+						// direct channel to the candidate set.
+						//
+						// Note that we *must* check if the last hop was added as `add_entry`
+						// always assumes that the third argument is a node to which we have a
+						// path.\
```
TheBlueMatt, line 963: nit: spare `\` at EOL here.

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+						_ => match compute_fees(aggregate_next_hops_fee_msat, hop.fees) {
+							Some(fees_msat) => fees_msat,
+							None => 0
+							}
+					}) {
+						Some(fees_msat) => fees_msat,
+						None =>  u64::max_value()
+					};
+
+					aggregate_next_hops_path_htlc_minimum_msat = match aggregate_next_hops_path_htlc_minimum_msat.checked_add(hop_htlc_minimum_msat) {
```
TheBlueMatt, line 985: The HTLC minimum is not increased at each hop, its the minimum *at* that hop. The value from previous hops needs to be increased by the feerate to meet the minimum at that hop, but you dont add the new hop's min. The way we calculate it in the `add_entry` loop is:

```
						// Note that low contribution here (limited by available_liquidity_msat)
						// might violate htlc_minimum_msat on the hops which are next along the
						// payment path (upstream to the payee). To avoid that, we recompute path
						// path fees knowing the final path contribution after constructing it.
						let path_htlc_minimum_msat = match compute_fees($next_hops_path_htlc_minimum_msat, $directional_info.fees)
								.map(|fee_msat| fee_msat.checked_add($next_hops_path_htlc_minimum_msat)) {
							Some(Some(value_msat)) => cmp::max(value_msat, $directional_info.htlc_minimum_msat),
							_ => u64::max_value()
						};
```

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+						// overall RouteHint.
+						// If this hop connects to a node with which we have a direct channel,
+						// ignore the network graph and, if the last hop was added, add our
+						// direct channel to the candidate set.
+						//
+						// Note that we *must* check if the last hop was added as `add_entry`
+						// always assumes that the third argument is a node to which we have a
+						// path.\
+						if let Some(&(ref first_hop, ref features, ref outbound_capacity_msat, _)) = first_hop_targets.get(&hop.src_node_id) {
+							add_entry!(first_hop, *our_node_id , hop.src_node_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat);
```
TheBlueMatt, line 965: Don't these fields need to be the ones from after the `aggregate_next_hops_fee_msat` update? ie why is `aggregate_next_hops_fee_msat` not considering the last `last_hops` hop in its next_hops_fee?

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					}
+
+					aggregate_next_hops_fee_msat = match aggregate_next_hops_fee_msat.checked_add(match idx {
+
+						// In the next values of the iterator, the aggregate fees already reflects
+						// the sum of value sent from payer (final_value_msat) and routing fees
+						// for the last node in the RouteHint. We need to just add the fees to
+						// route through the current node so that the preceeding node (next iteration)
+						// can use it.
+						_ => match compute_fees(aggregate_next_hops_fee_msat, hop.fees) {
```
TheBlueMatt, line 976: Why `match` on `idx` when you always hit this branch? :)

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					// invoices where the invoice does not specify value this may not be the case, but
+					// better to include the hints than not.
+					if add_entry!(hop.short_channel_id, hop.src_node_id, prev_hop, directional_info, reqd_channel_cap, &empty_channel_features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat) {
+
+						// If the index is 0, then its the last hop in the RoutHint. The
+						// aggregate_next_hops_fee_msat for this hop was 0 but from here on out
+						// it needs to be increased beyond the value to be routed to account
+						// for the fee charged at hops and value to send from payer to payee.
+						if idx == 0 {
+							aggregate_next_hops_fee_msat = match aggregate_next_hops_fee_msat.checked_add(final_value_msat) {
```
TheBlueMatt, line 939: Hmm, I may have been confusing in my previous review - `aggregate_next_hops_fee_msat` should not *include* `final_value_msat`, it should always be only the fee. However, the calls to `compute_fee` should include `final_value_msat` because the fees are always calculated against the total value at a hop, not against just the fee portion.

TheBlueMatt (commented):
This is a lot less logic to review compared to the last version, thanks. Makes it much easier to find bugs when there's fewer branches and ifs in the code :).

lightning/src/routing/router.rs
abhik-99: Regarding moving pre-```break``` ```add_entry!```  and out doing it always, that would mean checking for a direct channel between all node hops in ```RouteHint``` and I would also need to use another variable (eg, ```hop_used```) just to track if the hop was used or not and then break accordingly out of the loop after the ```add_entry!``` with the previous ```RouteHintHop```.

lightning/src/routing/router.rs
abhik-99: ```aggregate_next_hops_fee_msat``` is ```u64``` but ```compute_fees``` returns ```Some``` or ```None```. That's why I am matching it so that I can bring out the value only.

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					let directional_info = DummyDirectionalChannelInfo {
+						cltv_expiry_delta: hop.cltv_expiry_delta as u32,
+						htlc_minimum_msat: hop_htlc_minimum_msat,
+						htlc_maximum_msat: hop.htlc_maximum_msat,
+						fees: hop.fees,
+					};
+
+					prev_hop = match idx {
+						0 => &payee,
+						_ => match prev_route_it.next() {
```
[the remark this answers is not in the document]
abhik-99: Yup.

lightning/src/routing/router.rs
abhik-99: You are right. It escaped my mind. Adding in that modification.

lightning/src/routing/router.rs
abhik-99: Okay. I think I understand that. Adding that modification. I noticed it now that the test ```last_hops_multi_hint_with_disabled_channels_test``` fails with this modification even though it's supposed to find a path to  ```node6`` through channels ```5``` and ```8```. It errors out with ```{ err: "Failed to find a path to the given destination", action: IgnoreError }', lightning/src/routing/router.rs:2468:228```

lightning/src/routing/router.rs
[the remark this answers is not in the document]
jkczyz: Yeah, I think you can dispense with `prev_route_it` entirely and do:

```rust
prev_hop = if idx == 0 { &payee } else { &hop.src_node_id };
```

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
-						add_entry!(first_hop, *our_node_id , hop.src_node_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, 0, path_value_msat, 0);
+				// We start building the path from reverse, i.e., from payee
+				// to the first RouteHintHop in the path.
+				let route_it = route.0.iter().rev();
+				let mut prev_route_it = route.0.iter().rev();
+				let mut aggregate_next_hops_fee_msat: u64 = 0;
+				let mut aggregate_next_hops_path_htlc_minimum_msat: u64 = 0;
+				let mut prev_hop;
+
+				for (idx, hop) in route_it.enumerate(){
```
jkczyz, line 897: nit: space needed before `{`

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					if add_entry!(hop.short_channel_id, hop.src_node_id, prev_hop, directional_info, reqd_channel_cap, &empty_channel_features, aggregate_next_hops_fee_msat, path_value_msat, aggregate_next_hops_path_htlc_minimum_msat) {
+
+						// If the index is 0, then its the last hop in the RoutHint. The
+						// aggregate_next_hops_fee_msat for this hop was 0 but from here on out
+						// it needs to be increased beyond the value to be routed to account
+						// for the fee charged at hops and value to send from payer to payee.
+						if idx == 0 {
+							aggregate_next_hops_fee_msat = match aggregate_next_hops_fee_msat.checked_add(final_value_msat) {
+							Some(value) => value,
+							None => u64::max_value()
```
jkczyz, line 941: Add indentation

lightning/src/routing/router.rs
```diff
@@ -876,40 +877,115 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+							};
+						}
+					} else {
+						// If this hop was not used then there is no use checking the preceeding hops
```
jkczyz, line 945: s/preceeding/preceding

lightning/src/routing/router.rs
jkczyz: A one-line way of doing this is:

```rust
compute_fees(aggregate_next_hops_fee_msat, hop.fees).unwrap_or(0)
```

Likewise for the result of `checked_add`.

jkczyz (commented):
Just some minor comments for now. Will do a more thorough review once you push your most recent changes.

lightning/src/routing/router.rs
TheBlueMatt: Hmm, yea, it does feel clearer if you do that, I think. If we are going to look up first-hop connections in some cases, we might as well look it up in all cases.

lightning/src/routing/router.rs
```diff
@@ -876,39 +877,85 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
-					// direct channel to the candidate set.
-					//
-					// Note that we *must* check if the last hop was added as `add_entry`
-					// always assumes that the third argument is a node to which we have a
-					// path.
-					if let Some(&(ref first_hop, ref features, ref outbound_capacity_msat, _)) = first_hop_targets.get(&hop.src_node_id) {
-						add_entry!(first_hop, *our_node_id , hop.src_node_id, dummy_directional_info, Some(outbound_capacity_msat / 1000), features, 0, path_value_msat, 0);
+				// We start building the path from reverse, i.e., from payee
+				// to the first RouteHintHop in the path.
+				let rev_route = route.0.iter().rev().collect::<Vec<_>>();
```
TheBlueMatt, line 891: Why allocate this as a vec just to index backwards into it? Just access `route.0` in the `prev_hop` calculation, below.

lightning/src/routing/router.rs
```diff
@@ -876,39 +877,85 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					if !hop_used {
+						break;
+					}
+
+					// In the next values of the iterator, the aggregate fees already reflects
+					// the sum of value sent from payer (final_value_msat) and routing fees
+					// for the last node in the RouteHint. We need to just add the fees to
+					// route through the current node so that the preceeding node (next iteration)
+					// can use it.
+					aggregate_next_hops_fee_msat = aggregate_next_hops_fee_msat.checked_add(compute_fees(aggregate_next_hops_fee_msat + final_value_msat, hop.fees).unwrap_or(0)).unwrap_or(u64::max_value());
```
TheBlueMatt, line 942: I believe the `unwrap_or(0)`s here should be `unwrap_or(uXX::max_value())` as well - if we fail to calculate the fee at a hop its because we overflowed and should consider it maximum-fee, skipping any future hops on this path.

That said, for this and `aggregate_next_hops_path_htlc_minimum_msat` ,maybe it makes sense to not `unwrap_or` but instead do the calculation with a `compute_fees().map_or(None, |inc| inc.checked_add(aggregate_next_hops_fee_msat)`, like this (but with better variable names):

```suggestion
					let new_thing = compute_fees(aggregate_next_hops_fee_msat + final_value_msat, hop.fees).map_or(None, |inc| inc.checked_add(aggregate_next_hops_fee_msat));
					aggregate_next_hops_fee_msat = if let Some(val) = new_thing { val } else { break; };
```

lightning/src/routing/router.rs
```diff
@@ -876,39 +877,85 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+
+					aggregate_next_hops_path_htlc_minimum_msat = aggregate_next_hops_path_htlc_minimum_msat.checked_add(compute_fees(hop_htlc_minimum_msat, hop.fees).unwrap_or(0)).unwrap_or(u64::max_value());
```
TheBlueMatt, line 944: We need to calculate fees on the current backwards-walking path minimum, but `cmp::max` it with the current hop minimum, like this, I think:

```suggestion
					aggregate_next_hops_path_htlc_minimum_msat = cmp::max(hop_htlc_minimum_msat,
						aggregate_next_hops_path_htlc_minimum_msat.checked_add(
							compute_fees(aggregate_next_hops_path_htlc_minimum_msat, hop.fees).unwrap_or(u64::max_value())
						).unwrap_or(u64::max_value()));
```

TheBlueMatt (commented):
This is looking pretty good.

lightning/src/routing/router.rs
abhik-99: I was thinking that since this kind of comparison was already being done in the `add_entry` macro, it was not needed at this point.
But yes, you are right, we need to find the maximum of the two to satisfy to consider routing through the next hops. Committing the changes.

lightning/src/routing/router.rs
abhik-99: Yes, this has been changed in the latest commit.

lightning/src/routing/router.rs
abhik-99: I thought it would be easier this way. Would there be a problem with this?

jkczyz (commented):
Getting late. Will take a close look at the tests tomorrow.

lightning/src/routing/router.rs
```diff
@@ -876,39 +877,89 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					// for the last node in the RouteHint. We need to just add the fees to
+					// route through the current node so that the preceeding node (next iteration)
+					// can use it.
+					let hops_fee = compute_fees(aggregate_next_hops_fee_msat + final_value_msat, hop.fees).map_or(None, |inc| inc.checked_add(aggregate_next_hops_fee_msat));
+
+					aggregate_next_hops_fee_msat = if let Some(val) = hops_fee { val } else { break; };
+
+					let hops_path_htlc_minimum = aggregate_next_hops_path_htlc_minimum_msat.checked_add(compute_fees(hop_htlc_minimum_msat, hop.fees).unwrap_or(u64::max_value())).map_or(None, |inc| inc.checked_add(aggregate_next_hops_path_htlc_minimum_msat));
+
+					aggregate_next_hops_path_htlc_minimum_msat =  if let Some(val) = hops_path_htlc_minimum { cmp::max(hop_htlc_minimum_msat, val) } else { break; };
```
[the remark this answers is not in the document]
abhik-99: Sure!

lightning/src/routing/router.rs
```diff
@@ -876,39 +877,89 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		// If a caller provided us with last hops, add them to routing targets. Since this happens
 		// earlier than general path finding, they will be somewhat prioritized, although currently
 		// it matters only if the fees are exactly the same.
-		for hop in last_hops.iter() {
+		for route in last_hops.iter() {
+			let first_hop_in_route = &(route.0)[0];
```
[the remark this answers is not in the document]
abhik-99: Including a simple `if` check to make sure no empty `RouteHint`s are considered.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
abhik-99: Commit [9d0b1](https://github.com/rust-bitcoin/rust-lightning/pull/1040/commits/9d0b17971bae409429f9750ba479f5c5499de414) does this.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
abhik-99: Commit [9d0b1](https://github.com/rust-bitcoin/rust-lightning/pull/1040/commits/9d0b17971bae409429f9750ba479f5c5499de414) does this.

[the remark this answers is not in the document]
jkczyz: Could you use `filter` instead to avoid adding another level of indenting?

```rust
for route in last_hops.iter().filter(|route| !route.0.is_empty()) {
```

lightning/src/routing/router.rs
```diff
@@ -2194,6 +2253,326 @@ mod tests {
...
+			cltv_expiry_delta: (11 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
+		}, RouteHintHop {
+			src_node_id: nodes[3].clone(),
+			short_channel_id: 8,
+			fees: zero_fees,
+			cltv_expiry_delta: (8 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
```
jkczyz, line 2367: Is `last_hops_multi_hint_test` testing that a mutli-hop hint containing a public channel is used? And is `multi_hint_with_disabled_channels_last_hops` testing that a mutli-hop containing two private channels is used? If so, could you name the tests to reflect that? As of now, the second test seems to be named after how to trigger the behavior rather than after the behavior itself, IIUC.

jkczyz (commented):
Noted some more ways to clean up the code and tests IIUC what's being tested (see comments). There's also some earlier comments that have not yet been addressed. Could you go through to get everything that was missed?

lightning/src/routing/router.rs
```diff
@@ -876,39 +879,93 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+					// Only add the hops in this route to our candidate set if either
+					// we have a direct channel to the first hop or the first hop is
+					// in the regular network graph.
+					first_hop_targets.get(&first_hop_in_route.src_node_id).is_some() ||
+					network.get_nodes().get(&first_hop_in_route.src_node_id).is_some();
+				if have_hop_src_in_graph {
+					// We start building the path from reverse, i.e., from payee
+					// to the first RouteHintHop in the path.
+					let rev_route = route.0.iter().rev().collect::<Vec<_>>();
+					let route_it = route.0.iter().rev();
```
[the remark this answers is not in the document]
abhik-99: Sure

lightning/src/routing/router.rs
```diff
@@ -878,87 +880,91 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
...
+						// can use it.
+						let hops_fee = compute_fees(aggregate_next_hops_fee_msat + final_value_msat, hop.fees)
+						.map_or(None, |inc| inc.checked_add(aggregate_next_hops_fee_msat));
+						aggregate_next_hops_fee_msat = if let Some(val) = hops_fee { val } else { break; };
+
+						let hop_htlc_minimum_msat_inc = compute_fees(hop_htlc_minimum_msat, hop.fees).unwrap_or(u64::max_value());
+						let hops_path_htlc_minimum = aggregate_next_hops_path_htlc_minimum_msat
+						.checked_add(hop_htlc_minimum_msat_inc)
+						.map_or(None, |inc| inc.checked_add(aggregate_next_hops_path_htlc_minimum_msat));
+						aggregate_next_hops_path_htlc_minimum_msat =  if let Some(val) = hops_path_htlc_minimum { cmp::max(hop_htlc_minimum_msat, val) } else { break; };
```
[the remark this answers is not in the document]
abhik-99: Ok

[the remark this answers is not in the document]
abhik-99: Yup

lightning/src/routing/router.rs
```diff
@@ -2247,6 +2253,99 @@ mod tests {
...
+			htlc_maximum_msat: None,
+		}]);
+
+		let mut invalid_last_hops = empty_last_hops(&nodes);
+		invalid_last_hops.push(invalid_last_hop);
+		{
+			if let Err(LightningError{err, action: ErrorAction::IgnoreError}) = get_route(&our_id, &net_graph_msg_handler.network_graph.read().unwrap(), &nodes[6], None, None, &invalid_last_hops.iter().collect::<Vec<_>>(), 100, 42, Arc::clone(&logger)) {
+				assert_eq!(err, "Last hop cannot have a payee as a source.");
+			} else { panic!(); }
+		}
```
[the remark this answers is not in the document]
abhik-99: I was actually modifying based on the `last_hops_test`. I'm just sticking to testing the checking of invalid hops to `last_hops_test` only then.

lightning/src/routing/router.rs
```diff
@@ -2247,6 +2253,99 @@ mod tests {
...
+			short_channel_id: 10,
+			fees: zero_fees,
+			cltv_expiry_delta: (10 << 8) | 1,
+			htlc_minimum_msat: None,
+			htlc_maximum_msat: None,
+		}])]
+	}
+
+	#[test]
+	fn empty_last_hops_test() {
```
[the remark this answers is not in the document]
abhik-99: Sure

lightning/src/routing/router.rs
```diff
@@ -375,10 +375,13 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		return Err(LightningError{err: "Cannot send a payment of 0 msat".to_owned(), action: ErrorAction::IgnoreError});
 	}
 
-	let last_hops = last_hops.iter().filter_map(|hops| hops.0.last()).collect::<Vec<_>>();
-	for last_hop in last_hops.iter() {
-		if last_hop.src_node_id == *payee {
-			return Err(LightningError{err: "Last hop cannot have a payee as a source.".to_owned(), action: ErrorAction::IgnoreError});
+	for route in last_hops.iter() {
+		if route.0.len() > 0 {
```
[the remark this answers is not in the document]
abhik-99: The code was raising index out of bounds errors for that. This check guards against that.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
abhik-99: The latest commit has removed the invalid hops test in the successive tests which were added. As for creating inline helpers for hints, I feel the code is bit cleaner with the format used in `last_hops_test`. However, if needed, I can remove it in successive commit.

lightning/src/routing/router.rs
abhik-99: There are 3 tests:-
1. Checking empty hints are ignored - `ignores_empty_last_hops_test`.
2. Multi-hint last hops with all private channels - `multi_hint_last_hops_test`.
3. Multi-hint last hops test with one public channel used in hints - `last_hops_with_public_channel_test`.

lightning/src/routing/router.rs
abhik-99: From commit [0c0985](https://github.com/rust-bitcoin/rust-lightning/pull/1040/commits/0c09852311b3b7c42d1210aa69aa3f827a91655d) all hops are checked for a direct channel.

lightning/src/routing/router.rs
abhik-99: `rev_route` is supposed to stay one behind the current iterator value. It would have been much easier to use [prev_iter](https://docs.rs/prev-iter/0.1.2/prev_iter/struct.PrevPeekable.html) but I went without using it. If I zip the two iterators then both of them would essentially give the same value. [`reverse()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reverse) might have worked but it does the operation in place which I thought might cause complications.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
abhik-99: As of commit [0c09852](https://github.com/rust-bitcoin/rust-lightning/pull/1040/commits/0c09852311b3b7c42d1210aa69aa3f827a91655d) this became redundant because I used `.filter(|route| !route.0.is_empty())` as advised.

lightning/src/routing/router.rs
```diff
@@ -2194,6 +2248,262 @@ mod tests {
...
+			htlc_maximum_msat: None,
+		}])]
+	}
+
+	#[test]
+	fn last_hops_with_public_channel_test() {
+		let (secp_ctx, net_graph_msg_handler, _, logger) = build_graph();
+		let (_, our_id, _, nodes) = get_nodes(&secp_ctx);
+		// This test shows that public routes can be present in the invoice
+		// which would be handled in the same manner.
```
jkczyz, line 2464: Should we also test the case where there is a multi-hint last hops composed entirely of private channels, but we chose a public channel and only one of the private channels? Or does the algorithm avoid that?

lightning/src/routing/router.rs
```diff
@@ -3999,4 +4309,4 @@ mod benches {
 			idx += 1;
 		});
 	}
-}
+}
```
jkczyz, line 4312: Looks like your editor may be stripping off the trailing newline character?

.gitignore
```diff
@@ -7,4 +7,4 @@ lightning-c-bindings/a.out
 **/*.rs.bk
 Cargo.lock
 .idea
-
+lightning/target
```
jkczyz, line 10: Is this needed if you open the editor at the root of the repository rather than the `lightning` directory?

lightning/src/routing/router.rs
```diff
@@ -375,10 +375,11 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 		return Err(LightningError{err: "Cannot send a payment of 0 msat".to_owned(), action: ErrorAction::IgnoreError});
 	}
 
-	let last_hops = last_hops.iter().filter_map(|hops| hops.0.last()).collect::<Vec<_>>();
-	for last_hop in last_hops.iter() {
-		if last_hop.src_node_id == *payee {
-			return Err(LightningError{err: "Last hop cannot have a payee as a source.".to_owned(), action: ErrorAction::IgnoreError});
+	for route in last_hops.iter().filter(|route| !route.0.is_empty()) {
```
jkczyz, line 378: This works without the filter because the inner `for` loop will not execute if `route.0` is empty.

.gitignore
abhik-99: Yup. `lightning/target` was not being ignored even when open from the project root.

lightning/src/routing/router.rs
abhik-99: `last_hops_test` has been repurposed to test the scenario of multiple private `RouteHintHop`s in a `RouteHint` which may not result in the best path. The test shows that the algo partially uses the `RouteHint` along with a public channel to construct the best path.

abhik-99:
@TheBlueMatt and @jkczyz any more suggestions?

lightning/src/routing/router.rs
```diff
@@ -2145,15 +2245,87 @@ mod tests {
...
-		let mut invalid_last_hops = last_hops(&nodes);
+		let mut invalid_last_hops = last_hops_multi_private_channels(&nodes);
 		invalid_last_hops.push(invalid_last_hop);
 		{
 			if let Err(LightningError{err, action: ErrorAction::IgnoreError}) = get_route(&our_id, &net_graph_msg_handler.network_graph.read().unwrap(), &nodes[6], None, None, &invalid_last_hops.iter().collect::<Vec<_>>(), 100, 42, Arc::clone(&logger)) {
 				assert_eq!(err, "Last hop cannot have a payee as a source.");
 			} else { panic!(); }
 		}
 
-		let route = get_route(&our_id, &net_graph_msg_handler.network_graph.read().unwrap(), &nodes[6], None, None, &last_hops(&nodes).iter().collect::<Vec<_>>(), 100, 42, Arc::clone(&logger)).unwrap();
```
TheBlueMatt, line 2156: This test appears to pass both with the multi_private version and not, so maybe lets keep both versions of the test?

TheBlueMatt (commented):
A few hiccups, but ACK aside from the comments.

lightning/src/routing/router.rs
```diff
@@ -907,10 +907,14 @@ pub fn get_route<L: Deref>(our_node_id: &PublicKey, network: &NetworkGraph, paye
 						fees: hop.fees,
 					};
 
-					let reqd_channel_cap = Some(final_value_msat.checked_add(match idx {
+					// let reqd_channel_cap = Some(final_value_msat.checked_add(match idx {
+					// 	0 => 999,
+					// 	_ => aggregate_next_hops_fee_msat.checked_add(999).unwrap_or(u64::max_value())
+					// }).unwrap_or(u64::max_value()) / 1000); // converting from msat
```
jkczyz, line 913: Could you remove the commented  code?

jkczyz (approved):

TheBlueMatt:
Cool! Can you squash the commits down to not have fixes in followup commits? https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits should be helpful if you have never done this.

jkczyz:
ACK fd4d3bf3b76a9490c93d75a85e599989063b6575

PR #1043, opened 2021-08-14
Handle network updates from failed payments in BackgroundProcessor

jkczyz:
`MessageSendEvent::PaymentFailureNetworkUpdate` served as a hack to pass an `HTLCFailChannelUpdate` from `ChannelManager` to `NetGraphMsgHandler` via `PeerManager`. Instead, remove the event entirely and move the contained
data (renamed `NetworkUpdate`) into `Event::PaymentFailed` to be processed by an event handler.

This requires making `NetGraphMsgHandler` an `EventHandler`, which is used in `BackgroundProcessor` to decorate the user-provided `EventHandler` with the ability to update the `NetworkGraph` accordingly.

The general concept of intercepting events via a decorator will be used in a follow-up PR for scoring channels and retrying failed payments.

### DIFF
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -345,7 +345,7 @@ mod tests {
 			begin_open_channel!($node_a, $node_b, $channel_value);
 			let events = $node_a.node.get_and_clear_pending_events();
 			assert_eq!(events.len(), 1);
-			let (temporary_channel_id, tx) = handle_funding_generation_ready!(events[0], $channel_value);
+			let (temporary_channel_id, tx) = handle_funding_generation_ready!(&events[0], $channel_value);
 			end_open_channel!($node_a, $node_b, temporary_channel_id, tx);
 			tx
 		}}
@@ -362,14 +362,14 @@ mod tests {
 	macro_rules! handle_funding_generation_ready {
 		($event: expr, $channel_value: expr) => {{
 			match $event {
-				Event::FundingGenerationReady { ref temporary_channel_id, ref channel_value_satoshis, ref output_script, user_channel_id } => {
-					assert_eq!(*channel_value_satoshis, $channel_value);
+				&Event::FundingGenerationReady { temporary_channel_id, channel_value_satoshis, ref output_script, user_channel_id } => {
+					assert_eq!(channel_value_satoshis, $channel_value);
 					assert_eq!(user_channel_id, 42);
 
 					let tx = Transaction { version: 1 as i32, lock_time: 0, input: Vec::new(), output: vec![TxOut {
-						value: *channel_value_satoshis, script_pubkey: output_script.clone(),
+						value: channel_value_satoshis, script_pubkey: output_script.clone(),
 					}]};
-					(*temporary_channel_id, tx)
+					(temporary_channel_id, tx)
 				},
 				_ => panic!("Unexpected event"),
 			}
@@ -423,7 +423,7 @@ mod tests {
 		// Initiate the background processors to watch each node.
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = move |node: &ChannelManager<InMemorySigner, Arc<ChainMonitor>, Arc<test_utils::TestBroadcaster>, Arc<KeysManager>, Arc<test_utils::TestFeeEstimator>, Arc<test_utils::TestLogger>>| FilesystemPersister::persist_manager(data_dir.clone(), node);
-		let event_handler = |_| {};
+		let event_handler = |_: &_| {};
 		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
 
 		macro_rules! check_persisted_data {
@@ -476,7 +476,7 @@ mod tests {
 		let nodes = create_nodes(1, "test_timer_tick_called".to_string());
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = move |node: &ChannelManager<InMemorySigner, Arc<ChainMonitor>, Arc<test_utils::TestBroadcaster>, Arc<KeysManager>, Arc<test_utils::TestFeeEstimator>, Arc<test_utils::TestLogger>>| FilesystemPersister::persist_manager(data_dir.clone(), node);
-		let event_handler = |_| {};
+		let event_handler = |_: &_| {};
 		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
 		loop {
 			let log_entries = nodes[0].logger.lines.lock().unwrap();
@@ -498,7 +498,7 @@ mod tests {
 		open_channel!(nodes[0], nodes[1], 100000);
 
 		let persister = |_: &_| Err(std::io::Error::new(std::io::ErrorKind::Other, "test"));
-		let event_handler = |_| {};
+		let event_handler = |_: &_| {};
 		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
 		match bg_processor.join() {
 			Ok(_) => panic!("Expected error persisting manager"),
@@ -518,7 +518,7 @@ mod tests {
 
 		// Set up a background event handler for FundingGenerationReady events.
 		let (sender, receiver) = std::sync::mpsc::sync_channel(1);
-		let event_handler = move |event| {
+		let event_handler = move |event: &Event| {
 			sender.send(handle_funding_generation_ready!(event, channel_value)).unwrap();
 		};
 		let bg_processor = BackgroundProcessor::start(persister.clone(), event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
@@ -544,7 +544,7 @@ mod tests {
 
 		// Set up a background event handler for SpendableOutputs events.
 		let (sender, receiver) = std::sync::mpsc::sync_channel(1);
-		let event_handler = move |event| sender.send(event).unwrap();
+		let event_handler = move |event: &Event| sender.send(event.clone()).unwrap();
 		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
 
 		// Force close the channel and check that the SpendableOutputs event was handled.
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -17,7 +17,8 @@ use lightning::ln::channelmanager::ChannelManager;
 use lightning::ln::msgs::{ChannelMessageHandler, RoutingMessageHandler};
 use lightning::ln::peer_handler::{PeerManager, SocketDescriptor};
 use lightning::ln::peer_handler::CustomMessageHandler;
-use lightning::util::events::{EventHandler, EventsProvider};
+use lightning::routing::network_graph::NetGraphMsgHandler;
+use lightning::util::events::{Event, EventHandler, EventsProvider};
 use lightning::util::logger::Logger;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -99,6 +100,33 @@ ChannelManagerPersister<Signer, M, T, K, F, L> for Fun where
 	}
 }
 
+/// Decorates an [`EventHandler`] with common functionality provided by standard [`EventHandler`]s.
+struct DecoratingEventHandler<
+	E: EventHandler,
+	N: Deref<Target = NetGraphMsgHandler<A, L>>,
+	A: Deref,
+	L: Deref,
+>
+where A::Target: chain::Access, L::Target: Logger {
+	event_handler: E,
+	net_graph_msg_handler: Option<N>,
+}
+
+impl<
+	E: EventHandler,
+	N: Deref<Target = NetGraphMsgHandler<A, L>>,
+	A: Deref,
+	L: Deref,
+> EventHandler for DecoratingEventHandler<E, N, A, L>
+where A::Target: chain::Access, L::Target: Logger {
+	fn handle_event(&self, event: &Event) {
+		if let Some(event_handler) = &self.net_graph_msg_handler {
+			event_handler.handle_event(event);
+		}
+		self.event_handler.handle_event(event);
+	}
+}
+
 impl BackgroundProcessor {
 	/// Start a background thread that takes care of responsibilities enumerated in the [top-level
 	/// documentation].
@@ -121,8 +149,9 @@ impl BackgroundProcessor {
 	/// # Event Handling
 	///
 	/// `event_handler` is responsible for handling events that users should be notified of (e.g.,
-	/// payment failed). A user's [`EventHandler`] may be decorated with other handlers to implement
-	/// common functionality. See individual [`Event`]s for further details.
+	/// payment failed). [`BackgroundProcessor`] may decorate the given [`EventHandler`] with common
+	/// functionality implemented by other handlers.
+	/// * [`NetGraphMsgHandler`] if given will update the [`NetworkGraph`] based on payment failures.
 	///
 	/// [top-level documentation]: Self
 	/// [`join`]: Self::join
@@ -130,9 +159,10 @@ impl BackgroundProcessor {
 	/// [`ChannelManager`]: lightning::ln::channelmanager::ChannelManager
 	/// [`ChannelManager::write`]: lightning::ln::channelmanager::ChannelManager#impl-Writeable
 	/// [`FilesystemPersister::persist_manager`]: lightning_persister::FilesystemPersister::persist_manager
-	/// [`Event`]: lightning::util::events::Event
+	/// [`NetworkGraph`]: lightning::routing::network_graph::NetworkGraph
 	pub fn start<
 		Signer: 'static + Sign,
+		CA: 'static + Deref + Send + Sync,
 		CF: 'static + Deref + Send + Sync,
 		CW: 'static + Deref + Send + Sync,
 		T: 'static + Deref + Send + Sync,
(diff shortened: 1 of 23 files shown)


lightning/src/routing/network_graph.rs
```diff
@@ -103,6 +103,66 @@ impl_writeable_tlv_based_enum!(NetworkUpdate,
...
+impl<G: Deref<Target=NetworkGraph>, L: Deref, E: EventHandler> EventHandler for NetworkUpdateHandler<G, L, E>
+where L::Target: Logger {
+	fn handle_event(&self, event: Event) {
+		if let Event::PaymentFailed { payment_hash: _, rejected_by_dest: _, network_update, .. } = &event {
+			if let Some(network_update) = network_update {
+				self.handle_network_update(network_update);
+			}
+		}
+
+		self.inner.handle_event(event)
```
TheBlueMatt, line 162: Hmmmmmm, have we considered maybe just allowing events to be handled by multiple handlers? Passing them off to an inner handler works, I suppose, but its kinda awkward. I guess I'm not a huge fan of passing the events by reference instead of value, but most of them have no heap allocation, and eg `SpendableOutputs` should be pretty rare (requires on-chain txn).

TheBlueMatt (commented):
Concept ACK. Only did a high-level skim but I like it.

lightning/src/routing/network_graph.rs
```diff
@@ -616,13 +617,15 @@ impl Writeable for NetworkGraph {
 		write_ver_prefix!(writer, SERIALIZATION_VERSION, MIN_SERIALIZATION_VERSION);
 
 		self.genesis_hash.write(writer)?;
-		(self.channels.len() as u64).write(writer)?;
-		for (ref chan_id, ref chan_info) in self.channels.iter() {
+		let channels = self.channels.read().unwrap();
```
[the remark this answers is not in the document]
jkczyz: Ah, meant to update the docs. I'm not terribly happy with needing to expose the locks either. Are you thinking something like a method that returns `ReadOnlyNetworkGraph` which wraps both `RwLockReadGuard`s? It's API could be simple look-ups / counts, as I think that is all that `get_route` needs.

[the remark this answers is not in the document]
TheBlueMatt: > Are you thinking something like a method that returns ReadOnlyNetworkGraph which wraps both RwLockReadGuards? It's API could be simple look-ups / counts, as I think that is all that get_route needs.

Yes, that is what I'm suggesting.

lightning/src/routing/network_graph.rs
```diff
@@ -57,6 +57,52 @@ pub struct NetworkGraph {
...
+	NodeFailure {
+		/// The node id of the failed node.
+		node_id: PublicKey,
+		/// Whether the node should be permanently removed from consideration or can be restored
+		/// when a new `channel_update` message is received.
+		is_permanent: bool,
+	}
+}
+
+impl_writeable_tlv_based_enum!(NetworkUpdate,
```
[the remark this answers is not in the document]
TheBlueMatt: These are now upstream.

lightning/src/routing/network_graph.rs
TheBlueMatt: Hmm, yea, good point. This isn't a trait change, though, so we could do different things for different handlers, right?

jkczyz: Not quite sure I follow what you mean by do "different things for different handlers". Do you have a concrete example that may clarify?

TheBlueMatt: As in, there's nothing here that requires that we handle ignoring/passing down events the same for NetworkGraph as we do for PaymentRetryer, or whatever, is there?

TheBlueMatt: Aside from, of course, the fact that EventHandler takes the event by move, not by reference, but we could change that with minimal overhead, I think.

jkczyz: Correct. Though you're right that passing by move is not as flexible since it doesn't allow passing to multiple event handlers in interesting ways. Or post-handling an event after passing to an inner handler for that matter. Updated to pass by reference.

lightning-background-processor/src/lib.rs
```diff
@@ -143,6 +145,7 @@ impl BackgroundProcessor {
 			log_trace!(logger, "Calling ChannelManager's timer_tick_occurred on startup");
 			channel_manager.timer_tick_occurred();
 
+			let event_handler = NetworkUpdateHandler::new(network_graph, &*logger, event_handler);
```
valentinewallace, line 148: Could we update the [top-level BackgroundProcessor docs](https://github.com/rust-bitcoin/rust-lightning/blob/main/lightning-background-processor/src/lib.rs#L28) for this new behavior?

lightning/src/util/events.rs
```diff
@@ -128,6 +129,12 @@ pub enum Event {
 		/// the payment has failed, not just the route in question. If this is not set, you may
 		/// retry the payment via a different route.
 		rejected_by_dest: bool,
+		/// Any failure information conveyed via the Onion return packet by a node along the failed
+		/// payment route. Should be applied to the [`NetworkGraph`] so that routing decisions can
+		/// take into account the update.
+		///
+		/// [`NetworkGraph`]: crate::routing::network_graph::NetworkGraph
+		network_update: Option<NetworkUpdate>,
```
valentinewallace, line 137: Would be good to describe how to apply the event to the NetworkGraph and that BackgroundProcessor will do this for you

valentinewallace (commented):
General call for more docs on the usage of this new API, but this is looking good to me!

valentinewallace:
Implemented in the sample: https://github.com/valentinewallace/ldk-sample/commit/ddc94145bd2e9b4ebedbff4722783fa21478f362
I think it looks good, though it does add a `clone` to each Event and a few more clones of NetworkGraph.

TheBlueMatt:
> Implemented in the sample: valentinewallace/ldk-sample@ddc9414 I think it looks good, though it does add a clone to each Event and a few more clones of NetworkGraph.

Note that you don't actually need the clone - https://git.bitcoin.ninja/index.cgi?p=ldk-sample;a=commit;h=bc119d8fb23d8e55c91df002ca490062f9cc72a6

jkczyz:
Updated the docs!

There's a couple open issues with testing:
- `test_invalid_channel_announcement` seems non-sensical, so maybe just drop it as discussed on Slack
- `htlc_fail_async_shutdown` fails when performing checks in `expect_payment_failed_with_update`

For the latter, I commented out the assertion in `expect_payment_failed_with_update`, which I won't want to merge. So need to figure a workaround. Perhaps just not use the macro. 😕

lightning/src/ln/functional_tests.rs
```diff
@@ -1007,23 +1007,17 @@ fn htlc_fail_async_shutdown() {
 	nodes[0].node.handle_update_fail_htlc(&nodes[1].node.get_our_node_id(), &updates_2.update_fail_htlcs[0]);
 	commitment_signed_dance!(nodes[0], nodes[1], updates_2.commitment_signed, false, true);
 
-	expect_payment_failed!(nodes[0], our_payment_hash, false);
+	expect_payment_failed_with_update!(nodes[0], our_payment_hash, false, chan_1.0.contents.short_channel_id, false);
```
TheBlueMatt, line 1010: I'd argue the issue here is on the channel end - we fail with chan_disabled instead of permanent_channel_failure. I think we should swap the shutdown checks in `Channel::update_add_htlc` to instead use `0x4000|8` and then fix channelmanager to handle it (I haven't reviewed this, just hacked it up, so may need more thought):

```
@@ -3461,33 +3464,32 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
                                }
 
                                let create_pending_htlc_status = |chan: &Channel<Signer>, pending_forward_info: PendingHTLCStatus, error_code: u16| {
-                                       // Ensure error_code has the UPDATE flag set, since by default we send a
-                                       // channel update along as part of failing the HTLC.
-                                       assert!((error_code & 0x1000) != 0);
                                        // If the update_add is completely bogus, the call will Err and we will close,
                                        // but if we've sent a shutdown and they haven't acknowledged it yet, we just
                                        // want to reject the new HTLC and fail it backwards instead of forwarding.
                                        match pending_forward_info {
                                                PendingHTLCStatus::Forward(PendingHTLCInfo { ref incoming_shared_secret, .. }) => {
-                                                       let reason = if let Ok(upd) = self.get_channel_update_for_unicast(chan) {
-                                                               onion_utils::build_first_hop_failure_packet(incoming_shared_secret, error_code, &{
-                                                                       let mut res = Vec::with_capacity(8 + 128);
-                                                                       // TODO: underspecified, follow https://github.com/lightningnetwork/lightning-rfc/issues/791
-                                                                       res.extend_from_slice(&byte_utils::be16_to_array(0));
-                                                                       res.extend_from_slice(&upd.encode_with_len()[..]);
-                                                                       res
-                                                               }[..])
+                                                       let reason = if (error_code & 0x1000) != 0 {
+                                                               if let Ok(upd) = self.get_channel_update_for_unicast(chan) {
+                                                                       onion_utils::build_first_hop_failure_packet(incoming_shared_secret, error_code, &{
+                                                                               let mut res = Vec::with_capacity(8 + 128);
+                                                                               // TODO: underspecified, follow https://github.com/lightningnetwork/lightning-rfc/issues/791
+                                                                               res.extend_from_slice(&byte_utils::be16_to_array(0));
+                                                                               res.extend_from_slice(&upd.encode_with_len()[..]);
+                                                                               res
+                                                                       }[..])
+                                                               } else {
+                                                                       // The only case where we'd be unable to successfully get a
+                                                                       // channel update is if the channel isn't in the fully-funded
+                                                                       // state yet, implying our counterparty is trying to route
+                                                                       // payments over the channel back to themselves (cause no one
+                                                                       // else should know the short_id is a lightning channel yet).
+                                                                       // We should have no problem just calling this
+                                                                       // unknown_next_peer (0x4000|10).
+                                                                       onion_utils::build_first_hop_failure_packet(incoming_shared_secret, 0x4000|10, &[])
+                                                               }
                                                        } else {
-                                                               // The only case where we'd be unable to
-                                                               // successfully get a channel update is if the
-                                                               // channel isn't in the fully-funded state yet,
-                                                               // implying our counterparty is trying to route
-                                                               // payments over the channel back to themselves
-                                                               // (cause no one else should know the short_id
-                                                               // is a lightning channel yet). We should have
-                                                               // no problem just calling this
-                                                               // unknown_next_peer (0x4000|10).
-                                                               onion_utils::build_first_hop_failure_packet(incoming_shared_secret, 0x4000|10, &[])
+                                                               onion_utils::build_first_hop_failure_packet(incoming_shared_secret, error_code, &[])
                                                        };
                                                        let msg = msgs::UpdateFailHTLC {
                                                                channel_id: msg.channel_id,
```

jkczyz: Seems to work though I had to change which channel is used in the failing test otherwise the short channel id wouldn't match. Take a look at d5ea8e8ed7d1d97f2f1abf8cf284eb4ccf1b09ba. I'll properly backport the fixup if all looks well.

lightning/src/routing/network_graph.rs
jkczyz: Haven't pushed it yet, but I went with implementing it on `NetGraphMsgHandler`. It's also nice to not have to know about a decorated `EventHandler` when deserializing a `NetworkGraph`.

That said, just want to point out a drawback with either is that the [guide](https://lightningdevkit.org/docs/build_node_rust/) for setting up a node will need to change. Now event handling (step 16) can no longer be discussed independently of initializing `NetGraphMsgHandler` (optional step 12).

cc: @valentinewallace Any opinions on this?

valentinewallace: Thanks for pointing that out. That is a slight drawback but not a deal breaker IMO

TheBlueMatt: I think that's ok. Indeed, ideally we'd have everything be nice and separate, but if we're gonna have something related to the network graph handle events, that's the outcome we're gonna have. It is nice that all the bindings have moved towards using the lightning-background-processor crate, though, so it'll all be common infrastructure in any case.

jkczyz:
I removed the commit that made `NetworkGraph` have shared ownership since it is no longer necessary now that `NetGraphMsgHandler` is responsible for event handling.

lightning-background-processor/src/lib.rs
```diff
@@ -132,8 +147,9 @@ impl BackgroundProcessor {
 		CM: 'static + Deref<Target = ChannelManager<Signer, CW, T, K, F, L>> + Send + Sync,
 		UMH: 'static + Deref + Send + Sync,
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
+		G: 'static + Deref<Target = NetworkGraph> + Send + Sync,
 	>
-	(persister: CMP, event_handler: EH, chain_monitor: M, channel_manager: CM, peer_manager: PM, logger: L) -> Self
+	(persister: CMP, event_handler: EH, chain_monitor: M, channel_manager: CM, peer_manager: PM, network_graph: G, logger: L) -> Self
```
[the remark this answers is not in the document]
jkczyz: I removed the changes to `BackgroundProcessor`.

valentinewallace (commented):
I think the docs could be a little more explicit that you're expected to nest EventHandlers for `BackgroundProcessor::start` now but this is looking slick overall!

TheBlueMatt:
Needs rebase.

lightning/src/util/test_utils.rs
```diff
@@ -691,3 +691,9 @@ impl core::fmt::Debug for OnRegisterOutput {
 			.finish()
 	}
 }
+
+pub struct NullEventHandler;
```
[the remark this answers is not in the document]
jkczyz: Changed to `TestEvenHandler` to be consistent with other test-only code.

lightning/src/routing/network_graph.rs
```diff
@@ -110,6 +110,66 @@ impl_writeable_tlv_based_enum_upgradable!(NetworkUpdate,
...
+
+	/// Applies changes to the [`NetworkGraph`] from the given update.
+	fn handle_network_update(&self, update: &NetworkUpdate) {
+		match *update {
+			NetworkUpdate::ChannelUpdateMessage { ref msg } => {
+				let short_channel_id = msg.contents.short_channel_id;
+				let is_enabled = msg.contents.flags & (1 << 1) != (1 << 1);
+				let status = if is_enabled { "enabled" } else { "disabled" };
+				log_debug!(self.logger, "Updating channel with channel_update from a payment failure. Channel {} is {}.", short_channel_id, status);
+				let _ = self.network_graph.update_channel(msg, &self.secp_ctx);
```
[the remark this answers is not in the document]
jkczyz: Not sure. The code that produces them is reused in other places but likely isn't actionable here. Might be able to log one marked `IgnoreAndLog`.

jkczyz:
> I think the docs could be a little more explicit that you're expected to nest EventHandlers for `BackgroundProcessor::start` now but this is looking slick overall!

Added a short note about decorators in `start`. Otherwise, we'll likely want some better documentation which could go on a utility used to wrap an event handler with common ones.

TheBlueMatt:
Moving this (and related stuff) to 0.0.102 because it looks like we may want to ship an 0.0.101 to get some fixes out.

jkczyz:
Based on an offline conversation with @TheBlueMatt, I removed the decorated `EventHandler` from `NetGraphMsgHandler`. Instead, `BackgroundProcessor` now takes an optional `NetGraphMsgHandler` which it delegates to before letting the user's event handler run.

This approach is a bit cleaner since it simplifies `NetGraphMsgHandler` and makes it such that `BackgroundProcessor` does the right thing for the user.

lightning/src/ln/functional_test_utils.rs
```diff
@@ -1036,22 +1036,28 @@ macro_rules! expect_payment_forwarded {
...
+			Event::PaymentFailed { ref payment_hash, rejected_by_dest, ref network_update, ref error_code, ref error_data } => {
+				assert_eq!(*payment_hash, $expected_payment_hash, "unexpected payment_hash");
+				assert_eq!(rejected_by_dest, $rejected_by_dest, "unexpected rejected_by_dest value");
+				assert!(error_code.is_some(), "expected error_code.is_some() = true");
+				assert!(error_data.is_some(), "expected error_data.is_some() = true");
+				match network_update {
+					&Some(NetworkUpdate::ChannelUpdateMessage { ref msg }) if !$chan_closed => {
 						assert_eq!(msg.contents.short_channel_id, $scid);
-						assert_eq!(msg.contents.flags & 2, 0);
+						// TODO: Fails for htlc_fail_async_shutdown
```
TheBlueMatt, line 1052: Somehow this fixup got dropped?

TheBlueMatt:
ACK aside from the two comments above.

lightning/src/ln/functional_test_utils.rs
jkczyz: Good catch! I think I may have forgotten to uncomment it after re-ordering the fix.

lightning-background-processor/src/lib.rs
```diff
@@ -26,20 +27,27 @@ use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 use std::ops::Deref;
 
-/// BackgroundProcessor takes care of tasks that (1) need to happen periodically to keep
+/// `BackgroundProcessor` takes care of tasks that (1) need to happen periodically to keep
 /// Rust-Lightning running properly, and (2) either can or should be run in the background. Its
 /// responsibilities are:
-/// * Monitoring whether the ChannelManager needs to be re-persisted to disk, and if so,
+/// * Processing [`MessageSendEvent`]s by calling [`PeerManager::process_events`].
```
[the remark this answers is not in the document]
jkczyz: Moved this out of the list of responsibilities and added the caveat that you mentioned.

lightning-background-processor/src/lib.rs
```diff
@@ -26,20 +27,28 @@ use std::thread::JoinHandle;
...
+///   [`ChannelManager`] persistence should be done in the background.
+/// * Calling [`ChannelManager::timer_tick_occurred`] and [`PeerManager::timer_tick_occurred`]
+///   at the appropriate intervals.
 ///
-/// Note that if ChannelManager persistence fails and the persisted manager becomes out-of-date,
-/// then there is a risk of channels force-closing on startup when the manager realizes it's
-/// outdated. However, as long as `ChannelMonitor` backups are sound, no funds besides those used
-/// for unilateral chain closure fees are at risk.
+/// It will also call [`PeerManager::process_events`] periodically though this shouldn't be relied
+/// upon as doing so may result in high latency.
```
valentinewallace, line 41: Hmm, so I checked lightningdevkit.org guides, the sample, [`process_events` docs](https://docs.rs/lightning/0.0.100/lightning/ln/peer_handler/struct.PeerManager.html#method.process_events) and I don't think we indicate anywhere that the user should be calling this function perioically (i.e., users do currently rely on `BackgroundProcessor` to call `process_events` atm). 

Should we update some other docs/the sample, or open an issue to do so?

TheBlueMatt (approved):

valentinewallace (approved):
Nice, LGTM

lightning-background-processor/src/lib.rs
TheBlueMatt: Its primarily called by lightning-net-tokio or by the NioPeerHandler in Java, so I don't think users need to care, at least not normal users.

TheBlueMatt: I did open an issue at #1075.

PR #1053, opened 2021-08-20
Deduplicate PaymentSent events for MPP payments

valentinewallace:
TODOs

~- [x] dedup `PaymentFailed` events~
- [x] add serialization for pending_outbound_htlc's MPP IDs
~- [x] test that `PaymentFailed` events are deduplicated~

Also, track whether all paths have failed in `PaymentFailed`

### DIFF
--- a/lightning/src/ln/channelmanager.rs
+++ b/lightning/src/ln/channelmanager.rs
@@ -491,8 +491,11 @@ pub struct ChannelManager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, 
 	/// which may generate a claim event, we may receive similar duplicate claim/fail MonitorEvents
 	/// after reloading from disk while replaying blocks against ChannelMonitors.
 	///
+	/// Each payment has each of its MPP part's session_priv bytes in the HashSet of the map (even
+	/// payments over a single path).
+	///
 	/// Locked *after* channel_state.
-	pending_outbound_payments: Mutex<HashSet<[u8; 32]>>,
+	pending_outbound_payments: Mutex<HashMap<MppId, HashSet<[u8; 32]>>>,
 
 	our_network_key: SecretKey,
 	our_network_pubkey: PublicKey,
@@ -1156,7 +1159,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 				pending_msg_events: Vec::new(),
 			}),
 			pending_inbound_payments: Mutex::new(HashMap::new()),
-			pending_outbound_payments: Mutex::new(HashSet::new()),
+			pending_outbound_payments: Mutex::new(HashMap::new()),
 
 			our_network_key: keys_manager.get_node_secret(),
 			our_network_pubkey: PublicKey::from_secret_key(&secp_ctx, &keys_manager.get_node_secret()),
@@ -1853,7 +1856,9 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		let onion_packet = onion_utils::construct_onion_packet(onion_payloads, onion_keys, prng_seed, payment_hash);
 
 		let _persistence_guard = PersistenceNotifierGuard::notify_on_drop(&self.total_consistency_lock, &self.persistence_notifier);
-		assert!(self.pending_outbound_payments.lock().unwrap().insert(session_priv_bytes));
+		let mut pending_outbounds = self.pending_outbound_payments.lock().unwrap();
+		let sessions = pending_outbounds.entry(mpp_id).or_insert(HashSet::new());
+		assert!(sessions.insert(session_priv_bytes));
 
 		let err: Result<(), _> = loop {
 			let mut channel_lock = self.channel_state.lock().unwrap();
@@ -2832,23 +2837,27 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 					self.fail_htlc_backwards_internal(channel_state,
 						htlc_src, &payment_hash, HTLCFailReason::Reason { failure_code, data: onion_failure_data});
 				},
-				HTLCSource::OutboundRoute { session_priv, .. } => {
-					if {
-						let mut session_priv_bytes = [0; 32];
-						session_priv_bytes.copy_from_slice(&session_priv[..]);
-						self.pending_outbound_payments.lock().unwrap().remove(&session_priv_bytes)
-					} {
-						self.pending_events.lock().unwrap().push(
-							events::Event::PaymentFailed {
-								payment_hash,
-								rejected_by_dest: false,
-								network_update: None,
-#[cfg(test)]
-								error_code: None,
-#[cfg(test)]
-								error_data: None,
+				HTLCSource::OutboundRoute { session_priv, mpp_id, .. } => {
+					let mut session_priv_bytes = [0; 32];
+					session_priv_bytes.copy_from_slice(&session_priv[..]);
+					let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+					if let hash_map::Entry::Occupied(mut sessions) = outbounds.entry(mpp_id) {
+						if sessions.get_mut().remove(&session_priv_bytes) {
+							self.pending_events.lock().unwrap().push(
+								events::Event::PaymentFailed {
+									payment_hash,
+									rejected_by_dest: false,
+									network_update: None,
+									#[cfg(test)]
+									error_code: None,
+									#[cfg(test)]
+									error_data: None,
+								}
+							);
+							if sessions.get().len() == 0 {
+								sessions.remove();
 							}
-						)
+						}
 					} else {
 						log_trace!(self.logger, "Received duplicative fail for HTLC with payment_hash {}", log_bytes!(payment_hash.0));
 					}
@@ -2873,12 +2882,19 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		// from block_connected which may run during initialization prior to the chain_monitor
 		// being fully configured. See the docs for `ChannelManagerReadArgs` for more.
 		match source {
-			HTLCSource::OutboundRoute { ref path, session_priv, .. } => {
-				if {
-					let mut session_priv_bytes = [0; 32];
-					session_priv_bytes.copy_from_slice(&session_priv[..]);
-					!self.pending_outbound_payments.lock().unwrap().remove(&session_priv_bytes)
-				} {
+			HTLCSource::OutboundRoute { ref path, session_priv, mpp_id, .. } => {
+				let mut session_priv_bytes = [0; 32];
+				session_priv_bytes.copy_from_slice(&session_priv[..]);
+				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+				if let hash_map::Entry::Occupied(mut sessions) = outbounds.entry(mpp_id) {
+					if !sessions.get_mut().remove(&session_priv_bytes) {
+						log_trace!(self.logger, "Received duplicative fail for HTLC with payment_hash {}", log_bytes!(payment_hash.0));
+						return;
+					}
+					if sessions.get().len() == 0 {
+						sessions.remove();
+					}
+				} else {
 					log_trace!(self.logger, "Received duplicative fail for HTLC with payment_hash {}", log_bytes!(payment_hash.0));
 					return;
 				}
@@ -3119,17 +3135,22 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 
 	fn claim_funds_internal(&self, mut channel_state_lock: MutexGuard<ChannelHolder<Signer>>, source: HTLCSource, payment_preimage: PaymentPreimage, forwarded_htlc_value_msat: Option<u64>, from_onchain: bool) {
 		match source {
-			HTLCSource::OutboundRoute { session_priv, .. } => {
+			HTLCSource::OutboundRoute { session_priv, mpp_id, .. } => {
 				mem::drop(channel_state_lock);
-				if {
-					let mut session_priv_bytes = [0; 32];
-					session_priv_bytes.copy_from_slice(&session_priv[..]);
-					self.pending_outbound_payments.lock().unwrap().remove(&session_priv_bytes)
-				} {
-					let mut pending_events = self.pending_events.lock().unwrap();
-					pending_events.push(events::Event::PaymentSent {
-						payment_preimage
-					});
+				let mut session_priv_bytes = [0; 32];
+				session_priv_bytes.copy_from_slice(&session_priv[..]);
+				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+				if let Some(sessions) = outbounds.get_mut(&mpp_id) {
+					if sessions.remove(&session_priv_bytes) {
+						self.pending_events.lock().unwrap().push(
+							events::Event::PaymentSent { payment_preimage }
+						);
+						if sessions.len() == 0 {
+							outbounds.remove(&mpp_id);
+						}
+					} else {
+						log_trace!(self.logger, "Received duplicative fulfill for HTLC with payment_preimage {}", log_bytes!(payment_preimage.0));
+					}
 				} else {
 					log_trace!(self.logger, "Received duplicative fulfill for HTLC with payment_preimage {}", log_bytes!(payment_preimage.0));
 				}
@@ -5105,12 +5126,21 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> Writeable f
 		}
 
 		let pending_outbound_payments = self.pending_outbound_payments.lock().unwrap();
-		(pending_outbound_payments.len() as u64).write(writer)?;
-		for session_priv in pending_outbound_payments.iter() {
-			session_priv.write(writer)?;
+		// For backwards compat, write the session privs and their total length.
+		let mut num_pending_outbounds_compat: u64 = 0;
+		for (_, outbounds) in pending_outbound_payments.iter() {
+			num_pending_outbounds_compat += outbounds.len() as u64;
+		}
+		num_pending_outbounds_compat.write(writer)?;
+		for (_, outbounds) in pending_outbound_payments.iter() {
+			for outbound in outbounds.iter() {
+				outbound.write(writer)?;
+			}
 		}
 
-		write_tlv_fields!(writer, {});
+		write_tlv_fields!(writer, {
+			(1, pending_outbound_payments, required),
+		});
 
 		Ok(())
 	}
@@ -5363,15 +5393,23 @@ impl<'a, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
 			}
 		}
 
-		let pending_outbound_payments_count: u64 = Readable::read(reader)?;
-		let mut pending_outbound_payments: HashSet<[u8; 32]> = HashSet::with_capacity(cmp::min(pending_outbound_payments_count as usize, MAX_ALLOC_SIZE/32));
-		for _ in 0..pending_outbound_payments_count {
-			if !pending_outbound_payments.insert(Readable::read(reader)?) {
-				return Err(DecodeError::InvalidValue);
-			}
+		let pending_outbound_payments_count_compat: u64 = Readable::read(reader)?;
+		let mut pending_outbound_payments_compat: HashMap<MppId, HashSet<[u8; 32]>> =
+			HashMap::with_capacity(cmp::min(pending_outbound_payments_count_compat as usize, MAX_ALLOC_SIZE/32));
+		for _ in 0..pending_outbound_payments_count_compat {
+			let session_priv = Readable::read(reader)?;
+			if pending_outbound_payments_compat.insert(MppId(session_priv), [session_priv].iter().cloned().collect()).is_some() {
+				return Err(DecodeError::InvalidValue)
+			};
 		}
 
-		read_tlv_fields!(reader, {});
+		let mut pending_outbound_payments = None;
+		read_tlv_fields!(reader, {
+			(1, pending_outbound_payments, option),
+		});
+		if pending_outbound_payments.is_none() {
+			pending_outbound_payments = Some(pending_outbound_payments_compat);
+		}
 
 		let mut secp_ctx = Secp256k1::new();
 		secp_ctx.seeded_randomize(&args.keys_manager.get_secure_random_bytes());
@@ -5392,7 +5430,7 @@ impl<'a, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
 				pending_msg_events: Vec::new(),
 			}),
 			pending_inbound_payments: Mutex::new(pending_inbound_payments),
-			pending_outbound_payments: Mutex::new(pending_outbound_payments),
+			pending_outbound_payments: Mutex::new(pending_outbound_payments.unwrap()),
 
 			our_network_key: args.keys_manager.get_node_secret(),
 			our_network_pubkey: PublicKey::from_secret_key(&secp_ctx, &args.keys_manager.get_node_secret()),
--- a/lightning/src/ln/channelmanager.rs
+++ b/lightning/src/ln/channelmanager.rs
@@ -173,6 +173,22 @@ struct ClaimableHTLC {
 	onion_payload: OnionPayload,
 }
 
+/// A payment identifier used to correlate an MPP payment's per-path HTLC sources internally.
+#[derive(Hash, Copy, Clone, PartialEq, Eq, Debug)]
+pub(crate) struct MppId(pub [u8; 32]);
+
+impl Writeable for MppId {
+	fn write<W: Writer>(&self, w: &mut W) -> Result<(), io::Error> {
+		self.0.write(w)
+	}
+}
+
+impl Readable for MppId {
+	fn read<R: Read>(r: &mut R) -> Result<Self, DecodeError> {
+		let buf: [u8; 32] = Readable::read(r)?;
+		Ok(MppId(buf))
+	}
+}
 /// Tracks the inbound corresponding to an outbound HTLC
 #[derive(Clone, PartialEq)]
 pub(crate) enum HTLCSource {
@@ -183,6 +199,7 @@ pub(crate) enum HTLCSource {
 		/// Technically we can recalculate this from the route, but we cache it here to avoid
 		/// doing a double-pass on route when we get a failure back
 		first_hop_htlc_msat: u64,
+		mpp_id: MppId,
 	},
 }
 #[cfg(test)]
@@ -192,6 +209,7 @@ impl HTLCSource {
 			path: Vec::new(),
 			session_priv: SecretKey::from_slice(&[1; 32]).unwrap(),
 			first_hop_htlc_msat: 0,
+			mpp_id: MppId([2; 32]),
 		}
 	}
 }
@@ -1820,7 +1838,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 	}
 
 	// Only public for testing, this should otherwise never be called direcly
-	pub(crate) fn send_payment_along_path(&self, path: &Vec<RouteHop>, payment_hash: &PaymentHash, payment_secret: &Option<PaymentSecret>, total_value: u64, cur_height: u32, keysend_preimage: &Option<PaymentPreimage>) -> Result<(), APIError> {
+	pub(crate) fn send_payment_along_path(&self, path: &Vec<RouteHop>, payment_hash: &PaymentHash, payment_secret: &Option<PaymentSecret>, total_value: u64, cur_height: u32, mpp_id: MppId, keysend_preimage: &Option<PaymentPreimage>) -> Result<(), APIError> {
 		log_trace!(self.logger, "Attempting to send payment for path with next hop {}", path.first().unwrap().short_channel_id);
 		let prng_seed = self.keys_manager.get_secure_random_bytes();
 		let session_priv_bytes = self.keys_manager.get_secure_random_bytes();
@@ -1857,6 +1875,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 						path: path.clone(),
 						session_priv: session_priv.clone(),
 						first_hop_htlc_msat: htlc_msat,
+						mpp_id,
 					}, onion_packet, &self.logger), channel_state, chan)
 				} {
 					Some((update_add, commitment_signed, monitor_update)) => {
@@ -1956,6 +1975,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		let mut total_value = 0;
 		let our_node_id = self.get_our_node_id();
 		let mut path_errs = Vec::with_capacity(route.paths.len());
+		let mpp_id = MppId(self.keys_manager.get_secure_random_bytes());
 		'path_check: for path in route.paths.iter() {
 			if path.len() < 1 || path.len() > 20 {
 				path_errs.push(Err(APIError::RouteError{err: "Path didn't go anywhere/had bogus size"}));
@@ -1977,7 +1997,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		let cur_height = self.best_block.read().unwrap().height() + 1;
 		let mut results = Vec::new();
 		for path in route.paths.iter() {
-			results.push(self.send_payment_along_path(&path, &payment_hash, payment_secret, total_value, cur_height, &keysend_preimage));
+			results.push(self.send_payment_along_path(&path, &payment_hash, payment_secret, total_value, cur_height, mpp_id, &keysend_preimage));
 		}
 		let mut has_ok = false;
 		let mut has_err = false;
@@ -4911,14 +4931,60 @@ impl Readable for ClaimableHTLC {
 	}
 }
 
-impl_writeable_tlv_based_enum!(HTLCSource,
-	(0, OutboundRoute) => {
-		(0, session_priv, required),
-		(2, first_hop_htlc_msat, required),
-		(4, path, vec_type),
-	}, ;
-	(1, PreviousHopData)
-);
+impl Readable for HTLCSource {
+	fn read<R: Read>(reader: &mut R) -> Result<Self, DecodeError> {
+		let id: u8 = Readable::read(reader)?;
+		match id {
+			0 => {
+				let mut session_priv: ::util::ser::OptionDeserWrapper<SecretKey> = ::util::ser::OptionDeserWrapper(None);
+				let mut first_hop_htlc_msat: u64 = 0;
+				let mut path = Some(Vec::new());
+				let mut mpp_id = None;
+				read_tlv_fields!(reader, {
+					(0, session_priv, required),
+					(1, mpp_id, option),
+					(2, first_hop_htlc_msat, required),
+					(4, path, vec_type),
+				});
+				if mpp_id.is_none() {
+					// For backwards compat, if there was no mpp_id written, use the session_priv bytes
+					// instead.
+					mpp_id = Some(MppId(*session_priv.0.unwrap().as_ref()));
+				}
+				Ok(HTLCSource::OutboundRoute {
+					session_priv: session_priv.0.unwrap(),
+					first_hop_htlc_msat: first_hop_htlc_msat,
+					path: path.unwrap(),
+					mpp_id: mpp_id.unwrap(),
+				})
+			}
+			1 => Ok(HTLCSource::PreviousHopData(Readable::read(reader)?)),
+			_ => Err(DecodeError::UnknownRequiredFeature),
+		}
+	}
+}
+
+impl Writeable for HTLCSource {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ::io::Error> {
+		match self {
+			HTLCSource::OutboundRoute { ref session_priv, ref first_hop_htlc_msat, ref path, mpp_id } => {
+				0u8.write(writer)?;
+				let mpp_id_opt = Some(mpp_id);
+				write_tlv_fields!(writer, {
+					(0, session_priv, required),
+					(1, mpp_id_opt, option),
+					(2, first_hop_htlc_msat, required),
+					(4, path, vec_type),
+				 });
+			}
+			HTLCSource::PreviousHopData(ref field) => {
+				1u8.write(writer)?;
+				field.write(writer)?;
+			}
+		}
+		Ok(())
+	}
+}
 
 impl_writeable_tlv_based_enum!(HTLCFailReason,
 	(0, LightningError) => {
@@ -5364,7 +5430,7 @@ mod tests {
 	use bitcoin::hashes::sha256::Hash as Sha256;
 	use core::time::Duration;
 	use ln::{PaymentPreimage, PaymentHash, PaymentSecret};
-	use ln::channelmanager::PaymentSendFailure;
+	use ln::channelmanager::{MppId, PaymentSendFailure};
 	use ln::features::{InitFeatures, InvoiceFeatures};
 	use ln::functional_test_utils::*;
 	use ln::msgs;
@@ -5515,10 +5581,11 @@ mod tests {
 		let net_graph_msg_handler = &nodes[0].net_graph_msg_handler;
 		let route = get_route(&nodes[0].node.get_our_node_id(), &net_graph_msg_handler.network_graph, &nodes[1].node.get_our_node_id(), Some(InvoiceFeatures::known()), None, &Vec::new(), 100_000, TEST_FINAL_CLTV, &logger).unwrap();
 		let (payment_preimage, our_payment_hash, payment_secret) = get_payment_preimage_hash!(&nodes[1]);
+		let mpp_id = MppId([42; 32]);
 		// Use the utility function send_payment_along_path to send the payment with MPP data which
 		// indicates there are more HTLCs coming.
 		let cur_height = CHAN_CONFIRM_DEPTH + 1; // route_payment calls send_payment, which adds 1 to the current height. So we do the same here to match.
-		nodes[0].node.send_payment_along_path(&route.paths[0], &our_payment_hash, &Some(payment_secret), 200_000, cur_height, &None).unwrap();
+		nodes[0].node.send_payment_along_path(&route.paths[0], &our_payment_hash, &Some(payment_secret), 200_000, cur_height, mpp_id, &None).unwrap();
 		check_added_monitors!(nodes[0], 1);
 		let mut events = nodes[0].node.get_and_clear_pending_msg_events();
 		assert_eq!(events.len(), 1);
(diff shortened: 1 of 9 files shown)


lightning/src/ln/channelmanager.rs
```diff
@@ -2812,11 +2835,11 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 					self.fail_htlc_backwards_internal(channel_state,
 						htlc_src, &payment_hash, HTLCFailReason::Reason { failure_code, data: onion_failure_data});
 				},
-				HTLCSource::OutboundRoute { session_priv, .. } => {
+				HTLCSource::OutboundRoute { session_priv, mpp_id, .. } => {
 					if {
 						let mut session_priv_bytes = [0; 32];
 						session_priv_bytes.copy_from_slice(&session_priv[..]);
-						self.pending_outbound_payments.lock().unwrap().remove(&session_priv_bytes)
+						self.pending_outbound_payments.lock().unwrap().remove(&(session_priv_bytes, mpp_id))
```
TheBlueMatt, line 2842: So it looks like we could still get a `PaymentFailed` followed by a `PaymentSent` event, or possibly multiple `PaymentFailed` events here? Its somewhat unclear exactly what the solution should be, though - ideally we'd like to wait and only call it a "PaymentFailed" failure if all MPP fragments fail, but at the same time users need to be notified if a fragment fails so that they can retry that one path. Maybe we need a new `MPPFragmentFailed` event and then only generate `PaymentFailed` if *all* fragments fail (ie there are no more with the same mpp id)?

lightning/src/ln/channelmanager.rs
```diff
@@ -474,7 +493,7 @@ pub struct ChannelManager<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref,
 	/// after reloading from disk while replaying blocks against ChannelMonitors.
 	///
 	/// Locked *after* channel_state.
-	pending_outbound_payments: Mutex<HashSet<[u8; 32]>>,
+	pending_outbound_payments: Mutex<HashSet<([u8; 32], Option<MppId>)>>,
```
TheBlueMatt, line 496: I have to admit this datastructure "feels" wrong - you have to walk the full thing to remove all payments with a given mpp id, which is maybe ok, but if we also want to change the event types based on "are there pending htlcs with the same mpp id" we probably need it to be a map from MppId to the per-payment info. Of course that also sucks cause then we either have a multi-layer `HashMap` or a ton of entries for the no-MppId key. Another option would be a something b-tree based, which would work here, but Rust's BTreeSet doesn't support `drain(range)`, so doesn't quite have what we need.

Maybe we should do a `HashMap<MppId, Option<HashSet<[u8; 32]>>>` - always set an MppId even for non-MPP payments, and only fill in the second set if its MPP. I hate multi-level maps like that, but I'm not sure any alternative is better, and walking the map sucks too.

lightning/src/ln/channelmanager.rs
```diff
@@ -2852,14 +2891,23 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
-					!self.pending_outbound_payments.lock().unwrap().remove(&session_priv_bytes)
-				} {
+			HTLCSource::OutboundRoute { ref path, session_priv, mpp_id, .. } => {
+				let mut session_priv_bytes = [0; 32];
+				session_priv_bytes.copy_from_slice(&session_priv[..]);
+				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+				let mut partial_failure = true;
+				if let Some(sessions) = outbounds.get_mut(&mpp_id) {
+					if !sessions.remove(&session_priv_bytes) {
+							log_trace!(self.logger, "Received duplicative fail for HTLC with payment_hash {}", log_bytes!(payment_hash.0));
```
TheBlueMatt, line 2901: nit: double-indentation here.

lightning/src/ln/channelmanager.rs
```diff
@@ -5301,11 +5420,33 @@ impl<'a, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
...
-		let mut pending_outbound_payments: HashSet<[u8; 32]> = HashSet::with_capacity(cmp::min(pending_outbound_payments_count as usize, MAX_ALLOC_SIZE/32));
-		for _ in 0..pending_outbound_payments_count {
-			if !pending_outbound_payments.insert(Readable::read(reader)?) {
-				return Err(DecodeError::InvalidValue);
+		let pending_outbound_payments_count_compat: u64 = Readable::read(reader)?;
+		let mut pending_outbound_payments: HashMap<MppId, HashSet<[u8; 32]>> =
+			HashMap::with_capacity(cmp::min(pending_outbound_payments_count_compat as usize, MAX_ALLOC_SIZE/32));
+
+		match pending_outbound_payments_count_compat {
+			0 => {
```
TheBlueMatt, line 5428: Old clients can have zero entries in their pending outbound payment sets, in fact are likely to. Instead, I think we should read the old entries, insert them as needed, and read the new entries via a TLV.

lightning/src/ln/channelmanager.rs
```diff
@@ -3103,20 +3169,20 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
-					log_trace!(self.logger, "Received duplicative fulfill for HTLC with payment_preimage {}", log_bytes!(payment_preimage.0));
+				let mut session_priv_bytes = [0; 32];
+				session_priv_bytes.copy_from_slice(&session_priv[..]);
+				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+				if let Some(mut sessions) = outbounds.remove(&mpp_id) {
+					if sessions.remove(&session_priv_bytes) {
+						self.pending_events.lock().unwrap().push(
+							events::Event::PaymentSent { payment_preimage }
+						);
+						return
```
TheBlueMatt, line 3182: Do we not want to drop the whole pending payment at this point? If we get one preimage, we should consider the payment success, irrespective of future HTLC failures.

lightning/src/util/events.rs
```diff
@@ -246,6 +260,13 @@ impl Writeable for Event {
 					(2, claim_from_onchain_tx, required),
 				});
 			},
+			&Event::MPPFragmentFailed { ref payment_hash, ref payment_path } => {
+				8u8.write(writer)?;
```
TheBlueMatt, line 264: I feel like this should be odd. Old clients should be able to ignore `MPPFragmentFailed` events - they won't retry anyway so its not like they aren't broken in *new* ways, they're just broken, and shouldn't *also* fail to even run.

lightning/src/ln/channelmanager.rs
valentinewallace: I believe that's how it's implemented: `outbounds.remove(&mpp_id)` above

lightning/src/ln/channelmanager.rs
valentinewallace: That'd mean adapting the TLV macros to work with hashmaps, right?

TheBlueMatt: That would just mean implementing `Readable`/`Writeable` for hashmaps, I don't think any changes to the macros needs to happen. Just write a backwards-compat read block, then try to read an `optional` HashMap in the TLV read and if it comes back `Some`, overwrite the backwards-compat read.

lightning/src/ln/channelmanager.rs
TheBlueMatt: Ah, duh, yes. Next question, should we not just always generate `PaymentSent` even if we've already removed the session priv? Or, really, isn't that unreachable? I'm not sure how we could hit the first if but not the second.

valentinewallace: Made it unreachable!

lightning/src/ln/channelmanager.rs
```diff
@@ -5042,13 +5156,12 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> Writeable f
 			pending_payment.write(writer)?;
 		}
 
-		let pending_outbound_payments = self.pending_outbound_payments.lock().unwrap();
-		(pending_outbound_payments.len() as u64).write(writer)?;
-		for session_priv in pending_outbound_payments.iter() {
-			session_priv.write(writer)?;
-		}
+		(0 as u64).write(writer)?; // for backwards compat with the old format of pending_outbound_payments
```
TheBlueMatt, line 5159: Hmm, so this is a "backwards incompatibility". I dunno if we're *super* worried about it, but practically this means "old clients will not create *any* events for all payments initiated by new clients". In general, if its easy, I think we'd prefer to avoid things like this. One easy way to do that here would be to write out the full set of session privs but not the mpp ids until the TLV entry.

lightning/src/ln/channelmanager.rs
```diff
@@ -5301,15 +5414,25 @@ impl<'a, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
...
-		for _ in 0..pending_outbound_payments_count {
-			if !pending_outbound_payments.insert(Readable::read(reader)?) {
-				return Err(DecodeError::InvalidValue);
-			}
+		let pending_outbound_payments_count_compat: u64 = Readable::read(reader)?;
+		let mut pending_outbound_payments_compat: HashMap<MppId, HashSet<[u8; 32]>> =
+			HashMap::with_capacity(cmp::min(pending_outbound_payments_count_compat as usize, MAX_ALLOC_SIZE/32));
+		for _ in 0..pending_outbound_payments_count_compat {
+			let session_priv = Readable::read(reader)?;
+			if pending_outbound_payments_compat.insert(MppId(session_priv), [session_priv].iter().cloned().collect()).is_some() {
```
TheBlueMatt, line 5422: can't you just `vec![session_priv]`?

valentinewallace: We want to insert a HashMap, not a Vec iiuc

TheBlueMatt: Oh duh, lol, yes.

lightning/src/ln/channelmanager.rs
```diff
@@ -4915,14 +4981,62 @@ impl Readable for ClaimableHTLC {
...
+				read_tlv_fields!(reader, {
+					(0, session_priv, required),
+					(2, first_hop_htlc_msat, required),
+					(4, path, vec_type),
+					(6, mpp_id, option),
+				});
+				if mpp_id.is_none() {
+					// For backwards compat, if there was no mpp_id written, use the session_priv bytes
+					// instead.
+					mpp_id = Some(MppId(*session_priv.0.unwrap().as_ref()));
```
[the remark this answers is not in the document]
valentinewallace: Seems the session_priv won't have been read yet if I do that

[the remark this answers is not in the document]
TheBlueMatt: `read_tlv_fields` should always overwrite the fields being read, at least if they're present in the TLV stream being read. It should otherwise leave them alone.

lightning/src/ln/channelmanager.rs
valentinewallace: Wait, I think this is reachable if we receive a duplicate `update_fulfill_htlc` msg? So I think the right response is to log that there's been a duplicate, not be `unreachable`

lightning/src/ln/channelmanager.rs
valentinewallace: > One easy way to do that here would be to write out the full set of session privs but not the mpp ids until the TLV entry.

Went with this

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
valentinewallace: Think I'm misunderstanding, is this the diff you're thinking of? Or are you thinking of initializing `mpp_id` with a random number instead of `session_priv`? 
```
index 4d9a965f..5792f4d2 100644
@@ -4991,17 +4991,17 @@ impl Readable for HTLCSource {
                                let mut first_hop_htlc_msat: u64 = 0;
                                let mut path = Some(Vec::new());
                                let mut mpp_id = None;
+                               if mpp_id.is_none() {
+                                       // For backwards compat, if there was no mpp_id written, use the session_priv bytes
+                                       // instead.
+                                       mpp_id = Some(MppId(*session_priv.0.unwrap().as_ref()));
                                read_tlv_fields!(reader, {
                                        (0, session_priv, required),
                                        (1, mpp_id, option),
                                        (2, first_hop_htlc_msat, required),
                                        (4, path, vec_type),
                                });
-                               if mpp_id.is_none() {
-                                       // For backwards compat, if there was no mpp_id written, use the session_priv bytes
-                                       // instead.
-                                       mpp_id = Some(MppId(*session_priv.0.unwrap().as_ref()));
                                Ok(HTLCSource::OutboundRoute {
                                        session_priv: session_priv.0.unwrap(),
                                        first_hop_htlc_msat: first_hop_htlc_msat,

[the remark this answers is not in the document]
Comment (TheBlueMatt) [reply lightning/src/ln/channelmanager.rs:5003]: Ohhhhhhhhhhhhhhhhhhhhhh, yes, you're totally right, that is the diff I was thinking of, and, yes, it is nonsensical.

File: lightning/src/ln/channelmanager.rs (hunk 5 above)
Comment (TheBlueMatt) [reply lightning/src/ln/channelmanager.rs]: Ah, you mean if we sent two payments to the same payment hash and have a duplicate claim on the first? I believe you may be correct.

**Comment by TheBlueMatt** [comment]
Should this get a new test for the new event?

File: lightning/src/ln/channelmanager.rs [hunk 10]
```diff
@@ -2812,23 +2837,37 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
-#[cfg(test)]
-								error_code: None,
-#[cfg(test)]
-								error_data: None,
+				HTLCSource::OutboundRoute { session_priv, mpp_id, path, .. } => {
+					let mut session_priv_bytes = [0; 32];
+					session_priv_bytes.copy_from_slice(&session_priv[..]);
+					let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+					if let Some(sessions) = outbounds.get_mut(&mpp_id) {
+						if sessions.remove(&session_priv_bytes) {
```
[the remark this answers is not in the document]
valentinewallace: > now when we receive a failure on another fragment, we won't generate an event for it?

Confused, I think we'll generate an event for every `session_priv` removed, so that'd be 1 event per fragment iiuc?

valentinewallace:
> Should this get a new test for the new event?

Added `mpp_failure()` test in `functional_tests` which tests the new event

lightning/src/util/events.rs
```diff
@@ -147,6 +148,14 @@ pub enum Event {
 #[cfg(test)]
 		error_data: Option<Vec<u8>>,
 	},
+	/// Indicates an outbound MPP payment partially failed. Probably some intermediary node dropped
+	/// something. You may wish to retry this portion of the payment with a different path.
+	MPPFragmentFailed {
```
jkczyz, line 154: I've seen "part" used frequently. Wondering if we should use similar terminology.

lightning/src/util/events.rs
```diff
@@ -147,6 +148,14 @@ pub enum Event {
...
+		/// The hash which was given to ChannelManager::send_payment.
+		payment_hash: PaymentHash,
+		/// The path that failed.
+		payment_path: Vec<RouteHop>,
```
jkczyz, line 158: I'm working on a PR that adds the path to `PaymentFailed` for scoring purpose. It will also include the index of the failed hop. Given that, do you think we need a separate event? Could we add a marker for multi-path payments if needed?

lightning/src/ln/channelmanager.rs
```diff
@@ -173,6 +173,22 @@ struct ClaimableHTLC {
 	onion_payload: OnionPayload,
 }
 
+/// A payment identifier used to correlate an MPP payment's per-path HTLC sources internally.
+#[derive(Hash, Copy, Clone, PartialEq, Eq, Debug)]
+pub(crate) struct MppId(pub [u8; 32]);
```
jkczyz, line 178: Given this is only used internally, would an integer be sufficient? It could be tuple'd with the payment hash (or `session_priv`?) for lookups. I may not have a firm understanding of how `session_priv` is used, FWIW. So happy to get a better idea of how is used for each part. 🙂

jkczyz, line 178: Maybe `MppPartId`?

lightning/src/ln/channelmanager.rs
```diff
@@ -2873,42 +2921,60 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
-								network_update,
-#[cfg(test)]
-								error_code: onion_error_code,
-#[cfg(test)]
-								error_data: onion_error_data
-							}
-						);
+						if partial_failure {
+							self.pending_events.lock().unwrap().push(
+								events::Event::MPPFragmentFailed {
```
TheBlueMatt, line 2926: This should also contain the `network_update`.

lightning/src/ln/channelmanager.rs
```diff
@@ -3140,18 +3140,18 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
 				session_priv_bytes.copy_from_slice(&session_priv[..]);
 				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
-				if let Some(sessions) = outbounds.get_mut(&mpp_id) {
-					if sessions.remove(&session_priv_bytes) {
-						self.pending_events.lock().unwrap().push(
-							events::Event::PaymentSent { payment_preimage }
-						);
-						if sessions.len() == 0 {
-							outbounds.remove(&mpp_id);
+				let found_payment = loop {
```
jkczyz, line 3173: Is `loop` needed if you always `break`? Maybe just log in the appropriate `else`.

lightning/src/ln/channelmanager.rs
```diff
@@ -2812,23 +2837,37 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
+											payment_hash,
+											rejected_by_dest: false,
+											network_update: None,
+											#[cfg(test)]
+											error_code: None,
+											#[cfg(test)]
+											error_data: None,
+										}
+									);
+									outbounds.remove(&mpp_id);
```
TheBlueMatt, line 2859: This (and other cases later) can be more elegantly handled with the Entry API:

```
@@ -2843,5 +2843,5 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
-                                       if let Some(sessions) = outbounds.get_mut(&mpp_id) {
-                                               if sessions.remove(&session_priv_bytes) {
-                                                       match sessions.len() {
+                                       if let hash_map::Entry::Occupied(mut sessions) = outbounds.entry(mpp_id) {
+                                               if sessions.get_mut().remove(&session_priv_bytes) {
+                                                       match sessions.get().len() {
                                                                0 => {
@@ -2858,3 +2858,3 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
                                                                        );
-                                                                       outbounds.remove(&mpp_id);
+                                                                       sessions.remove();
                                                                },
```

lightning/src/util/events.rs
```diff
@@ -112,8 +112,12 @@ pub enum Event {
 		/// payment is to pay an invoice or to send a spontaneous payment.
 		purpose: PaymentPurpose,
 	},
-	/// Indicates an outbound payment we made succeeded (ie it made it all the way to its target
+	/// Indicates an outbound payment we made succeeded (i.e. it made it all the way to its target
 	/// and we got back the payment preimage for it).
+	///
+	/// Note for MPP payments: we generate this event as soon as one of the payment paths succeeds. In
+	/// the case of a buggy payee, this may be followed by a PaymentFailed or MPPFragmentFailed event
```
jkczyz, line 120: nit: `MPPFragmentFailed` mentioned before added

jkczyz, line 120: nit: tick event names

lightning/src/ln/channelmanager.rs
```diff
@@ -5038,13 +5155,25 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> Writeable f
 			pending_payment.write(writer)?;
 		}
 
-		let pending_outbound_payments = self.pending_outbound_payments.lock().unwrap();
-		(pending_outbound_payments.len() as u64).write(writer)?;
-		for session_priv in pending_outbound_payments.iter() {
-			session_priv.write(writer)?;
+		let pending_outbounds_compat = {
+			let mut pending_session_privs: HashSet<[u8; 32]> = HashSet::new();
```
TheBlueMatt, line 5159: Do we have to create a hashset here? I believe the actual session privs are guaranteed to be unique. We could instead just count them in the loop here instead and then loop twice, no?

TheBlueMatt (commented):
I guess I was quite distracted when I looked at this yesterday, sorry about that :(.

In any case, it looks good, one note just a rebase change and otherwise minor gripes.

jkczyz (commented):
Took largely a high-level pass. Looking pretty good.

lightning/src/util/events.rs
valentinewallace: I don't have a strong opinion. Matt pointed out in [this comment](https://github.com/rust-bitcoin/rust-lightning/pull/1053#discussion_r694306740) that there could be value in users knowing whether to retry a portion of a payment vs. retrying the entire payment. @TheBlueMatt I tend to agree that this might make `MPPFragmentFailed` a bit redundant, thoughts?

lightning/src/ln/channelmanager.rs
valentinewallace: I guess this is pretty much equivalent to a u64 integer? Not sure I follow re: the tuple idea, let me know what `type` of `pending_outbound_payments` you're thinking of

lightning/src/ln/channelmanager.rs
valentinewallace: This avoids duplicating a long-ish log message (both `else`'s would need one). Lmk if you prefer duplicate logs or if there's a better way.

lightning/src/util/events.rs
valentinewallace: I always mess up these types of rebases, gonna let this one slide if that's OK :sweat_smile:

lightning/src/ln/channelmanager.rs
valentinewallace: This is used to correlate multiple MPP parts, there isn't an `MppId` per-part, if that makes sense

lightning/src/util/events.rs
jkczyz: I could break out the relevant commits into their own PR if it helps. I think from a retry perspective, though, we'd need to know the part amount unless `InvoicePayer` tracks that.

lightning/src/ln/channelmanager.rs
jkczyz: Ah, I see. Thanks for explaining.

lightning/src/ln/channelmanager.rs
```diff
@@ -3099,18 +3165,23 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
-					});
-				} else {
+				let mut session_priv_bytes = [0; 32];
+				session_priv_bytes.copy_from_slice(&session_priv[..]);
+				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
+				let found_payment = loop {
+					if let Some(mut sessions) = outbounds.remove(&mpp_id) {
+						if sessions.remove(&session_priv_bytes) {
+							self.pending_events.lock().unwrap().push(
+								events::Event::PaymentSent { payment_preimage }
```
jkczyz, line 3177: IIUC, an MPP will only result in one `PaymentSent` event -- let me know if I misunderstood. With that in mind, I was thinking of adding the successful path to `PaymentSent` for scoring. That way a channel's score could be improved if successfully routing through it. But if there is only one `PaymentSent` event for MPP, then such improvements will only be possible for one path. Would it be possible to include all the paths for `PaymentSent` event? Or would that require waiting for all paths to succeed before sending such an event?

valentinewallace: > IIUC, an MPP will only result in one PaymentSent event

Yup! We generate it as soon as one path of the MPP succeeds (i.e., we do not wait for all paths to succeed). Reasoning: as soon as we have the preimage, the rest of the paths are irrevocably "solved."

> Would it be possible to include all the paths for PaymentSent event?

Hmm, I think it's possible but not sure how pretty. Briefly looked but would have to dig more. I don't think it'd require waiting for all paths to succeed though, since we do store the path for pending payments in `HTLCSource`s

TheBlueMatt: I think as long as the receipient is "honest" (or at least wants their money and isn't buggy), we know all the paths succeeded when we get the first preimage here, so I think it would be "correct", but code-wise its probably quite tricky - the actual `HTLCSource` is stored away in the HTLC set of some other channel so not readily available here. Maybe we'll need a separate event here to indicate further paths that succeeded?

lightning/src/util/events.rs
TheBlueMatt: Discussed this one offline, but we just need to make sure that we expose to users somehow the distinction between a *payment* failing and an MPP fragment failing, as they have significant implication for what the user does in response. We *really* don't want users to think that a `PaymentFailed` event implies the payment failed, when it was only one mpp fragment that failed.

lightning/src/ln/channelmanager.rs
```diff
@@ -5039,12 +5131,22 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> Writeable f
...
+		for (_, outbounds) in pending_outbound_payments.iter() {
+			for outbound in outbounds.iter() {
+				outbound.write(writer)?;
+			}
 		}
 
-		write_tlv_fields!(writer, {});
+		let pending_outbound_payments = Some(pending_outbound_payments);
+		write_tlv_fields!(writer, {
+			(0, pending_outbound_payments, option),
```
TheBlueMatt, line 5148: This doesn't need the `Some` wrapper and should probably be odd.

```suggestion
		write_tlv_fields!(writer, {
			(1, pending_outbound_payments, required),
```

lightning/src/ln/channelmanager.rs
```diff
@@ -5586,20 +5697,13 @@ mod tests {
 		nodes[0].node.handle_revoke_and_ack(&nodes[1].node.get_our_node_id(), &bs_third_raa);
 		check_added_monitors!(nodes[0], 1);
 
-		// There's an existing bug that generates a PaymentSent event for each MPP path, so handle that here.
```
TheBlueMatt, line 5589: nit: Is it worth having a comment here that explicitly calls out that we only get one event here?

lightning/src/util/events.rs
```diff
@@ -112,8 +112,12 @@ pub enum Event {
 		/// payment is to pay an invoice or to send a spontaneous payment.
 		purpose: PaymentPurpose,
 	},
-	/// Indicates an outbound payment we made succeeded (ie it made it all the way to its target
+	/// Indicates an outbound payment we made succeeded (i.e. it made it all the way to its target
 	/// and we got back the payment preimage for it).
+	///
+	/// Note for MPP payments: we generate this event as soon as one of the payment paths succeeds. In
+	/// the case of a buggy payee, this may be followed by a PaymentFailed event if subsequent path(s)
```
TheBlueMatt, line 119: I don't believe it can be *followed* by a PaymentFailed, but it could be preceded by. It may be worth mentioning that users SHOULD treat this as a full payment sent, and MUST NOT attempt to retry parts of the payment.

lightning/src/util/events.rs
```diff
@@ -319,15 +328,18 @@ impl MaybeReadable for Event {
 					let mut payment_hash = PaymentHash([0; 32]);
 					let mut rejected_by_dest = false;
 					let mut network_update = None;
+					let mut all_paths_failed = true;
 					read_tlv_fields!(reader, {
 						(0, payment_hash, required),
 						(1, network_update, ignorable),
 						(2, rejected_by_dest, required),
+						(3, all_paths_failed, required),
```
TheBlueMatt, line 336: `required` implies that we'll fail to read old versions that are missing the field. I think we should just initialize it to `Some(true)` and unwrap it below, keeping the broken behavior for events that were pending.

lightning/src/util/events.rs
```diff
@@ -138,6 +142,10 @@ pub enum Event {
 		/// [`NetworkGraph`]: crate::routing::network_graph::NetworkGraph
 		/// [`NetGraphMsgHandler`]: crate::routing::network_graph::NetGraphMsgHandler
 		network_update: Option<NetworkUpdate>,
+		/// For both single-path and multi-path payments, this is set if all paths of the payment have
+		/// failed. This will be set to false if (1) this is an MPP payment and (2) other parts of the
+		/// larger MPP payment were still in flight when this event was generated.
+		all_paths_failed: bool,
```
TheBlueMatt, line 148: We really need to rename `PaymentFailed` after this IMO - maybe `PaymentPartFailed` or `HTLCFailed`, because it definitely no longer means the whole payment failed. I'm ok with pushing this to a followup if you just wanna get this merged without blowing up the diff.

TheBlueMatt (commented):
Ok, A few nits and minor things here and there but I'm basically happy.

lightning/src/ln/channelmanager.rs
jkczyz: Hmmm... yeah, that may be a good argument for separating path succeeding into its own event. Alternatively, `ChannelManager` could be parameterized by the scorer, which it would call as needed instead of using events. Though that could introduce the possibility of reentrancy. Maybe not the best idea and would be inconsistent with network updates.

Related to the [other discussion](https://github.com/rust-bitcoin/rust-lightning/pull/1053#discussion_r710621797), I guess for multi-path payments the user (or `InvoicePayer`) ultimately decides if a payment has failed when it no longer wants to retry failed paths (and hasn't received the preimage). Likewise for single-path payments. Either the user has the preimage (success) or doesn't (possible failure) and needs to decided whether to give up (i.e., no longer retry or wait).

With that in mind, maybe having a failed or successful payment at the event level isn't the right abstraction? Instead, there's `PreimageReceived`, `SuccessfulPaymentPath`, and `FailedPaymentPath`. Maybe the last two could be named in terms of HTLCs instead? Sorta thinking out loud on this, so let me know if I'm off on something.

lightning/src/util/events.rs
jkczyz: Drafted https://github.com/rust-bitcoin/rust-lightning/pull/1077

lightning/src/ln/channelmanager.rs
TheBlueMatt: > Hmmm... yeah, that may be a good argument for separating path succeeding into its own event. 

I think that makes sense. It's a bit awkward to do something different for success and failure but the semantics are pretty different and Val *just* removed the separate failure event from path event. For path success the probability increases can happen later, we're not trying to make sure we get them in before retrying a new path so I think it's fine.

> Alternatively, ChannelManager could be parameterized by the scorer, which it would call as needed instead of using events. Though that could introduce the possibility of reentrancy. Maybe not the best idea and would be inconsistent with network updates.

Yea, I don't think the complexity of the reentrancy is worth it there.

> With that in mind, maybe having a failed or successful payment at the event level isn't the right abstraction? Instead, there's PreimageReceived, SuccessfulPaymentPath, and FailedPaymentPath. Maybe the last two could be named in terms of HTLCs instead? Sorta thinking out loud on this, so let me know if I'm off on something.

I think that kinda makes sense. I just suggested renaming PaymentFailed to HTLC/Path/PartFailed. For preimage receipt it does mean you need to treat the payment as sent, though, so I'm not sure as really need to swap that one.

lightning/src/util/events.rs
jkczyz: Agree with renaming `PaymentFailed`. I do like using HTLC in the name as mentioned in another comment. Presumably these won't be deduplicated?

However, I'm still uncertain about adding the `all_paths_failed` field. I feel like I might be missing something, though. How would the retrier make use of it? If false, it needs to retry for the corresponding part unless it has received a `PaymentSent`. If true, doesn't it need to do the same thing? If not, is there any overlap with `rejected_by_dest`?

TheBlueMatt: > However, I'm still uncertain about adding the all_paths_failed field. I feel like I might be missing something, though. How would the retrier make use of it? If false, it needs to retry for the corresponding part unless it has received a PaymentSent. If true, doesn't it need to do the same thing?

I believe we'll need a different call in ChannelManager to retry a failed MPP path, as we won't be able to just blindly call send_payment again to keep state consistent. Thus, we'll need to decide which function we call based on that bool.

More generally, the bool isn't *just* for the retrier - its also to tell the user if a *payment* has failed, or if a *path* has failed. Until you either see `all_paths_failed` or `PaymentSent`, you cannot consider a payment resolved.

> If not, is there any overlap with rejected_by_dest?

They're two very different things - whether a final node rejected the payment or not, we can't consider a payment failed unless we get all the paths failed.

lightning/src/util/events.rs
valentinewallace: > and MUST NOT attempt to retry parts of the payment.

I guess that might not be helpful since they probably retried already when they received the preceding `PaymentFailed`?

I'm mildly in favor of removing this comment entirely, but I'll amend with your suggestion for now :)

TheBlueMatt: > I guess that might not be helpful since they probably retried already when they received the preceding PaymentFailed?

Ah, true lol.

> I'm mildly in favor of removing this comment entirely, but I'll amend with your suggestion for now :)

I think its worth at least mentioning that you can get a `PaymentFailed` followed immediately by a `PaymentSent`, which is confusing, and we should make sure users are aware of how they should treat that (payment success).

lightning/src/util/events.rs
jkczyz: > I believe we'll need a different call in ChannelManager to retry a failed MPP path, as we won't be able to just blindly call send_payment again to keep state consistent. Thus, we'll need to decide which function we call based on that bool.
> 
> More generally, the bool isn't _just_ for the retrier - its also to tell the user if a _payment_ has failed, or if a _path_ has failed. Until you either see `all_paths_failed` or `PaymentSent`, you cannot consider a payment resolved.
> 

Ah, that's right. For the situation where we retry some paths of an MPP but they all ultimately fail, would the retrier then retry the full payment? Seems it may complicate the retrier as either there are two levels of retries (paths and full payment) or payments that are MPP need to be treated differently from non-MPP. But if we want to avoid the two-levels of retries, we'd also need to know if the payment was MPP. But I don't think the bool will tell us that?

I had hoped this could be generalized in some way where we are always just retrying paths regardless of whether it's MPP or not. But I guess that would require complicating the state tracking in `ChannelManager`, possibly such that the user would need to say when they have given up (and thus considered the payment failed). 🙁

jkczyz: Actually, come to think of it the retrier will know how many paths it gets back from the router, so this should be sufficient.

TheBlueMatt: > Ah, that's right. For the situation where we retry some paths of an MPP but they all ultimately fail, would the retrier then retry the full payment?

Yes, I believe you generally would, as long as you haven't run out of time/retries.

> Seems it may complicate the retrier as either there are two levels of retries (paths and full payment) or payments that are MPP need to be treated differently from non-MPP. But if we want to avoid the two-levels of retries, we'd also need to know if the payment was MPP. But I don't think the bool will tell us that?

Yes, the complexity is somewhat unavoidable, I think. Ultimately the channelmanager needs to know, when the retry HTLC(s) are sent, whether its an MPP retry or not. If the full MPP has failed, the channelmanager will (have to) forget about the full payment, so the retrier needs to inform the channelmanager of this. The way I was thinking about it was that the retrier would:
 * if `all_paths_failed` is *not* set, call `channelmanager.retry_mpp_part(new_route)` and the channelmanager would look up the pending mpp data, generate the appropriate retry for only the given path, and send the new part(s),
 * if the above call fails with an `Err(NoPendingMPPFound)` or `all_paths_failed` *is* set, the payment retrier will generate a route for the full payment value and call `send_payment`.

I'm not really sure how we could handle this better on the channelmanager end, sadly.

> Actually, come to think of it the retrier will know how many paths it gets back from the router, so this should be sufficient.

Does the retrier need to store the pending route? I figured it would not.

TheBlueMatt: > I'm not really sure how we could handle this better on the channelmanager end, sadly.

The only alternative, really, would be to change the API to allow `retry_mpp_part` to reconstruct the state if you provide the full payment value, but I'm not sure that's ideal - if we get a `PaymentFailed` that raced the first `PaymentFailed` and is pending in the queue, with the last node having failed the payment, we *shouldn't* retry, and this API change would result in us retrying, but only part of the MPP.

lightning/src/util/events.rs
valentinewallace: Ah, I thought if `all_paths_failed` was an `option` `fieldty` in `read_tlv_fields`, it had to be an `option` in `write_tlv_fields` also. Updating mental model since they def can be asymmetric.

Sorry, I finally drew out the 2x2 tlv ser matrix (`type` odd/even is for old clients reading new, `fieldty` option/required is for new clients reading old)

TheBlueMatt: Right, everything is written the same whether its an option or not, option just declares whether we require it to be present or not, and odd/even declares whether old versions that don't understand it will fail or ignore.

lightning/src/util/events.rs
valentinewallace: Prefer a quick follow-up for renaming `PaymentFailed`

valentinewallace:
Think I've addressed outstanding feedback

lightning/src/util/events.rs
jkczyz: > Does the retrier need to store the pending route? I figured it would not.

Correct. I was thinking it would see that `get_route` returned multiple paths and note that (as a bool) for later. But I guess that isn't necessary if we allow retrying the full payment as well.

For counting retries, were you thinking the number of attempts across all paths and full payments would be counted together? Or would each path get its own attempts? Was thinking the former for simplicity.

As another related concern, what if `get_route` returns multiple paths when attempting to retry a part? Seems like this might be desirable, but `retry_mpp_part` would need to be able to handle this. Maybe that is what you meant by "part(s)".

TheBlueMatt: > For counting retries, were you thinking the number of attempts across all paths and full payments would be counted together? Or would each path get its own attempts? Was thinking the former for simplicity.

Dunno, should count something, I'll leave it up to the author :p.

> As another related concern, what if get_route returns multiple paths when attempting to retry a part? Seems like this might be desirable, but retry_mpp_part would need to be able to handle this. Maybe that is what you meant by "part(s)".

Yes, ChannelManager should be able to handle this seamlessly.

lightning/src/ln/functional_test_utils.rs
```diff
@@ -1287,77 +1289,102 @@ pub fn send_payment<'a, 'b, 'c>(origin: &Node<'a, 'b, 'c>, expected_route: &[&No
...
-			Event::PaymentFailed { payment_hash, rejected_by_dest, .. } => {
-				assert_eq!(payment_hash, our_payment_hash);
-				assert!(rejected_by_dest);
-			},
-			_ => panic!("Unexpected event"),
+		for (idx, node) in expected_route.iter().rev().enumerate().skip(1) {
+			assert_eq!(expected_next_node, node.node.get_our_node_id());
+			if next_msgs.is_some() {
+				node.node.handle_update_fail_htlc(&prev_node.node.get_our_node_id(), &next_msgs.as_ref().unwrap().0);
+				let last_node = !(skip_last && idx == expected_route.len() - 1);
```
TheBlueMatt, line 1335: This is equivalent to `!skip_last || idx != expected_route.len() - 1`, which I find more readable, and which is pretty different from "last node". I'm not really sure what is supposed to be going on here, but note that all tests pass if I set last_node to `idx == expected_route.len() - 1 && !skip_last` or, even, `idx == expected_route.len() - 1` (which I think is what really should be here).

lightning/src/ln/functional_test_utils.rs
```diff
@@ -1287,77 +1289,102 @@ pub fn send_payment<'a, 'b, 'c>(origin: &Node<'a, 'b, 'c>, expected_route: &[&No
...
-			// second-to-last node!
-			update_fail_dance!(node, prev_node, skip_last && idx == expected_route.len() - 1);
-		}
-
-		let events = node.node.get_and_clear_pending_msg_events();
-		if !skip_last || idx != expected_route.len() - 1 {
-			assert_eq!(events.len(), 1);
-			match events[0] {
-				MessageSendEvent::UpdateHTLCs { ref node_id, updates: msgs::CommitmentUpdate { ref update_add_htlcs, ref update_fulfill_htlcs, ref update_fail_htlcs, ref update_fail_malformed_htlcs, ref update_fee, ref commitment_signed } } => {
+	macro_rules! msgs_from_ev {
```
TheBlueMatt, line 1301: Note that as this is only used once you could inline it.

TheBlueMatt (approved):

lightning/src/ln/channelmanager.rs
jkczyz: This is the cleanest I came up with:

```rust
let found_payment = if let Some(mut sessions) = outbounds.remove(&mpp_id) {
	sessions.remove(&session_priv_bytes)
} else { false };
```

Then condition on `found_payment` to either push the event or log.

jkczyz (commented):
For clarification, are we still sending `PaymentFailed` events for each path? If so, could you update the PR title and description?

lightning/src/ln/channelmanager.rs
```diff
@@ -2873,14 +2882,21 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
+						log_trace!(self.logger, "Received duplicative fail for HTLC with payment_hash {}", log_bytes!(payment_hash.0));
+						return;
+					}
+					if sessions.get().len() == 0 {
+						sessions.remove();
+					}
+				} else {
 					log_trace!(self.logger, "Received duplicative fail for HTLC with payment_hash {}", log_bytes!(payment_hash.0));
-					return;
+					return
```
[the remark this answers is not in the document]
valentinewallace: is there an advantage to keeping the semicolon?

lightning/src/ln/functional_test_utils.rs
valentinewallace: > note that all tests pass if I set last_node to idx == expected_route.len() - 1 && !skip_last or, even, idx == expected_route.len() - 1

I get test failures with both of those^ :thinking: 

> This is equivalent to !skip_last || idx != expected_route.len() - 1, which I find more readable, and which is pretty different from "last node".

That is more readable, and yeah the name was wrong. Took the suggestion and renamed

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
valentinewallace: I guess I like MppId because that speaks to the actual purpose of the ID (correlating paths)

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
jkczyz: Mostly to keep the blame layer clean.

lightning/src/ln/channelmanager.rs
valentinewallace: Cleaner, thanks

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
jkczyz: I don't have a strong opinion, so feel free to ignore. But I'd think naming after what is being identified is future-proof if it is later used for another purpose, too.

lightning/src/ln/functional_test_utils.rs
TheBlueMatt: > I get test failures with both of those^ thinking

lol sorry was running out of the coffee shop and ran cargo test in the wrong directory :facepalm:.

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
valentinewallace: Gotcha, guess I'll pass for now because this seems quite close, but it's a fair point

jkczyz (approved):
ACK c828ff42c006abf3ceffe552dfe510f9f3538e11. LGTM!

PR #1120, opened 2021-10-13
Pre-scoring test clean-ups

jkczyz:
Replaces `get_route` with `get_route_and_payment_hash` in tests wherever possible, since a scorer will need to be passed to `get_route`. Also includes some fairly trivial test clean ups.


jkczyz:
I left a few TODO comments in the primary commit where I was unsure of the change. So please take a close look at those. Everything else should be fairly mechanical.

lightning/src/ln/shutdown_tests.rs
```diff
@@ -71,6 +71,9 @@ fn pre_funding_lock_shutdown_test() {
 	check_closed_event!(nodes[1], 1, ClosureReason::CooperativeClosure);
 }
 
+// TODO: Determine why get_route fails when replacing get_payment_preimage_hash with
```
TheBlueMatt, line 74: You have another todo here.

lightning/src/ln/functional_tests.rs
```diff
@@ -7906,6 +7740,8 @@ fn test_bump_penalty_txn_on_revoked_htlcs() {
 		assert_ne!(node_txn[0].input[0].previous_output, node_txn[2].input[0].previous_output);
 		assert_ne!(node_txn[1].input[0].previous_output, node_txn[2].input[0].previous_output);
 
+		// TODO: Determine why replacing get_route with get_route_and_payment_hash above causes
```
TheBlueMatt, line 7743: Because the tx ordering is based on the payment hash and we end up with a different hash, I assume? Or something like that, but maybe we should just drop the TODO and open a general "move the remaining explicit route calls to the routing macro" issue?

lightning/src/ln/functional_tests.rs
```diff
@@ -4025,36 +3972,28 @@ fn do_test_holding_cell_htlc_add_timeouts(forwarded_htlc: bool) {
...
 		nodes[1].node.send_payment(&route, first_payment_hash, &Some(first_payment_secret)).unwrap();
 	}
 	assert_eq!(nodes[1].node.get_and_clear_pending_msg_events().len(), 1);
 	check_added_monitors!(nodes[1], 1);
 
 	// Now attempt to route a second payment, which should be placed in the holding cell
-	let (_, second_payment_hash, second_payment_secret) = get_payment_preimage_hash!(nodes[2]);
+	let sending_node = if forwarded_htlc { &nodes[0] } else { &nodes[1] };
+	let (route, second_payment_hash, _, second_payment_secret) = get_route_and_payment_hash!(sending_node, nodes[2], 100000);
+	// TODO: Why use a different payment secret here?
```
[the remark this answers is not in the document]
jkczyz: Was unsure because they both used `second_payment_hash` but a different secret.

lightning/src/ln/shutdown_tests.rs
jkczyz: Mainly to note why it may be that a simple replacement doesn't work here. Can drop it if you prefer.

lightning/src/ln/functional_tests.rs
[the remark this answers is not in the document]
TheBlueMatt: Ah, I was comparing it to the first send event further up, its just a typo I believe, feel free to drop it if the test still passes and always use the second secret.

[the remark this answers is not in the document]
jkczyz: Updated to use the second secret.

lightning/src/ln/functional_tests.rs
```diff
@@ -9376,11 +9189,7 @@ fn test_keysend_payments_to_private_node() {
 	nodes[1].node.peer_connected(&payer_pubkey, &msgs::Init { features: InitFeatures::known() });
 
 	let _chan = create_chan_between_nodes(&nodes[0], &nodes[1], InitFeatures::known(), InitFeatures::known());
-	let network_graph = &nodes[0].net_graph_msg_handler.network_graph;
-	let first_hops = nodes[0].node.list_usable_channels();
-	let route = get_keysend_route(&payer_pubkey, &network_graph, &payee_pubkey,
```
TheBlueMatt, line 9381: This seems to be the only place we call `get_keysend_route` in tests through a private channel, so I think best to keep it.

jkczyz: Reverted and updated the other keysend test to use `get_keysend_route`.

valentinewallace (approved):

PR #1155, opened 2021-11-03
Introduce graph sync crate

arik-so:
Its use is for fast-forwarding through gossip data downloaded from a server.

Currently still in draft status, pending better code organization and tests.


TheBlueMatt:
Cool. I assume you're still working on compression for the objects?

arik-so:
Yes, I'm trying to build an encoder that is distinct from gzip, and just RLE, preferably with Huffman on top.

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,241 @@
+use core::ops::Deref;
```
TheBlueMatt, line 1: Can you set the `deny(missing_docs)` and `forbid(unsafe_code)` attrs?

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+use lightning::routing::network_graph;
+use lightning_block_sync::http::*;
+
+use crate::error::GraphSyncError;
+
+pub mod error;
+pub mod processing;
+
+pub async fn sync_network_graph_with_url<CS: Deref>(
+    network_graph: &network_graph::NetworkGraph,
```
TheBlueMatt, line 14: Can we use tabs here too?

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    sync_port: u16,
+    sync_path: &str,
+    chain_source: &Option<CS>,
+) -> Result<(), GraphSyncError>
+where
+    CS::Target: chain::Access,
+{
+    // make sure there is precisely one leading slash
+    let canonical_path = format!("/{}", sync_path.trim_start_matches("/"));
+    println!(
```
TheBlueMatt, line 25: Lets add a logger parameter and re-use our logging stuff?

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+
+pub mod error;
+pub mod processing;
+
+pub async fn sync_network_graph_with_url<CS: Deref>(
+    network_graph: &network_graph::NetworkGraph,
+    sync_host: &str,
+    sync_port: u16,
+    sync_path: &str,
+    chain_source: &Option<CS>,
```
TheBlueMatt, line 18: Lets just drop the `chain_source`, we dont have signatures so its not like we can validate anything.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    // shadow variable for ownership release after length-restricted reads
+    let mut read_cursor = read_cursor;
+
+	let announcement_count: BigSize = Readable::read(read_cursor)?;
+	for _ in 0..announcement_count.0 {
+		let message_length: BigSize = Readable::read(read_cursor)?;
+		let mut restricted_cursor = read_cursor.take(message_length.0);
+		let channel_announcement: msgs::UnsignedChannelAnnouncement =
+			Readable::read(&mut restricted_cursor)?;
```
TheBlueMatt, line 56: Hmmm, we can cut the size here by a good chunk if we drop the two on-chain public keys. We only need the node public keys. I'm also curious if there's a way to avoid including the full node public keys for nodes that already have a channel (ie by providing an SCID + direction flag). I'm not sure if that'd be too slow on the decode side, though, need benchmarks.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,390 @@
...
+			};
+
+			let mut excess_data = Vec::new();
+			restricted_cursor.read_to_end(&mut excess_data)?;
+			read_cursor = restricted_cursor.into_inner();
+
+			let synthetic_update = UnsignedChannelUpdate {
+				chain_hash: Default::default(),
+				short_channel_id,
+				timestamp: current_timestamp,
```
TheBlueMatt, line 105: Hmmmm, I think we need to use the update's real timestamp as otherwise we may get an update from an HTLC failure or otherwise and reject it cause the timestamp is "old".

arik-so: I will get back to this one in a separate commit.

lightning-block-sync/src/http.rs
```diff
@@ -97,7 +97,7 @@ impl<'a> std::net::ToSocketAddrs for &'a HttpEndpoint {
 }
 
 /// Client for making HTTP requests.
-pub(crate) struct HttpClient {
+pub struct HttpClient {
```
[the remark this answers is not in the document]
arik-so: I'd honestly be perfectly fine to remove the HTTP client functionality from the graph sync crate altogether. Half the compression benefits come from gzipping, which it doesn't support anyway.

lightning-graph-sync/src/processing.rs
arik-so: Isn't the timestamp only tangentially related to the time of the actual update? I thought nodes had some logic to update the timestamp of a particular update when gossiping anyway – or is the timestamp part of the signed data when a signature is present?

arik-so: Additionally, if the timestamp inaccuracy has implications on HTLC handling, wouldn't that mean that gossip data is slightly more than just a DDoS risk vector?

lightning-graph-sync/src/lib.rs
arik-so: Done

lightning-graph-sync/src/lib.rs
arik-so: This doesn't even need a print, I had just used it for debugging, so I removed it instead.

lightning-graph-sync/src/lib.rs
arik-so: Done (I created an .editorconfig locally, we might wanna have one in the project tree)

lightning-graph-sync/src/lib.rs
arik-so: Done

lightning-block-sync/src/http.rs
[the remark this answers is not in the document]
arik-so: For the time being, I actually introduced a feature flag to gate HTTP client access. However, I stand by my previous point. Arguably, removing this functionality from here altogether would make more sense.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,75 @@
...
+For the purpose of rapid syncing, we have deviated from the channel update format
+specified in BOLT 7 significantly. Our custom channel updates are structured as follows:
+
+1. `short_channel_id`: `u64`
+2. `custom_channel_flags`: `u8`
+3. `update_data`
+
+Specifically, our custom channel flags break down like this:
+
+| 128                 | 64 | 32 | 16 | 8 | 4   | 2                | 1         |
```
jkczyz, line 34: Is there some significance in the spacing difference in the columns for `8` and `4`?

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+
+	let response_result = http_client
+		.get::<BinaryResponse>(&canonical_path, sync_host)
+		.await;
+
+	let response = response_result?;
+	let response_bytes = response.0;
+
+	let update_result = processing::update_network_graph(&network_graph, &response_bytes[..])?;
+	Ok(update_result)
```
jkczyz, line 41: It's a bit more idiomatic to forgo the intermediary variables:

```rust
let mut http_client = HttpClient::connect(&http_endpoint)?;
let response_bytes = http_client.get::<BinaryResponse>(&canonical_path, sync_host).await?.0;
processing::update_network_graph(&network_graph, &response_bytes[..], &chain_source)
```

lightning-graph-sync/README.md
arik-so: no, my IDE just decided to add it. I'll remove.

lightning/src/routing/network_graph.rs
```diff
@@ -1170,6 +1190,70 @@ impl NetworkGraph {
...
+								directional_info.htlc_minimum_msat = htlc_minimum_msat;
+							}
+							ChannelMutationProperty::FeeBaseMsat(fee_base_msat) => {
+								directional_info.fees.base_msat = fee_base_msat;
+							}
+							ChannelMutationProperty::FeeProportionalMillionths(fee_proportional_millionths) => {
+								directional_info.fees.proportional_millionths = fee_proportional_millionths;
+							}
+							ChannelMutationProperty::HtlcMaximumMsat(htlc_maximum_msat) => {
+								// repeat the same HTLC max checks from update_channel_intern
```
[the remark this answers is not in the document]
arik-so: that's a good idea. I'll try to make that happen.

lightning-graph-sync/src/lib.rs
arik-so: sounds good

lightning/src/routing/network_graph.rs
```diff
@@ -1170,6 +1190,74 @@ impl NetworkGraph {
...
+	/// Override a (directed) channel's individual properties without having a full update
+	/// This method should be used exclusively in conjunction with rapid graph sync, which is intended
+	/// for mobile devices.
+	///
+	/// `short_channel_id`: Channel to be updated
+	///
+	/// `flags`: Channel flags to set. Least significant bit serves as directional indicator
+	///
+	/// `mutations: Vec<ChannelMutationProperty>`: Array of properties to mutate
+	pub fn override_directional_channel_properties(&self, short_channel_id: u64, flags: u8, timestamp: u32, chain_hash: BlockHash, mutations: ChannelMutationProperties) -> Result<(), LightningError> {
```
jkczyz, line 1202: Given the use of `CustomChannelUpdate` in the docs, would it make sense to rename `ChannelMutationProperties` to that and include the remaining parameters as fields? Or maybe use the name `IncrementalChannelUpdate` given this is one variation of `CustomChannelUpdate`.

lightning/src/routing/network_graph.rs
```diff
@@ -1170,6 +1190,74 @@ impl NetworkGraph {
...
+							short_channel_id,
+							timestamp,
+							flags,
+							cltv_expiry_delta: directional_info.cltv_expiry_delta,
+							htlc_minimum_msat: directional_info.htlc_minimum_msat,
+							htlc_maximum_msat,
+							fee_base_msat: directional_info.fees.base_msat,
+							fee_proportional_millionths: directional_info.fees.proportional_millionths,
+							excess_data: vec![]
+						};
```
jkczyz, line 1233: The `UnsignedChannelUpdate` for the full update is constructed in `lightning-graph-sync` while the incremental update is formed here. For consistency and to avoid blurring layers, could we do both in `lightning-graph-sync`? (Related: see next comment about locking.)

Then you can make a `CustomChannelUpdate` enum in `lightning-graph-sync` with variants for full and incremental with and implementation of `Readable` there and possibly conversion functions to `UnsignedChannelUpdate`.

lightning/src/routing/network_graph.rs
```diff
@@ -1170,6 +1190,74 @@ impl NetworkGraph {
...
+							synthetic_update.htlc_maximum_msat = htlc_maximum_msat;
+						}
+						synthetic_update
+					}else{
+						return Err(LightningError{err: "Couldn't find previous directional data for update".to_owned(), action: ErrorAction::IgnoreError});
+					}
+				}
+			}
+		};
+		self.update_channel_unsigned(&channel_update)
```
jkczyz, line 1258: This may need to be atomic with the earlier read. What would happen in the following scenario?
- Form an `UnsignedChannelUpdate` from incremental update using the read-locked `NetworkGraph` above.
- Apply `ChannelUpdate` from failed payment
- Apply earlier formed `UnsignedChannelUpdate`

Seems the incremental update would override one we just received from a failed payment.

lightning/src/routing/network_graph.rs
arik-so: hm, interesting point. In the docs, `CustomChannelUpdate` refers to the over-the-wire format that is a deviation from the BOLT7 spec, but hadn't really considered there to be a necessity for the internal representation of the delta to abide by that naming.

I think this point is somewhat related to your next one regarding where incremental and full updates are constructed and applied.

lightning/src/routing/network_graph.rs
arik-so: I agree, and I would have loved to pull the incremental update out of the lightning crate. The reason I haven't is because we need access to the previous directional channel info, and that for sure seems like a property that should not be made public. Very open to suggestions. 

Additionally, conversion functions don't really work because it it is reliant on the previous directional channel info.

lightning/src/routing/network_graph.rs
arik-so: It shouldn't be a problem, because all the rapid sync updates are backdated a week, so any "true" update from direct gossip should result in this update being discarded.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,473 @@
...
+		&[76, 68, 75, 2] => true,
+		_ => {
+			return Err(GraphSyncError::ProcessingError(
+				"Prefix must equal 'LDK' and the byte 2".to_string(),
+			));
+		}
+	};
+
+	let chain_source: Option<Box<dyn lightning::chain::Access>> = None;
+	let current_timestamp = SystemTime::now()
```
TheBlueMatt, line 42: We're gonna get this from the update file itself, right?

TheBlueMatt (commented):
In whatever release we include this in, we'll need to fix the network graph data timeout logic to somehow not run until after graph sync completes on startup. I'm not sure how to do this, but it has to be in the same release as this (but can be a separate PR). Maybe open an issue to track it?

lightning-graph-sync/src/processing.rs
arik-so: Yeah, I was actually debating as to whether to include the timestamp in the update data. I'll modify the serialization accordingly.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,473 @@
...
+			short_channel_id,
+			node_id_1,
+			node_id_2,
+			bitcoin_key_1: node_id_1,
+			bitcoin_key_2: node_id_2,
+			excess_data: vec![],
+		};
+
+		let announcement_result = network_graph
+			.update_channel_from_unsigned_announcement(&synthetic_announcement, &chain_source);
```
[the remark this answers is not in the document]
jkczyz: > I'd much prefer to have a method on the `NetworkGraph` that says "add channel with minimal field set of X" and use that.

Isn't that a bit of a layer violation, too? Taking a concept from a higher layer (i.e., the incremental update) and defining it in a lower layer?

lightning/src/routing/network_graph.rs
jkczyz: > I agree, and I would have loved to pull the incremental update out of the lightning crate. The reason I haven't is because we need access to the previous directional channel info, and that for sure seems like a property that should not be made public. Very open to suggestions.
> 

Isn't the directional info already part of `NetworkGraph`'s public interface? We even have `ChannelInfo::as_direct_to`.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Mechanism
+
+The (presumed) server sends a compressed gossip response containing gossip data.
+The gossip data is formatted compactly, omitting signatures and opportunistically
+incremental.
+
+Essentially, the serialization structure is as follows:
+
+1. Fixed prefix bytes `76, 68, 75, 2` (the first three bytes are ASCII for `LDK`)
```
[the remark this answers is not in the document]
jkczyz: Ah, could you document this? Also, why not `1` if namespaced by "LDK"?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,473 @@
...
+			));
+		}
+	};
+
+	let chain_source: Option<Box<dyn lightning::chain::Access>> = None;
+	let current_timestamp = SystemTime::now()
+		.duration_since(UNIX_EPOCH)
+		.unwrap()
+		.as_secs() as u32;
+	let regressed_timestamp = current_timestamp - 7 * 24 * 3600; // back-date updates by one week
```
jkczyz, line 46: Could you make a constant and document _why_ backdating is used?

lightning-graph-sync/src/processing.rs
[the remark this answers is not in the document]
TheBlueMatt: Hmm, doesn't feel like one? Its not crazy for `NetworkGraph` to have a method that just says "manually add a new channel to the network graph, with these parameters". Its not really specific to this kind of sync mechanism, and could be used for several different sync mechanisms.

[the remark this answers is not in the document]
jkczyz: True. I had been thinking more about gossip as the layer.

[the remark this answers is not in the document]
TheBlueMatt: Right, we kinda have to layers, the gossip message handler, and the underlying `NetworkGraph`. I agree it'd be strange to put it in the gossip message handler, but we can put some of these update methods on the `NetworkGraph` itself.

lightning-graph-sync/src/processing.rs
arik-so: done

lightning/src/routing/network_graph.rs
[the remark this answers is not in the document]
arik-so: done

lightning-graph-sync/src/lib.rs
arik-so: done

lightning/src/routing/network_graph.rs
arik-so: I think I have found a way to do that.

lightning-graph-sync/src/processing.rs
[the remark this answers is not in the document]
arik-so: I'll think about possible approaches to this in the new few commits.

lightning-graph-sync/src/processing.rs
arik-so: done

lightning/src/routing/network_graph.rs
arik-so: done

lightning/src/routing/network_graph.rs
arik-so: obsolete now

lightning-graph-sync/README.md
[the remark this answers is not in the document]
arik-so: done, and switched to 1. I guess I thought it would be a courtesy thing, but w/e

lightning-graph-sync/src/processing.rs
arik-so: Will update and reinitialize the server to have received timestamp fields tonight, and then regenerate test vectors.

lightning/src/routing/network_graph.rs
```diff
@@ -1085,6 +1085,63 @@ impl NetworkGraph {
 		self.update_channel_from_unsigned_announcement_intern(msg, None, chain_access)
 	}
 
+	/// Update channel from partial announcement data received via rapid gossip
+	///
+	/// `timestamp: 64`: Timestamp the announcement was originally received.
+	///
+	/// All other parameters as used in `UnsignedChannelAnnouncement` fields.
+	pub fn update_channel_from_announcement_fragments(&self, short_channel_id: u64, timestamp: u64, features: ChannelFeatures, node_id_1: PublicKey, node_id_2: PublicKey) -> Result<(), LightningError>{
```
TheBlueMatt, line 1093: Lets just call this `add_channel` or something? This name is much more specific than it need be.

lightning/src/routing/network_graph.rs
```diff
@@ -1085,6 +1085,63 @@ impl NetworkGraph {
...
+		if node_id_1 == node_id_2 {
+			return Err(LightningError{err: "Channel announcement node had a channel with itself".to_owned(), action: ErrorAction::IgnoreError});
+		};
+
+		let node_1 = NodeId::from_pubkey(&node_id_1);
+		let node_2 = NodeId::from_pubkey(&node_id_2);
+		let chan_info = ChannelInfo {
+			features: features.clone(),
```
TheBlueMatt, line 1101: Why is this cloned?

lightning/src/routing/network_graph.rs
```diff
@@ -1085,6 +1085,63 @@ impl NetworkGraph {
...
+				// perhaps we could just return `Ok(())` here
+				return Err(LightningError{err: "Already have knowledge of channel".to_owned(), action: ErrorAction::IgnoreDuplicateGossip});
+			},
+			BtreeEntry::Vacant(entry) => {
+				entry.insert(chan_info);
+			}
+		};
+
+		let mut add_channel_to_node = | node_id: NodeId | {
+			match nodes.entry(node_id) {
```
TheBlueMatt, line 1125: Can you DRY this with the below function?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,506 @@
...
+
+	// shadow variable for ownership release after length-restricted reads
+	let mut read_cursor = read_cursor;
+
+	let chain_hash: BlockHash = Readable::read(read_cursor)?;
+
+	let announcement_count: u32 = Readable::read(read_cursor)?;
+	for _ in 0..announcement_count {
+		let features = Readable::read(read_cursor)?;
+		let short_channel_id = Readable::read(read_cursor)?;
```
TheBlueMatt, line 68: Can we sort by SCID and just include a variable-length-int difference from the previous one? That should cut the size down a lot, no?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,506 @@
...
+		let node_id_1 = Readable::read(read_cursor)?;
```
TheBlueMatt, line 69: We should probably do something to de-duplicate node_ids, no? Like maybe just do a list of node_ids to start and then put the index in that list here instead of the full node_id?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,506 @@
...
+				return Err(lightning_error.into());
+			}
+		}
+	}
+
+	let update_count: u32 = Readable::read(read_cursor)?;
+	for _current_update_index in 0..update_count {
+		// println!("update index: {}", _current_update_index);
+
+		let short_channel_id: u64 = Readable::read(read_cursor)?;
```
TheBlueMatt, line 92: Same here, maybe sort them and put a difference instead of the full one?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,506 @@
...
+		let synthetic_update = if channel_flags & 0b_1000_0000 == 0 {
+			// full update
+			// println!("channel flags: {:b}, scid: {}, full update", channel_flags, short_channel_id);
+
+			let cltv_expiry_delta: u16 = Readable::read(&mut read_cursor)?;
+			let htlc_minimum_msat: u64 = Readable::read(&mut read_cursor)?;
+			let fee_base_msat: u32 = Readable::read(&mut read_cursor)?;
+			let fee_proportional_millionths: u32 = Readable::read(&mut read_cursor)?;
+
+			let has_htlc_maximum_msat: u8 = Readable::read(read_cursor)?;
```
TheBlueMatt, line 107: Maybe set a flag value instead? ie MAX_UINT64 means "no htlc-max", and the server-side can cap it down to 21 million BTC without loss.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,506 @@
...
+		let channel_flags: u8 = Readable::read(read_cursor)?;
+
+		// flags are always sent in full, and hence always need updating
+		let standard_channel_flags = channel_flags & 0b_0000_0011;
+
+		let synthetic_update = if channel_flags & 0b_1000_0000 == 0 {
+			// full update
+			// println!("channel flags: {:b}, scid: {}, full update", channel_flags, short_channel_id);
+
+			let cltv_expiry_delta: u16 = Readable::read(&mut read_cursor)?;
```
TheBlueMatt, line 102: Why are we reading these all here? It looks like we can just set the flags and read them later anyway, might as well optimize a hair and not bother reading the fields that have flags assigned here and instead assign them default (ie today's-most-common) values and let the server skip them if they match.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,90 @@
...
+2. `custom_channel_flags`: `u8`
+3. `update_data`
+
+Specifically, our custom channel flags break down like this:
+
+| 128                 | 64 | 32 | 16 | 8 | 4 | 2                | 1         |
+|---------------------|----|----|----|---|---|------------------|-----------|
+| Incremental update? |    |    |    |   |   | Disable channel? | Direction |
+
+If the most significant bit is set to `0`, indicating a non-incremental update,
```
TheBlueMatt, line 53: nit: whitespace at EOL here and a few places. local `git show` should highlight them.

lightning-graph-sync/src/processing.rs
arik-so: I'm not sure this saves space, because if there is no HTLC maximum, sending a u8 takes up less space than a u64.

arik-so: we could do something interesting with BigSize?

arik-so: The sizes are equivalent when the number of present HTLCs == 7 * absent HTLCs. If there are more present HTLCs than absent ones, using the flag saves space, and otherwise, using the u8-prefix saves space. We could add a global u8 flag at the beginning specifying which encoding is used in this transmission, and set it depending on the relation of the set/unset HTLC maxima.

TheBlueMatt: My point was based on the assumption that the vast majority of full updates would have a htlc max, but, more importantly, see https://github.com/lightningdevkit/rust-lightning/pull/1155#discussion_r813359975

lightning-graph-sync/src/processing.rs
arik-so: I think I'm still missing something here. The flags are set when the value differs from the previous one, otherwise the server doesn't send the value.

lightning-graph-sync/src/processing.rs
arik-so: Based on testing, it does indeed appear that the 7:1 ratio of set/updated HTLC maxima vs removed HTLC maxima is preserved, so for now we'll just always use u64::MAX

lightning-graph-sync/src/processing.rs
arik-so: done, using BigSize

lightning-graph-sync/src/processing.rs
arik-so: done, passing index using BigSize. I also explored sorting the pubkeys, but the differences are too big to be useful.

lightning-graph-sync/src/processing.rs
arik-so: done, see https://github.com/lightningdevkit/rust-lightning/pull/1155#discussion_r822400030

lightning/src/routing/network_graph.rs
arik-so: done, I think

lightning/src/routing/network_graph.rs
arik-so: decloned

lightning/src/routing/network_graph.rs
arik-so: I came up with a different name, but `add_channel` is definitely not clear given the context of when this function should be called.

lightning-graph-sync/README.md
arik-so: I thought it's considered best practice to have a space at the end of a word prior to a newline, but I can remove it

lightning-graph-sync/src/processing.rs
TheBlueMatt: We already have `channel_flags` read outside the if block here, so we can take the reads at https://github.com/lightningdevkit/rust-lightning/blob/7ea4a63c9c68d68266ceca98eb41c7be2ad3a25f/lightning-graph-sync/src/processing.rs#L215-L241 and move them outside the if block. Then, here, in the `if channel_flags & 0b1000_000 == 0` block, assign each field a "default" that is the most common we see on the network today. That way, we can avoid reading a bunch of the fields if they're the most common field.

lightning-graph-sync/README.md
TheBlueMatt: Hmm, I've never heard of that, and my almost-default vim will automagically remove them when it line-wraps.

lightning-graph-sync/src/processing.rs
arik-so: I think I understand. I wonder whether it's the most efficient mechanism for incremental updates, given that here, we have the a trivariate combination of a) was the setting changed? b) was it set to the default? c) if it was not set to the default, which value was it set to? thereby requiring two bit flags instead of one.

TheBlueMatt: Hmm? when you see a "full update" treat it as "all fields were set to the default, except those with the changed flag".

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,591 @@
...
+			let channel = read_only_network_graph.channels().get(&short_channel_id);
+			let channel = if let Some(channel) = channel {
+				channel
+			} else {
+				return Err(LightningError {
+					err: "Couldn't find channel for update".to_owned(),
+					action: ErrorAction::IgnoreError,
+				}
+					.into());
+			};
```
jkczyz, line 199: FWIW, wherever you have an `Option` and need to return an error for `None` you can simplify with `ok_or_else` and `?`, e.g.:

```rust
			let channel = read_only_network_graph
				.channels()
				.get(&short_channel_id)
				.ok_or_else(|| LightningError {
					err: "Couldn't find channel for update".to_owned(),
					action: ErrorAction::IgnoreError,
				})?;

```

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,591 @@
...
+
+			let direction = channel_flags & 1u8;
+			let directional_info = if direction == 0 {
+				channel.one_to_two.as_ref()
+			} else {
+				channel.two_to_one.as_ref()
+			};
```
jkczyz, line 206: Consider making an inline method on `ChannelInfo` for this so you can easily chain it with the previous and/or following logic.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,591 @@
...
+		Readable::read(&mut read_cursor)?
+	} else {
+		0
+	};
+	let tentative_default_htlc_maximum_msat: u64 = if has_non_incremental_updates {
+		Readable::read(&mut read_cursor)?
+	} else {
+		0
+	};
+	let default_htlc_maximum_msat = if tentative_default_htlc_maximum_msat == u64::MAX {
```
jkczyz, line 147: Note that `u64::MAX` is only available as of 1.43 so you may need to use `u64::max_value()` here and elsewhere.

lightning/src/routing/network_graph.rs
```diff
@@ -1085,6 +1085,66 @@ impl NetworkGraph {
...
+				entry.insert(chan_info);
+			}
+		};
+
+		Self::add_channel_between_nodes(nodes, node_1, node_2, short_channel_id)?;
+
+		Ok(())
+	}
+
+	fn add_channel_between_nodes(mut nodes: RwLockWriteGuard<BTreeMap<NodeId, NodeInfo>>, node_id_a: NodeId, node_id_b: NodeId, short_channel_id: u64) -> Result<(), LightningError> {
```
jkczyz, line 1129: Does this need to return a `Result`? I'm seeing a build warning at one of the call sites.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,591 @@
...
+	}
+
+	previous_scid = None; // updates start at a new scid
+
+	let update_count: u32 = Readable::read(read_cursor)?;
+	if update_count < 1 {
+		return Ok(());
+	}
+	let non_incremental_update_presence_flag: u8 = Readable::read(read_cursor)?;
+	let has_non_incremental_updates = (non_incremental_update_presence_flag == 1);
```
jkczyz, line 120: Seeing a build warning here from usage of parentheses.

lightning-graph-sync/src/processing.rs
arik-so: this doesn't appear to be tripping up any CI tests, and max_value is deprecated. I'll look into why the CI tests aren't failing, because they should.

jkczyz: Hmm, may need to update `.github/workflows/build.yml` to include the new crate.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,576 @@
...
+	let retrogressed_timestamp = current_timestamp - TIMESTAMP_RETROGRESSION_DELTA; // back-date updates by one week
+
+	// shadow variable for ownership release after length-restricted reads
+	let mut read_cursor = read_cursor;
+
+	let chain_hash: BlockHash = Readable::read(read_cursor)?;
+	let latest_seen_timestamp: u32 = Readable::read(read_cursor)?; //
+	let emulated_last_seen_timestamp = u32::max(latest_seen_timestamp, retrogressed_timestamp);
+
+	let mut node_ids: Vec<PublicKey> = vec![];
```
jkczyz, line 70: Consider using `with_capacity` after reading the count.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,576 @@
...
+				excess_data: vec![],
+			}
+		} else {
+			// incremental update
+			// println!("channel flags: {:b}, incremental update", channel_flags);
+
+			let read_only_network_graph = network_graph.read_only();
+			let channel = read_only_network_graph
+				.channels()
+				.get(&short_channel_id)
```
jkczyz, line 184: If we have performance concerns with the `BTreeMap`, one possibility would be to traverse the (sorted) channel updates in parallel with the tree rather than performing look-ups for each update. That would be quite an involved change though.

jkczyz (commented):
Feel free to sqaush and rebase on `main` along with removing commented code, etc.

lightning/src/routing/network_graph.rs
```diff
@@ -674,6 +674,16 @@ impl ChannelInfo {
 		};
 		Some((DirectedChannelInfo { channel: self, direction }, source))
 	}
+
+	/// Returns a [`&ChannelUpdateInfo`] based on the direction flag.
+	/// The direction flag can be extracted from the channel flags using `channel_flags & 1u8`.
```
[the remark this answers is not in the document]
arik-so: Hm I considered that option, and didn't feel strongly either way. Happy to change it.

lightning-graph-sync/src/processing.rs
arik-so: Spot on

lightning/src/routing/network_graph.rs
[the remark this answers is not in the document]
jkczyz: yeah, I think I'd prefer passing the flags to avoid having user interpret them

lightning-graph-sync/src/processing.rs
jkczyz: yup, looks like you need to use `u64::max_value()`

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,568 @@
...
+					action: ErrorAction::IgnoreError,
+				})?;
+
+			let directional_info =
+				channel
+					.get_directional_info(channel_flags)
+					.ok_or(LightningError {
+						err: "Couldn't find previous directional data for update".to_owned(),
+						action: ErrorAction::IgnoreError,
+					})?;
```
jkczyz, line 190: Repeating previous lost comment: move `channel` up one line and reduce ident of following lines.

arik-so: rustfmt seems to strongly disagree with that FYI

TheBlueMatt:
We need a new fuzzer to test the deserialization logic here, should be pretty trivial to add, though.

lightning-graph-sync/src/processing.rs
jkczyz: lol... rustfmt may disagree with itself given it did the opposite a few lines up

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,576 @@
...
+
+use crate::error::GraphSyncError;
+
+/// Backdating is used a mechanism to avoid accidentally overriding higher-fidelity gossip updates.
+/// Current delta amount is set to a week flat. Note that gossip updates are purged at the
+/// two-week-mark.
+///
+/// In the future, we will probably start delaying the purging of unsigned updates to their
+/// four-week-marks, or perhaps simply trust a provided server-timestamp that marks
+/// data receipt, as opposed to originally signed value. TODO: TBD
```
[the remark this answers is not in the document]
TheBlueMatt: Oof, yea, that's a tough one, too - in the current design the server has to send every SCID of channels in the current graph in each update or users will just have channels that get more and more out of date and then eventually remove them. I think the only realistic solution is to track where each update came from and only time it out based on that info, sadly, with some ability for users to "convert" a graph from "server-sync" to "p2p-sync"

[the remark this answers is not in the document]
TheBlueMatt: I'm not sure that needs to be solved here, but we absolutely cannot ship this PR as-is without that followup, so would be nice to see it at least in a draft PR IMO.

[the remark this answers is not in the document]
arik-so: Yes, I'll try to have one up with the fuzzing for this one shortly.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,576 @@
...
+	// let chain_source: Option<Box<dyn lightning::chain::Access>> = None;
+	let current_timestamp = SystemTime::now()
+		.duration_since(UNIX_EPOCH)
+		.unwrap()
+		.as_secs() as u32;
+
+	let retrogressed_timestamp = current_timestamp - TIMESTAMP_RETROGRESSION_DELTA; // back-date updates by one week
+
+	// shadow variable for ownership release after length-restricted reads
+	let mut read_cursor = read_cursor;
```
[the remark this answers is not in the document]
arik-so: oh good point, addressed

lightning-graph-sync/src/processing.rs
arik-so: I did fix it, but I'm low key annoyed by this. I find with = and ==, parentheses really improve legibility.

lightning-graph-sync/src/processing.rs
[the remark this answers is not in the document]
jkczyz: Was this addressed in an un-pushed commit?

TheBlueMatt:
Needs rebase.

lightning-graph-sync/src/processing.rs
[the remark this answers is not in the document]
TheBlueMatt: I think we can remove this TODO now?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,565 @@
...
+	}
+
+	previous_scid = 0; // updates start at a new scid
+
+	let update_count: u32 = Readable::read(read_cursor)?;
+	if update_count < 1 {
+		return Ok(());
+	}
+	let non_incremental_update_presence_flag: u8 = Readable::read(read_cursor)?;
+	let has_non_incremental_updates = non_incremental_update_presence_flag == 1;
```
TheBlueMatt, line 112: Is it really worth saving the 26 static bytes for this? Lets simplify the code.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,565 @@
...
+			return Err(GraphSyncError::ProcessingError(
+				"Prefix must equal 'LDK' and the version byte 1".to_string(),
+			));
+		}
+	};
+
+	// let chain_source: Option<Box<dyn lightning::chain::Access>> = None;
+	let current_timestamp = SystemTime::now()
+		.duration_since(UNIX_EPOCH)
+		.unwrap()
```
TheBlueMatt, line 58: nit: lol this is why we don't use rustfmt, lets not make our code so vertical please :)

lightning-graph-sync/src/processing.rs
TheBlueMatt: No.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,565 @@
...
+		let features = Readable::read(read_cursor)?;
+
+		// handle SCID
+		let scid_delta: BigSize = Readable::read(read_cursor)?;
+		let short_channel_id = previous_scid + scid_delta.0;
+		previous_scid = short_channel_id;
+
+		let node_id_1_index: BigSize = Readable::read(read_cursor)?;
+		let node_id_2_index: BigSize = Readable::read(read_cursor)?;
+		let node_id_1 = node_ids[node_id_1_index.0 as usize];
```
TheBlueMatt, line 86: These accesses need to be checked or we can panic - should return an error if they're out of bounds - a fuzzer for this should catch that.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,565 @@
...
+		.as_secs() as u32;
+
+	let retrogressed_timestamp = current_timestamp - TIMESTAMP_RETROGRESSION_DELTA; // back-date updates by one week
+
+	let chain_hash: BlockHash = Readable::read(read_cursor)?;
+	let latest_seen_timestamp: u32 = Readable::read(read_cursor)?; //
+	let emulated_last_seen_timestamp = u32::max(latest_seen_timestamp, retrogressed_timestamp);
+
+	let node_id_count: u32 = Readable::read(read_cursor)?;
+	let mut node_ids: Vec<PublicKey> = Vec::with_capacity(node_id_count as usize);
```
TheBlueMatt, line 68: I believe a corrupt file can OOM us here by setting this value to 4G, causing us to allocate 4G*32 bytes. We should bound it by some sensible maximum memory usage (its okay if we have to reallocate once or twice).

lightning/src/routing/network_graph.rs
```diff
@@ -1123,6 +1133,64 @@ impl NetworkGraph {
...
+			node_two: node_2.clone(),
+			two_to_one: None,
+			capacity_sats: None,
+			announcement_message: None,
+			announcement_received_time: timestamp,
+		};
+
+		let mut channels = self.channels.write().unwrap();
+		let nodes = self.nodes.write().unwrap();
+		match channels.entry(short_channel_id) {
```
TheBlueMatt, line 1161: Should we inline this in `add_channel_between_nodes` as well, with the `utxo_value` field from the bottom implementation passed in and always `None` when called from `update_channel_from_partial_announcement`?

lightning/src/routing/network_graph.rs
```diff
@@ -1123,6 +1133,64 @@ impl NetworkGraph {
 		self.update_channel_from_unsigned_announcement_intern(msg, None, chain_access)
 	}
 
+	/// Update channel from partial announcement data received via rapid gossip
+	///
+	/// `timestamp: u64`: Timestamp the announcement was originally received.
+	///
+	/// All other parameters as used in [`msgs::UnsignedChannelAnnouncement`] fields.
+	pub fn update_channel_from_partial_announcement(&self, short_channel_id: u64, timestamp: u64, features: ChannelFeatures, node_id_1: PublicKey, node_id_2: PublicKey) -> Result<(), LightningError> {
```
TheBlueMatt, line 1141: Should this be called something like `add_channel_from_partial_announcement` instead?

arik-so: the other from_announcement method is called update, so I figured I'd leave it consistent, but I don't have any objections either way.

lightning/src/routing/network_graph.rs
arik-so: we might? It will change the signature of add channels between nodes and require it to return a result.

lightning-graph-sync/src/processing.rs
arik-so: Hopefully is now, though I wanna refrain from pushing until my fuzz test's working.

lightning-graph-sync/src/processing.rs
arik-so: true. You figure I should just do a silent cpm::min or have a maximum value and return an error if it's exceeded?

lightning-graph-sync/src/processing.rs
arik-so: It's actually a convenient simplification for the server not having to create a histogram of these values. I guess the server can just return some empty bytes here. Or I could abstract it away into a method that returns a defaults struct? 

But also happy to remove this code, too.

arik-so: Actually, I don't fully agree. The only code we're removing is the presence check, which is two lines. But we still have to read all the defaults after this.

lightning/src/routing/network_graph.rs
arik-so: Actually, no, because the channel tree logic here is different from the one that's required by the other call site in the `update_channel_from_unsigned_announcement_intern` function. We could add a flag to indicate not to remove a channel between nodes, but at that point the functionality gets confusing for the reader.

lightning-graph-sync/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+repository = "https://github.com/lightningdevkit/rust-lightning"
+edition = "2018"
+description = """
+Utility to process gossip routing data from LNSync-like server (protocol name TBD)
+"""
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+_bench_unstable = []
+_test_utils = []
```
arik-so, line 15: Admittedly, I think we can probably drop _test_utils here.

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+#![deny(non_camel_case_types)]
+#![deny(non_snake_case)]
+#![deny(unused_mut)]
+#![deny(unused_variables)]
+#![deny(unused_imports)]
+//! This crate exposes functionality to rapidly sync gossip data, aimed primarily at mobile
+//! devices.
+
+// Allow and import test features for benching
+#![cfg_attr(all(any(test, feature = "_test_utils"), feature = "_bench_unstable"), feature(test))]
```
arik-so, line 14: if we drop _test_utils, this could be simplified to 
`#![cfg_attr(all(test, feature = "_bench_unstable"), feature(test))]`

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,138 @@
+#![deny(missing_docs)]
+#![deny(unsafe_code)]
+#![deny(broken_intra_doc_links)]
+#![deny(non_upper_case_globals)]
+#![deny(non_camel_case_types)]
+#![deny(non_snake_case)]
+#![deny(unused_mut)]
+#![deny(unused_variables)]
+#![deny(unused_imports)]
+//! This crate exposes functionality to rapidly sync gossip data, aimed primarily at mobile
```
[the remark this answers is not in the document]
arik-so: Pretty much all of the readme is wire format info. But I'll come up with some appropriate copy.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,539 @@
...
+			return Err(GraphSyncError::ProcessingError(
+				"Prefix must equal 'LDK' and the version byte 1".to_string(),
+			));
+		}
+	};
+
+	let chain_hash: BlockHash = Readable::read(read_cursor)?;
+	let latest_seen_timestamp: u32 = Readable::read(read_cursor)?; //
+
+	let node_id_count: u32 = Readable::read(read_cursor)?;
```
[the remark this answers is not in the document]
jkczyz: Seems odd we speak generally about the types of top-level fields but don't do the same for fields of inner structs.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,434 @@
...
+		let update_result = update_network_graph(&network_graph, &example_input[..]);
+		assert!(update_result.is_err());
+		if let Err(GraphSyncError::DecodeError(DecodeError::ShortRead)) = update_result {
+			// this is the expected error type
+		} else {
+			assert!(false);
```
jkczyz, line 252: Would it be simple to panic with the `Debug` representation of the unexpected result?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,434 @@
...
+
+	use crate::error::GraphSyncError;
+	use crate::processing::update_network_graph;
+
+	#[test]
+	fn network_graph_fails_to_update_from_clipped_input() {
+		let block_hash = genesis_block(Network::Bitcoin).block_hash();
+		let network_graph = NetworkGraph::new(block_hash);
+
+		// OldestDataDirection0, recent timestamp, with last byte removed
```
jkczyz, line 245: What does `OldestDataDirection0` refer to? Same for similar comments in other tests.

lightning/src/routing/network_graph.rs
```diff
@@ -1132,6 +1142,81 @@ impl NetworkGraph {
...
+			node_two: node_2.clone(),
+			two_to_one: None,
+			capacity_sats: None,
+			announcement_message: None,
+			announcement_received_time: timestamp,
+		};
+
+		let channels = self.channels.write().unwrap();
+		let nodes = self.nodes.write().unwrap();
+		Self::add_channel_between_nodes(channels, nodes, short_channel_id, channel_info, None)
```
jkczyz, line 1170: If we have a `chain::Access`, are we not checking for the UTXO because we are trusting the update from the server?

jkczyz (commented):
Largely looks good. Most of my comments should require much changes.

lightning-graph-sync/src/processing.rs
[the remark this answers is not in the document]
arik-so: Are you referring to the readme, Jeff? I wanted to leave in some top-level types to avoid confusion and because they're not subject to change. Block hash will always be 32 bytes. Timestamp will be 32 bytes until 2038, but by then we'll hopefully have eltoo anyway.

lightning-graph-sync/src/processing.rs
arik-so: This is a comment I left for myself that I use in combination with some experimental variables in the server code to generate these reference updates whenever the format changes. I'd prefer to take out these comments in a follow-up PR.

lightning-graph-sync/src/processing.rs
arik-so: I can implement a Debug for GraphSyncError if that's what you're asking? I haven't so far because it's merely a proxy for other error types, lmk.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+        5. `default_htlc_maximum_msat` (`u64`, and if the default is no maximum, `u64::MAX`)
+    - The defaults are calculated by the server based on the frequency within non-incremental updates within this
+      particular message
+10. `[CustomChannelUpdate]`
+
+You will also notice that `NodeAnnouncement` messages are omitted altogether as the node IDs are implicitly extracted
+from the channel announcements and updates.
+
+The data is then applied to the current network graph, artificially backdated 7 days from the current time to make sure
+more recent updates obtained directly from gossip (if there are any) are not accidentally overwritten.
```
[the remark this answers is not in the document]
arik-so: It's no longer backdated, will update.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+Essentially, the serialization structure is as follows:
+
+1. Fixed prefix bytes `76, 68, 75, 1` (the first three bytes are ASCII for `LDK`)
+    - The purpose of this prefix is to identify the serialization format, should other rapid gossip sync formats arise
+      in the future.
+    - The fourth byte is the protocol version in case our format gets updated
+2. Chain hash (32 bytes)
+3. Latest seen timestamp (`u32`)
+4. An unsigned int indicating the number of node IDs to follow
+5. `[PublicKey]`: array of compressed node ID pubkeys (all pubkeys are presumed to be standard compressed
```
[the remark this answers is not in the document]
arik-so: No idea, but I thought Rust had some readme magic.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+8. An unsigned int indicating the number of channel update messages to follow
+9. A series of default values used for non-incremental channel updates.
+    - The values are defined as follows:
+        1. `default_cltv_expiry_delta`
+        2. `default_htlc_minimum_msat`
+        3. `default_fee_base_msat`
+        4. `default_fee_proportional_millionths`
+        5. `default_htlc_maximum_msat` (`u64`, and if the default is no maximum, `u64::MAX`)
+    - The defaults are calculated by the server based on the frequency within non-incremental updates within this
+      particular message
```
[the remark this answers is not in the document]
arik-so: Will rephrase.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+      in the future.
+    - The fourth byte is the protocol version in case our format gets updated
+2. Chain hash (32 bytes)
+3. Latest seen timestamp (`u32`)
+4. An unsigned int indicating the number of node IDs to follow
+5. `[PublicKey]`: array of compressed node ID pubkeys (all pubkeys are presumed to be standard compressed
+33-byte-serializations)
+6. An unsigned int indicating the number of channel announcement messages to follow
+7. `[CustomChannelAnnouncement]` (array of significantly stripped down channel announcements)
+8. An unsigned int indicating the number of channel update messages to follow
```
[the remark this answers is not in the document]
arik-so: there is no longer a type `CustomChannelAnnouncement`. Given that both announcements and updates are represented in a very makeshift manner, perhaps doing away with backquotes altogether is the way forward.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,93 @@
+# lightning-graph-sync
+
+This crate exposes functionality for rapid gossip graph syncing, aimed primarily at mobile clients.
+
+## Mechanism
+
+The (presumed) server sends a compressed gossip response containing gossip data. The gossip data is formatted compactly,
```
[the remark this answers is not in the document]
arik-so: will do

lightning-graph-sync/src/processing.rs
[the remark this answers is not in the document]
jkczyz: Yeah, though, I'm referring more to the inconsistency of, say, not giving the type for node id count but giving the types for every field within `CustomChannelAnnouncements` and `CustomChannelUpdate` sections. I'm more inclined to give all the types if the `README` is intended to describe the encoding.

I guess it's more a question if we want the encoding in the `README`. I think having it somewhere is valuable, even if in code-level comments. Just seems odd to remove a few types yet leave all the nitty-gritty details of, say, `CustomChannelUpdate`'s layout.

lightning-graph-sync/src/processing.rs
jkczyz: Yeah, just so you can have a better failure message:

```rust
panic!("Unexpected result: {:?}", update_result);
```

lightning-graph-sync/README.md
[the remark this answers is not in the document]
arik-so: Correction, the square brackets are meant to indicate an array.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,431 @@
...
+	} else {
+		OptionalField::Present(tentative_default_htlc_maximum_msat)
+	};
+
+	for _current_update_index in 0..update_count {
+		let scid_delta: BigSize = Readable::read(read_cursor)?;
+		let short_channel_id = previous_scid.checked_add(scid_delta.0).ok_or(ProcessingError("Invalid update SCID delta producing u64 overflow.".to_string()))?;
+		previous_scid = short_channel_id;
+
+		// let short_channel_id: u64 = Readable::read(read_cursor)?;
```
jkczyz, line 124: Remove dead code

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,431 @@
...
+	let tentative_default_htlc_maximum_msat: u64 = Readable::read(&mut read_cursor)?;
+	let default_htlc_maximum_msat = if tentative_default_htlc_maximum_msat == u64::max_value() {
+		OptionalField::Absent
+	} else {
+		OptionalField::Present(tentative_default_htlc_maximum_msat)
+	};
+
+	for _current_update_index in 0..update_count {
+		let scid_delta: BigSize = Readable::read(read_cursor)?;
+		let short_channel_id = previous_scid.checked_add(scid_delta.0).ok_or(ProcessingError("Invalid update SCID delta producing u64 overflow.".to_string()))?;
```
jkczyz, line 121: Could you break by chaining across lines to stay within 100 characters?

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,431 @@
...
+
+		let before = network_graph.to_string();
+		assert_eq!(before.len(), 31);
+
+		let initialization_result = update_network_graph(&network_graph, &initialization_input[..]);
+		assert!(initialization_result.is_ok());
+
+		let single_direction_incremental_update_input = vec![76, 68, 75, 1, 111, 226, 140, 10, 182, 241, 179, 114, 193, 166, 162, 70, 174, 99, 247, 79, 147, 30, 131, 101, 225, 90, 8, 156, 104, 214, 25, 0, 0, 0, 0, 0, 97, 229, 183, 167, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 8, 153, 192, 0, 2, 27, 0, 0, 136, 0, 0, 0, 221, 255, 2, 68, 226, 0, 6, 11, 0, 1, 128];
+		// TODO: use timestamp mocks instead
+		thread::sleep(Duration::from_millis(1000)); // sleep for one second to force a timestamp change
```
jkczyz, line 372: Is this sleep needed? The tests pass without it. I suppose plenty of time has passed since `single_direction_incremental_update_input` was calculated. 😛

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+//! timestamp provided by the client. It's not included in either channel announcement or update,
+//! (not least due to announcements not including any timestamps at all, but only a block height)
+//! but rather, it's a timestamp of when the server saw a particular message.
+//!
+//! If a particular channel update had never occurred before, the full update is sent. If a channel
+//! has had updates prior to the provided timestamp, the latest update prior to the timestamp is
+//! taken as a reference, and the delta is calculated against it.
+//!
+//! In practice, snapshots of server data will be calculated at regular intervals, such that deltas
+//! are cached and needn't be calculated on the fly.
```
jkczyz, line 29: If the client can give any timestamp / block, how would such caching work?

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+
+		fs::create_dir_all("./tmp/graph-sync-tests").unwrap();
+		fs::write("./tmp/graph-sync-tests/test_data.lngossip", valid_response).unwrap();
```
jkczyz, line 78: This leaves a file (and directory) in my repo. Could you clean up and handle failed tests cases similar to what we do in the background processor tests?

https://github.com/lightningdevkit/rust-lightning/blob/d166044f2723ca8602ce654f2c12d4921d0ced9b/lightning-background-processor/src/lib.rs#L419-L427

jkczyz:
Any chance you could use fix-up commits and and not rebasing mid review going forward? Makes it tougher to review otherwise. 🙁

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,550 @@
...
+/// avoid malicious updates being able to trigger excessive memory allocation.
+const MAX_INITIAL_NODE_ID_VECTOR_CAPACITY: u32 = 50_000;
+
+/// Update network graph from binary data
+///
+/// `network_graph`: network graph to be updated
+///
+/// `update_data`: `&[u8]` binary stream that comprises the update data
+pub fn update_network_graph(
+    network_graph: &network_graph::NetworkGraph,
```
jkczyz, line 33: Looks like the tabs in this file was converted to spaces.

lightning-graph-sync/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+pub mod processing;
+
+/// Sync gossip data from a file
+///
+/// `network_graph`: The network graph to apply the updates to
+///
+/// `sync_path`: Path to the file where the gossip update data is located
+///
+pub fn sync_network_graph_with_file_path(
+    network_graph: &network_graph::NetworkGraph,
```
jkczyz, line 56: Likewise here for tabs converted to spaces.

lightning-graph-sync/src/processing.rs
arik-so: Ah I ran rustfmt for idiomatic 100-column wrapping. Will undo.

lightning-graph-sync/src/processing.rs
jkczyz: rustfmt at it again lol

arik-so: oh ffs, this is gonna keep happening

lightning-graph-sync/src/processing.rs
arik-so: They do indeed.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,97 @@
...
+8. An unsigned int indicating the number of channel update messages to follow
+9. A series of default values used for non-incremental channel updates
+    - The values are defined as follows:
+        1. `default_cltv_expiry_delta`
+        2. `default_htlc_minimum_msat`
+        3. `default_fee_base_msat`
+        4. `default_fee_proportional_millionths`
+        5. `default_htlc_maximum_msat` (`u64`, and if the default is no maximum, `u64::MAX`)
+    - The defaults are calculated by the server based on the frequency within non-incremental
+      updates within this particular graph sync stream
```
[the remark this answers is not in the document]
arik-so: I rephrased it to be a bit clearer, I think.

jkczyz (commented):
I'm happy with this now. Could you squash?

arik-so:
Certainly!

jkczyz:
> nit: Please restrict git commit messages to 80-90 chars long. The title should be one line of max 80-90 chars. I generally keep mine even shorter but others let it go a bit longer.

Same. I use 50 characters and vim looks angry if I go over.

lightning-graph-sync/src/processing.rs
```diff
@@ -0,0 +1,549 @@
...
+				OptionalField::Absent
+			} else {
+				OptionalField::Present(tentative_htlc_maximum_msat)
+			};
+		}
+
+		network_graph.update_channel_unsigned(&synthetic_update)?;
+	}
+
+	let input_termination_check: Result<u8, DecodeError> = Readable::read(read_cursor);
```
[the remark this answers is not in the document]
arik-so: Not in this version, no. I think for now it's safer to make sure the entire response can be parsed correctly end to end.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,120 @@
...
+updates are known.
+
+Essentially, the serialization structure is as follows:
+
+1. Fixed prefix bytes `76, 68, 75, 1` (the first three bytes are ASCII for `LDK`)
+    - The purpose of this prefix is to identify the serialization format, should other rapid gossip
+      sync formats arise in the future
+    - The fourth byte is the protocol version in case our format gets updated
+2. Chain hash (32 bytes)
+3. Latest seen timestamp (`u32`)
```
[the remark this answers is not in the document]
arik-so: Different data sets generated at the same time can have different latest seen timestamps. I think keeping it as is does not unnecessarily expand the attack surface, so I'd rather keep it.

lightning-graph-sync/README.md
```diff
@@ -0,0 +1,120 @@
+# lightning-graph-sync
+
+This crate exposes functionality for rapid gossip graph syncing, aimed primarily at mobile clients.
+
+## Mechanism
+
+The (presumed) server sends a compressed gossip response containing gossip data. The gossip data is
```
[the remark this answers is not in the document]
arik-so: good point

lightning-graph-sync/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "lightning-graph-sync"
+version = "0.0.104"
+authors = ["Arik Sosman <git@arik.io>"]
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/lightningdevkit/rust-lightning"
+edition = "2018"
+description = """
+Utility to process gossip routing data from LNSync-like server (protocol name TBD)
```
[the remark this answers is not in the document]
arik-so: I expect there are going to be multiple iterations here. My proposal is `lightning-rapid-gossip-sync` or `lightning-rapid-gossip-sync-client`, though I think `client` is somewhat implied given what repository it's in.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+//! (not least due to announcements not including any timestamps at all, but only a block height)
+//! but rather, it's a timestamp of when the server saw a particular message.
+//!
+//! If a particular channel update had never occurred before, the full update is sent. If a channel
+//! has had updates prior to the provided timestamp, the latest update prior to the timestamp is
+//! taken as a reference, and the delta is calculated against it.
+//!
+//! In practice, snapshots of server data will be calculated at regular intervals, such that deltas
+//! are cached and needn't be calculated on the fly. All the client needs to provide to the server
+//! is the timestamp included in the previously received (and hence snapshotted) delta.
```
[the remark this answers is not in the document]
arik-so: There are docs in the server.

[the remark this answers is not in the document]
arik-so: Well, the client does link to the server, which does have comments in the readme showing the request paths. I'm happy to copy the request paths section here, but given that every server instance is gonna use a different API prefix, not to mention domain, I don't know what else to say here.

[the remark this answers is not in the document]
arik-so: Once we have the example service up, sure. Should we just say that that's gonna be an example URL in the future?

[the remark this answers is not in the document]
arik-so: No, I can't commit to responding to all the server comments within minutes, some are a bit painful to addrress.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -57,7 +75,7 @@ pub mod processing;
 pub fn sync_network_graph_with_file_path(
 	network_graph: &network_graph::NetworkGraph,
 	sync_path: &str,
-) -> Result<(u32), GraphSyncError> {
+) -> Result<u32, GraphSyncError> {
```
jkczyz, line 78: Document return type here, too.

jkczyz (approved):

PR #1178, opened 2021-11-18
Generate PaymentPathSuccessful event for each path

jkczyz:
A single `PaymentSent` event is generated when a payment is fulfilled. This is occurs when the preimage is revealed on the first claimed HTLC. For subsequent HTLCs, the event is not generated.

In order to score channels involved with a successful payments, the scorer must be notified of each successful path involved in the payment. Add a `PaymentPathSuccessful` event for this purpose. Generate it whenever a part is removed from a pending outbound payment. This avoids duplicate events when reconnecting to a peer.

### DIFF
--- a/lightning/src/ln/channelmanager.rs
+++ b/lightning/src/ln/channelmanager.rs
@@ -449,6 +449,7 @@ pub(crate) enum PendingOutboundPayment {
 	/// and add a pending payment that was already fulfilled.
 	Fulfilled {
 		session_privs: HashSet<[u8; 32]>,
+		payment_hash: Option<PaymentHash>,
 	},
 }
 
@@ -472,15 +473,24 @@ impl PendingOutboundPayment {
 		}
 	}
 
+	fn payment_hash(&self) -> Option<PaymentHash> {
+		match self {
+			PendingOutboundPayment::Legacy { .. } => None,
+			PendingOutboundPayment::Retryable { payment_hash, .. } => Some(*payment_hash),
+			PendingOutboundPayment::Fulfilled { payment_hash, .. } => *payment_hash,
+		}
+	}
+
 	fn mark_fulfilled(&mut self) {
 		let mut session_privs = HashSet::new();
 		core::mem::swap(&mut session_privs, match self {
 			PendingOutboundPayment::Legacy { session_privs } |
 			PendingOutboundPayment::Retryable { session_privs, .. } |
-			PendingOutboundPayment::Fulfilled { session_privs }
+			PendingOutboundPayment::Fulfilled { session_privs, .. }
 				=> session_privs
 		});
-		*self = PendingOutboundPayment::Fulfilled { session_privs };
+		let payment_hash = self.payment_hash();
+		*self = PendingOutboundPayment::Fulfilled { session_privs, payment_hash };
 	}
 
 	/// panics if path is None and !self.is_fulfilled
@@ -488,7 +498,7 @@ impl PendingOutboundPayment {
 		let remove_res = match self {
 			PendingOutboundPayment::Legacy { session_privs } |
 			PendingOutboundPayment::Retryable { session_privs, .. } |
-			PendingOutboundPayment::Fulfilled { session_privs } => {
+			PendingOutboundPayment::Fulfilled { session_privs, .. } => {
 				session_privs.remove(session_priv)
 			}
 		};
@@ -529,7 +539,7 @@ impl PendingOutboundPayment {
 		match self {
 			PendingOutboundPayment::Legacy { session_privs } |
 			PendingOutboundPayment::Retryable { session_privs, .. } |
-			PendingOutboundPayment::Fulfilled { session_privs } => {
+			PendingOutboundPayment::Fulfilled { session_privs, .. } => {
 				session_privs.len()
 			}
 		}
@@ -3491,14 +3501,23 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 	}
 
 	fn finalize_claims(&self, mut sources: Vec<HTLCSource>) {
+		let mut pending_events = self.pending_events.lock().unwrap();
 		for source in sources.drain(..) {
-			if let HTLCSource::OutboundRoute { session_priv, payment_id, .. } = source {
+			if let HTLCSource::OutboundRoute { session_priv, payment_id, path, .. } = source {
 				let mut session_priv_bytes = [0; 32];
 				session_priv_bytes.copy_from_slice(&session_priv[..]);
 				let mut outbounds = self.pending_outbound_payments.lock().unwrap();
 				if let hash_map::Entry::Occupied(mut payment) = outbounds.entry(payment_id) {
 					assert!(payment.get().is_fulfilled());
-					payment.get_mut().remove(&session_priv_bytes, None);
+					if payment.get_mut().remove(&session_priv_bytes, None) {
+						pending_events.push(
+							events::Event::PaymentPathSuccessful {
+								payment_id,
+								payment_hash: payment.get().payment_hash(),
+								path,
+							}
+						);
+					}
 					if payment.get().remaining_parts() == 0 {
 						payment.remove();
 					}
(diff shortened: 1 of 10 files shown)


lightning/src/ln/functional_test_utils.rs
```diff
@@ -1098,6 +1115,26 @@ macro_rules! expect_payment_sent {
 			},
 			_ => panic!("Unexpected event"),
 		}
+		if $expect_paths {
+			for i in 1..events.len() {
+				match events[i] {
+					Event::PaymentPathSuccessful { .. } => {},
```
TheBlueMatt, line 1121: Does it make sense to check the payment_id here matches the paymentsent event?

lightning/src/util/events.rs
```diff
@@ -338,7 +338,20 @@ pub enum Event {
...
+	/// Indicates that a path for an outbound payment was successful.
+	///
+	/// Useful for scoring channels. See [`Event::PaymentSent`] for obtaining the payment preimage.
+	PaymentPathSuccessful {
+		/// The id returned by [`ChannelManager::send_payment`] and used with
+		/// [`ChannelManager::retry_payment`].
+		///
+		/// [`ChannelManager::send_payment`]: crate::ln::channelmanager::ChannelManager::send_payment
+		/// [`ChannelManager::retry_payment`]: crate::ln::channelmanager::ChannelManager::retry_payment
+		payment_id: PaymentId,
```
TheBlueMatt, line 351: We should probably add the `payment_hash` here, no?

jkczyz: Yeah, I think so. Though when producing an event in `finalize_claims` we don't have the payment hash. I guess we could pull it from `PendingOutboundPayment` but we wouldn't have it for `Legacy` or `Fulfilled`. Or we could include it in `MonitorRestoreUpdates::finalized_claimed_htlcs`, but I think that would require changing `Channel::monitor_pending_finalized_fulfills`.

Seems we'd have to make it optional in `PaymentPathSuccessful` and change the serialization format for either `Channel` or `PendingOutboundPayment`. Let me know if I'm missing something, though.

TheBlueMatt: > I guess we could pull it from PendingOutboundPayment but we wouldn't have it for Legacy or Fulfilled.

That seems fine IMO. We'll want to drop support for those legacy versions probably sooner rather than later, just given the amount of overhead they create.

lightning/src/ln/functional_test_utils.rs
jkczyz: Indeed. Also added for `payment_hash`, though that resulted in a failure in `ln::onion_route_tests::test_onion_failure`. I hacked in a workaround, but I'm wondering if there is a better way to make the tests less fragile / dependent on previous parts. Making them separate tests is probably the obvious way.

TheBlueMatt:
Need to add the new event to the `chanmon_consistency` list of valid events:

```diff
diff --git a/fuzz/src/chanmon_consistency.rs b/fuzz/src/chanmon_consistency.rs
index acdf3cb3c..3dd0cd5a4 100644
--- a/fuzz/src/chanmon_consistency.rs
+++ b/fuzz/src/chanmon_consistency.rs
@@ -829,6 +829,7 @@ pub fn do_test<Out: test_logger::Output>(data: &[u8], out: Out) {
                                                        }
                                                },
                                                events::Event::PaymentSent { .. } => {},
+                                               events::Event::PaymentPathSuccessful { .. } => {},
                                                events::Event::PaymentPathFailed { .. } => {},
                                                events::Event::PaymentForwarded { .. } if $node == 1 => {},
                                                events::Event::PendingHTLCsForwardable { .. } => {

```

lightning/src/ln/channelmanager.rs
```diff
@@ -3515,32 +3534,43 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
...
 						// irrevocably fulfilled.
-						payment.get_mut().remove(&session_priv_bytes, Some(&path));
+						if payment.get_mut().remove(&session_priv_bytes, Some(&path)) {
+							let payment_hash = Some(PaymentHash(Sha256::hash(&payment_preimage.0).into_inner()));
+							pending_events.push(
+								events::Event::PaymentPathSuccessful {
+									payment_id,
+									payment_hash,
+									path,
+								}
```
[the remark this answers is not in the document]
TheBlueMatt: Hmmm, its definitely success, though. We could add a flag for it, but because routing is generally done via the `list_usable_channels` list we won't consider the now-closed channel anyway.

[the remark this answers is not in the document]
jkczyz: Could this also result from another channel along the path?

From a scoring perspective, I'd imagine we'd want to score positively any channel that was used to route the HTLC to the payee. And any closed channels should be avoided through other means.

lightning/src/ln/onion_route_tests.rs
```diff
@@ -477,11 +477,19 @@ fn test_onion_failure() {
 	bogus_route.paths[0][route_len-1].fee_msat = amt_to_forward;
 	run_onion_failure_test("amount_below_minimum", 0, &nodes, &bogus_route, &payment_hash, &payment_secret, |_| {}, ||{}, true, Some(UPDATE|11), Some(NetworkUpdate::ChannelUpdateMessage{msg: ChannelUpdate::dummy(short_channel_id)}), Some(short_channel_id));
 
+	for node in nodes.iter() {
+		*node.keys_manager.override_session_priv.lock().unwrap() = Some([4; 32]);
+	}
```
valentinewallace, line 482: Seems pre-existing but as you mention these parts are a bit weird/could use comments at some point

lightning/src/util/events.rs
```diff
@@ -600,6 +627,20 @@ impl MaybeReadable for Event {
...
 			},
+			13u8 => {
+				let f = || {
+					let mut payment_id = PaymentId([0; 32]);
+					let mut payment_hash = None;
+					let mut path = VecReadWrapper(Vec::new());
+					read_tlv_fields!(reader, {
+						(0, payment_id, required),
+						(2, payment_hash, option),
+						(4, path, required),
```
valentinewallace, line 638: I think using `vec_type` would clean up the serialization a bit:

```
index 77ad9f75..f47e4793 100644
@@ -462,7 +462,7 @@ impl Writeable for Event {
                                write_tlv_fields!(writer, {
                                        (0, payment_id, required),
                                        (2, payment_hash, option),
-                                       (4, VecWriteWrapper(path), required)
+                                       (4, path, vec_type)
                                })
                        },
                        // Note that, going forward, all new events must only write data inside of
@@ -631,13 +631,13 @@ impl MaybeReadable for Event {
                                let f = || {
                                        let mut payment_id = PaymentId([0; 32]);
                                        let mut payment_hash = None;
-                                       let mut path = VecReadWrapper(Vec::new());
+                                       let mut path = None;
                                        read_tlv_fields!(reader, {
                                                (0, payment_id, required),
                                                (2, payment_hash, option),
-                                               (4, path, required),
+                                               (4, path, vec_type),
                                        });
-                                       Ok(Some(Event::PaymentPathSuccessful { payment_id, payment_hash, path: path.0 } ))
+                                       Ok(Some(Event::PaymentPathSuccessful { payment_id, payment_hash, path: path.unwrap() } ))
                                };
                                f()
                        },
```

valentinewallace (commented):
I'm ACK mod one serialization suggestion

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
valentinewallace: I believe this can only result from one of our own channel closes. Could document that `path` may contain a closed channel in rare cases.

lightning/src/ln/onion_route_tests.rs
jkczyz: Yeah, agreed! Mostly looking for a better way to do this rather than the hack here. The problem is the previous test cases use the same `payment_id` because of the earlier override. So we end up getting a pending outbound payment with `payment_hash` rather than the one generated inside `send_along_route`. Maybe there's a way to write these tests without needing the override?

lightning/src/util/events.rs
jkczyz: Ah, good point!

lightning/src/ln/channelmanager.rs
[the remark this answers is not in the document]
TheBlueMatt: Yea, this specific case, but the general case of "we got an accept because of an on-chain claim" is also an issue (and hits the other case). I don't think we can detect the difference there (aside from the HTLC fulfill being really slow), so I'm not sure there's anything we can do in the general case.

valentinewallace (approved):

lightning/src/util/events.rs
```diff
@@ -338,7 +340,26 @@ pub enum Event {
 		channel_id: [u8; 32],
 		/// The full transaction received from the user
 		transaction: Transaction
-	}
+	},
+	/// Indicates that a path for an outbound payment was successful.
```
TheBlueMatt, line 344: Maybe mention that this is always generated after `PaymentSuccessfull`? I can otherwise read this as something that will be generated when one part of an MPP payment reaches the target even if other parts haven't yet (I know the protocol doesn't support doing that, but still).

TheBlueMatt (approved):
Can you squash the fixup commits?

lightning/src/ln/onion_route_tests.rs
jkczyz: FYI, I ended up adding a test-only `clear_pending_payments` method to call instead of using this hack.

valentinewallace (approved):

PR #1234, opened 2022-01-11
Limit maximum CLTV delta during routing.

tnull:
This PR attempts to address #996. As this is my first contact with the 950+ lines of `get_route`, I hope that I didn't mess up too much. I also fixed some typos I found while going through the code.

Three remarks though: 

1. Currently a lot of unit tests fail. This seems to be the case since many of the supplied `cltv_expiry_delta` values in the dummy graph data are chosen very high. If all agree that the failures do not come from something else, I could change them to some sane(r)(?) values, even though I'm not entirely sure yet what they would be.
2. How should we choose the `MAX_TOTAL_CLTV_DELTA` limit? The value of 1000 seems like a good starting point, however, a decent number of nodes use values that are close enough or higher than that (see [attached data](https://github.com/lightningdevkit/rust-lightning/files/7848527/cltv_delta_counts.txt) I took from a recent graph snapshot). These would be excluded from any potential paths, which could impair connectivity and payment success. 
3. In order to be minimal invasive I currently set this `MAX_TOTAL_CLTV_DELTA` limit inline. But maybe this parameter should be configurable, i.e., by defining it as part of `RouteParameters` or a new `RoutingConfig` in `config.rs`?


lightning/src/routing/router.rs
```diff
@@ -784,6 +785,13 @@ where L::Target: Logger {
 					// Verify the liquidity offered by this channel complies to the minimal contribution.
 					let contributes_sufficient_value = available_value_contribution_msat >= minimal_value_contribution_msat;
 
+
+					// Do not consider candidates that exceed the MAX_TOTAL_CLTV_DELTA limit.
+					const MAX_TOTAL_CLTV_DELTA: u32 = 1000;
+					let hop_total_cltv_delta = $next_hops_cltv_delta + $directional_info.cltv_expiry_delta as u32;
+					let doesnt_exceed_cltv_delta_limit = hop_total_cltv_delta <= MAX_TOTAL_CLTV_DELTA;
+					log_trace!(logger, "Checking total CLTV delta: {} / {}. doesnt_exceed_cltv_delta_limit: {}", hop_total_cltv_delta, MAX_TOTAL_CLTV_DELTA, doesnt_exceed_cltv_delta_limit);
```
TheBlueMatt, line 793: Hmm, we currently don't log *this* verbosely even at TRACE. Honestly we should probablve move that direciton, but for now let's not make TRACE fill GBs if you're calculating routes.

lightning/src/routing/router.rs
```diff
@@ -784,6 +785,13 @@ where L::Target: Logger {
 					// Verify the liquidity offered by this channel complies to the minimal contribution.
 					let contributes_sufficient_value = available_value_contribution_msat >= minimal_value_contribution_msat;
 
+
+					// Do not consider candidates that exceed the MAX_TOTAL_CLTV_DELTA limit.
+					const MAX_TOTAL_CLTV_DELTA: u32 = 1000;
```
TheBlueMatt, line 790: 1008 is a nice-round-week, but maybe lets make it configurable. I could totally use-cases for lightning that need longer timeouts, and also people not want to wait a week. I guess we can shove it through the `Payee` thing?

TheBlueMatt (commented):
This looks great! Some minor comments and could use a test to demonstrate it works, but looks great.

lightning/src/routing/router.rs
tnull: I totally agree. I just left that in there for now for debugging purposes. Will remove it before eventually squashing.

lightning/src/routing/router.rs
tnull: Great! But it seem to me that the limit doesn't have anything to do with the `Payee` side of things, but is chosen/enforced by the payer? So probably `RouteParameters` or a user config option makes more sense?

TheBlueMatt: Oh, right, forgot we had a `RouteParameters` struct, yea, put it there :)

TheBlueMatt: Also, instead of changing every test, if we make it configurable we could just disable it for testing, not that it matters a ton either way.

TheBlueMatt:
Basically looking good, needs to be configurable and needs a test for the max-expiry behavior specifically.

lightning/src/routing/router.rs
tnull: I now added a `max_total_cltv_expiry_delta` field to `RouteParameters` in order to make the limit configurable and handed it down to `get_route` in the same way that the other params are provided (see https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/beb91f7f8093fa672851d03908ea92e0dcfb6e5f). However, this required to touch a lot of places, most of them being tests. Not sure if a global user config would not be the cleaner way to it.

lightning/src/ln/channelmanager.rs
```diff
@@ -1119,6 +1119,9 @@ pub(super) const CLTV_FAR_FAR_AWAY: u32 = 6 * 24 * 7; //TODO?
 // a payment was being routed, so we add an extra block to be safe.
 pub const MIN_FINAL_CLTV_EXPIRY: u32 = HTLC_FAIL_BACK_BUFFER + 3;
 
+/// Maximum total CTLV difference we allow for a full payment path
+pub const MAX_TOTAL_CLTV_EXPIRY_DELTA: u32 = 1000;
```
TheBlueMatt, line 1123: This likely belongs in the `router` module, IMO. It's not specific to our channel logic or even defined in the spec, its just a limit on the router. Also re: naming: maybe call it "SUGGESTED" or "DEFAULT"?

lightning/src/routing/router.rs
TheBlueMatt: Yea, the `get_route` is a bit of a compatibility wrapper, we could switch to using the public interface in tests, but the test diff isn't all that bad right now IMO so no big deal.

tnull: Great!

lightning/src/ln/channelmanager.rs
tnull: Alright, moved and renamed the limit with https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/54a98eec3232ad4180ac17dc433316729d484159.

tnull:
> Basically looking good, needs to be configurable and needs a test for the max-expiry behavior specifically.

I now added a simple test against the default value and limit of 0 with https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/0c93422b77a2bdd6c6e21178c1c4738f9081659b.
Let me know if you think more in-depth testing is desirable.

lightning/src/ln/channelmanager.rs
```diff
@@ -2651,6 +2651,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 							payee: payee.clone(),
 							final_value_msat: pending_amt_unsent,
 							final_cltv_expiry_delta: max_unsent_cltv_delta,
+							max_total_cltv_expiry_delta: DEFAULT_MAX_TOTAL_CLTV_DELTA,
```
jkczyz, line 2654: Should this be the value used in the `RouteParameters` when finding the `Route`? Similarly elsewhere?

lightning/src/ln/functional_tests.rs
```diff
@@ -7738,11 +7738,11 @@ fn test_bump_penalty_txn_on_revoked_htlcs() {
 	let payee = Payee::from_node_id(nodes[1].node.get_our_node_id()).with_features(InvoiceFeatures::known());
 	let scorer = test_utils::TestScorer::with_fixed_penalty(0);
 	let route = get_route(&nodes[0].node.get_our_node_id(), &payee, &nodes[0].network_graph, None,
-		3_000_000, 50, nodes[0].logger, &scorer).unwrap();
+		3_000_000, 50, 1000,  nodes[0].logger, &scorer).unwrap();
```
jkczyz, line 7741: nit: remove extra space

lightning/src/routing/router.rs
```diff
@@ -165,14 +165,22 @@ pub struct RouteParameters {
...
 	(0, payee, required),
 	(2, final_value_msat, required),
 	(4, final_cltv_expiry_delta, required),
+	(6, max_total_cltv_expiry_delta, required),
 });
 
+/// Maximum total CTLV difference we allow for a full payment path
+pub const DEFAULT_MAX_TOTAL_CLTV_DELTA: u32 = 1000;
+
+
```
jkczyz, line 183: nit: remove extra line

lightning/src/routing/router.rs
```diff
@@ -165,14 +165,22 @@ pub struct RouteParameters {
...
 
 impl_writeable_tlv_based!(RouteParameters, {
 	(0, payee, required),
 	(2, final_value_msat, required),
 	(4, final_cltv_expiry_delta, required),
+	(6, max_total_cltv_expiry_delta, required),
 });
 
+/// Maximum total CTLV difference we allow for a full payment path
+pub const DEFAULT_MAX_TOTAL_CLTV_DELTA: u32 = 1000;
```
jkczyz, line 181: `DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA`?

lightning/src/routing/router.rs
```diff
@@ -536,13 +545,13 @@ pub fn find_route<L: Deref, S: Score>(
...
 		our_node_pubkey, &params.payee, network, first_hops, params.final_value_msat,
-		params.final_cltv_expiry_delta, logger, scorer
+		params.final_cltv_expiry_delta, params.max_total_cltv_expiry_delta, logger, scorer
 	)
 }
 
 pub(crate) fn get_route<L: Deref, S: Score>(
 	our_node_pubkey: &PublicKey, payee: &Payee, network: &NetworkGraph,
-	first_hops: Option<&[&ChannelDetails]>, final_value_msat: u64, final_cltv_expiry_delta: u32,
+	first_hops: Option<&[&ChannelDetails]>, final_value_msat: u64, final_cltv_expiry_delta: u32, max_total_cltv_expiry_delta: u32,
```
jkczyz, line 554: nit: move `max_total_cltv_expiry_delta` to following line to keep with our (admittedly loose) 100-char per line limit.

lightning/src/routing/router.rs
```diff
@@ -721,8 +730,8 @@ where L::Target: Logger {
 		// $next_hops_fee_msat represents the fees paid for using all the channels *after* this one,
 		// since that value has to be transferred over this channel.
 		// Returns whether this channel caused an update to `targets`.
-		( $chan_id: expr, $src_node_id: expr, $dest_node_id: expr, $directional_info: expr, $capacity_sats: expr, $chan_features: expr, $next_hops_fee_msat: expr,
-		   $next_hops_value_contribution: expr, $next_hops_path_htlc_minimum_msat: expr, $next_hops_path_penalty_msat: expr ) => { {
+		( $chan_id: expr, $src_node_id: expr, $dest_node_id: expr, $directional_info: expr, $capacity_sats: expr, $chan_features: expr, $max_total_cltv_expiry_delta: expr, $next_hops_fee_msat: expr,
```
jkczyz, line 733: I believe you don't need `$max_total_cltv_expiry_delta` since it doesn't change.

lightning/src/routing/router.rs
```diff
@@ -990,7 +1005,7 @@ where L::Target: Logger {
 	// meaning how much will be paid in fees after this node (to the best of our knowledge).
 	// This data can later be helpful to optimize routing (pay lower fees).
 	macro_rules! add_entries_to_cheapest_to_target_node {
-		( $node: expr, $node_id: expr, $fee_to_target_msat: expr, $next_hops_value_contribution: expr, $next_hops_path_htlc_minimum_msat: expr, $next_hops_path_penalty_msat: expr ) => {
+		( $node: expr, $node_id: expr, $fee_to_target_msat: expr, $max_total_cltv_expiry_delta: expr, $next_hops_value_contribution: expr, $next_hops_path_htlc_minimum_msat: expr, $next_hops_path_penalty_msat: expr, $next_hops_cltv_delta: expr ) => {
```
jkczyz, line 1008: Likewise here.

lightning/src/routing/router.rs
```diff
@@ -165,14 +165,22 @@ pub struct RouteParameters {
...
+
+	/// The maximum total CLTV delta we accept for the route.
+	pub max_total_cltv_expiry_delta: u32,
 }
 
 impl_writeable_tlv_based!(RouteParameters, {
 	(0, payee, required),
 	(2, final_value_msat, required),
 	(4, final_cltv_expiry_delta, required),
+	(6, max_total_cltv_expiry_delta, required),
```
jkczyz, line 177: I think we'll want to make this an odd type (`1`). But we can use `(default_value, DEFAULT_MAX_TOTAL_CLTV_DELTA)` instead of `required`.

jkczyz (commented):
Looks pretty good so far. Thanks for putting this together!

lightning/src/ln/functional_tests.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/b7cb0f3eca8f491de09c97e41b8f090c3d27264e.

lightning/src/routing/router.rs
tnull: Yeah, I hesitated to go there since that is one long variable name. But you're right, for consistencies sake it's likely better. Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/b7cb0f3eca8f491de09c97e41b8f090c3d27264e.

lightning/src/routing/router.rs
tnull: But it should be configurable, i.e. not always `DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA`, even thought this is currently only used for testing purposes?

lightning/src/routing/router.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/b7cb0f3eca8f491de09c97e41b8f090c3d27264e.

lightning/src/routing/router.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/b7cb0f3eca8f491de09c97e41b8f090c3d27264e.

lightning/src/ln/channelmanager.rs
tnull: I'm not sure what the question is? Whether the default value of 1000 is always applicable?

lightning/src/routing/router.rs
tnull: Ah, sorry, you right of course. Addressed in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/cb7269de6a7ffa38aa7cdd222ab699c97e6eac8a.

lightning/src/routing/router.rs
tnull: Addressed in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/cb7269de6a7ffa38aa7cdd222ab699c97e6eac8a.

lightning/src/ln/channelmanager.rs
tnull: Yes, I stumbled also across that but refrained from too invasive changes. The issue is that the `RouteParameters` for the failed route are not available in `send_payment`/`send_payment_internal` etc. 
Maybe they should be  included in the `Route` struct and supplied fully to `get_route`? This way, every method calling `find_route`/`get_route` would have access to the parameters of the (failed) route?

lightning/src/routing/router.rs
```diff
@@ -4801,6 +4820,31 @@ mod tests {
...
+		let scorer = test_utils::TestScorer::with_fixed_penalty(0);
+
+		// Make sure that generally there is at least one route available
+		let feasible_max_total_cltv_delta = 1000;
+		let route = get_route(&our_id, &payee, &network_graph, None, 100, 42, feasible_max_total_cltv_delta, Arc::clone(&logger), &scorer).unwrap();
+		let path = route.paths[0].iter().map(|hop| hop.short_channel_id).collect::<Vec<_>>();
+		assert_ne!(path.len(), 0);
+
+		// But not if we exclude all paths on the basis of their accumulated CLTV delta
+		let fail_max_total_cltv_delta = 0;
```
TheBlueMatt, line 4838: Maybe set this to something non-0? Not that it matters much.

TheBlueMatt (commented):
LGTM. A few super-trivial comments that might as well be addressed when squashing, but I'd let @jkczyz follow up first.

lightning/src/ln/channelmanager.rs
jkczyz: Yeah, I was thinking something like that. It would work here, but in the places where we form `PaymentPathFailed` events, we'll need it as part of `HTLCSource::OutboundRoute` or maybe have it directly in the path somehow? The latter is currently a `Vec` though. @TheBlueMatt Not sure if you have an opinion here. Seems it would need to be added to both `Route` and `HTLCSource::OutboundRoute`.

TheBlueMatt: Ah, hmmmm, I wonder if we should really move it back to `Payee`? Then we'd get this for free. It is kinda awkward in `RouteParameters` here, as we'd really want the payment retryer to use some user-configured value. Otherwise, yea, I think we'd want to put it in the `Route`.

lightning/src/routing/router.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/be78fa3e2ae094f67bf80b9ecdfc93a388b7cd5d.

lightning/src/routing/router.rs
```diff
@@ -165,14 +165,21 @@ pub struct RouteParameters {
...
 
 impl_writeable_tlv_based!(RouteParameters, {
 	(0, payee, required),
+	(1, max_total_cltv_expiry_delta, (default_value, DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA)),
 	(2, final_value_msat, required),
 	(4, final_cltv_expiry_delta, required),
 });
 
+/// Maximum total CTLV difference we allow for a full payment path
+pub const DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA: u32 = 1000;
```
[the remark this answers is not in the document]
tnull: Alright. I changed it in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/be78fa3e2ae094f67bf80b9ecdfc93a388b7cd5d and also adjusted and renamed the corresponding testing variable for the sake of consistency.

lightning/src/routing/router.rs
```diff
@@ -802,10 +815,10 @@ where L::Target: Logger {
 					// as not sufficient.
 					if !over_path_minimum_msat {
```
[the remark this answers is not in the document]
tnull: Okay, I changed it to `!over_path_minimum_msat && doesnt_exceed_cltv_delta_limit` with https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/be78fa3e2ae094f67bf80b9ecdfc93a388b7cd5d.

lightning/src/ln/channelmanager.rs
tnull: Hum, so while I see that putting it in `Payee` would be a quick fix, I'd think this could be pretty confusing, especially in the long term (since the limit itself can be either thought of to be a property of the route or the payer, but not the payee). 

So putting `RouteParameters` or directly `max_total_cltv_expiry_delta` in `Route` and `OutboundRoute` would make a bit more sense to me, even though it is also kind of redundant to have the same parameters in multiple structs. This is however already the case with these two structs and other variables.

An alternative could maybe be to supply the `max_total_cltv_expiry_delta` to `find_route`/`get_route` etc. and store the used value on a per-payment basis in `PendingOutboundPayment`?

TheBlueMatt: We could also take the view that `Payee` is simply different from what it says on the tin, and seek to rename it. The way the code is structured, absent any names, there's no question this should go in `Payee` - its used to communicate route parameters which can be used to calculate a new route to retry a payment. `RouteParameters` is info specific to the specific route. Really `final_cltv_expiry_delta` should be in the `Payee`, it only isn't because it can be trivially recalculated from the route itself.

I think I'd somewhat favor putting it in `Payee`, updating the docs, land this, and take renaming `Payee` as a followup. What do you think @jkczyz?

jkczyz: > Hum, so while I see that putting it in `Payee` would be a quick fix, I'd think this could be pretty confusing, especially in the long term (since the limit itself can be either thought of to be a property of the route or the payer, but not the payee).
> 
> So putting `RouteParameters` or directly `max_total_cltv_expiry_delta` in `Route` and `OutboundRoute` would make a bit more sense to me, even though it is also kind of redundant to have the same parameters in multiple structs. This is however already the case with these two structs and other variables.
> 

I'd support putting `RouteParameters` in both `Route` and `OutboundRoute`, but for the latter I think we'd want to update the `final_value_msat` and `final_cltv_expiry_delta` values to reflect that of the path in question. Then we'd deprecate the `OutboundRoute::payee` field since it would be redundant with the one in `RouteParameters`.

Though, I guess this is a little redundant since `final_value_msat` and `final_cltv_expiry_delta` could be learned from the path.

> An alternative could maybe be to supply the `max_total_cltv_expiry_delta` to `find_route`/`get_route` etc. as a parameter and store the used value on a per-payment basis in `PendingOutboundPayment::Retryable`?

I guess this means the user would have to look it up in order to call `find_route` for a retry? We'd have to add a method for this to `ChannelManager` and `Payer` (trait in `lightning-invoice` crate). I'm more inclined to take the other approach, but could be convinced otherwise.

TheBlueMatt: > Yeah, it's starting to sink in that this may be more ideal. Would s/Payee/PaymentParameters and s/pubkey/payee_pubkey work for you?
> 
> We may event want to re-consider if RotueParameters is even necessary if final_cltv_expiry_delta is also moved. That just leaves the amount which can be recomputed from the path, IIUC.

Yea, both of those seem reasonable to me. I'm happy to drop the amount, too, but given its already there and a nice struct to contain "amount + parameters" I don't see why remove it.

tnull: Alright, moving `max_total_cltv_expiry_delta` to `Payee` seems like a good solution, especially since renaming it to something along the line of `PaymentParameters` solves my issues with it. I now moved the value it with https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/7fe7a763ad7e1df311b17bd991d596729ba928a4 and will open a follow-up PR for the renaming as soon as this is merged.

tnull:
Rebased and squashed now, should be ready to merge.

lightning/src/ln/channelmanager.rs
```diff
@@ -6903,7 +6903,8 @@ mod tests {
 
 		// Next, attempt a keysend payment and make sure it fails.
 		let params = RouteParameters {
-			payee: Payee::for_keysend(expected_route.last().unwrap().node.get_our_node_id()),
+			payee: Payee::for_keysend(expected_route.last().unwrap().node.get_our_node_id())
+				.with_max_total_cltv_expiry_delta(TEST_MAX_TOTAL_CLTV_EXPIRY_DELTA),
```
TheBlueMatt, line 6907: Since its now optional, why modify the tests at all?

TheBlueMatt: Ah, I see its always enforced, nevermind.

lightning/src/routing/router.rs
```diff
@@ -173,6 +173,9 @@ impl_writeable_tlv_based!(RouteParameters, {
 	(4, final_cltv_expiry_delta, required),
 });
 
+/// Maximum total CTLV difference we allow for a full payment path
```
jkczyz, line 176: nit: end doc with period.

lightning/src/routing/router.rs
```diff
@@ -542,7 +558,7 @@ where L::Target: Logger {
 
 pub(crate) fn get_route<L: Deref, S: Score>(
 	our_node_pubkey: &PublicKey, payee: &Payee, network: &NetworkGraph,
-	first_hops: Option<&[&ChannelDetails]>, final_value_msat: u64, final_cltv_expiry_delta: u32,
+	first_hops: Option<&[&ChannelDetails]>, final_value_msat: u64, final_cltv_expiry_delta: u32,
```
jkczyz, line 561: nit: remove trailing space

lightning/src/routing/router.rs
```diff
@@ -575,7 +591,7 @@ where L::Target: Logger {
 	//    any ~sufficient (described later) value.
 	//    If succeed, remember which channels were used and how much liquidity they have available,
 	//    so that future paths don't rely on the same liquidity.
-	// 3. Prooceed to the next step if:
```
jkczyz, line 578: If it's not much of a problem, could you keep the initial clean-up commit separate rather than squashing it with the others? May be able to recover it with `git reflog`.

jkczyz (commented):
Yeah, looks good module some nits.

lightning/src/routing/router.rs
```diff
@@ -192,10 +195,14 @@ pub struct Payee {
...
 	/// Expiration of a payment to the payee, in seconds relative to the UNIX epoch.
 	pub expiry_time: Option<u64>,
+
+	/// The maximum total CLTV delta we accept for the route.
+	pub max_total_cltv_expiry_delta: Option<u32>,
 }
 
 impl_writeable_tlv_based!(Payee, {
 	(0, pubkey, required),
+	(1, max_total_cltv_expiry_delta, option),
```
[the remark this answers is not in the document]
tnull: Hum, I have no strong opinion on this, but making it non-optional would  probably require that `max_total_cltv_expiry_delta` is explicitly given with each call of `from_node_id(..)` and `for_keysend(..)`, right? 
While implementing the changes I found that making it optional and using the default value when no other value is set would be way cleaner, especially it would not require to know about, import,  and use `DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA` all over the place.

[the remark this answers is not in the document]
TheBlueMatt: > Hum, I have no strong opinion on this, but making it non-optional would probably require that max_total_cltv_expiry_delta is explicitly given with each call of from_node_id(..) and for_keysend(..), right?

I'd think we'd just set it to the default value - that is, after all, the implicit behavior in the router - it uses the default value if none is set.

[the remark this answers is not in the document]
TheBlueMatt: Before I'd scrolled to that point in the code, when I saw the value was an `Option`, my assumption was that, if unset, no limit would be enforced, which feels intuitive for a `maximum_*: Option<..>` field. Given its still enforced (which feels like the correct behavior), it seems strange the field is an option.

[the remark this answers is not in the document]
tnull: Ah, I see. Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/689518520e9daadb7fddd2b836bc19815ef24921.

lightning/src/routing/router.rs
tnull: Sure thing, I now split of https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/689518520e9daadb7fddd2b836bc19815ef24921.

lightning/src/routing/router.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/ebecbddf9c7375ddea65b0263e959b60eb308b6d.

lightning/src/routing/router.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/ebecbddf9c7375ddea65b0263e959b60eb308b6d.

lightning/src/routing/router.rs
```diff
@@ -237,6 +245,13 @@ impl Payee {
 	pub fn with_expiry_time(self, expiry_time: u64) -> Self {
 		Self { expiry_time: Some(expiry_time), ..self }
 	}
+
+	/// Includes a limit for the total CLTV expiry delta which is considered during routing
+	///
+	/// (C-not exported) since bindings don't support move semantics
+	pub fn with_max_total_cltv_expiry_delta(self, max_total_cltv_expiry_delta: u32) -> Self {
+		Self { max_total_cltv_expiry_delta: max_total_cltv_expiry_delta, ..self }
```
jkczyz, line 253: nit: don't need to repeat the field name now it is not an option.

```rust
Self { max_total_cltv_expiry_delta, ..self }
```

lightning/src/ln/channelmanager.rs
jkczyz, line 6907: Do we need to modify all these tests now when it's not explicitly being tested?

jkczyz (approved):

lightning/src/ln/channelmanager.rs
tnull: No, probably not. :)

lightning/src/routing/router.rs
tnull: Fixed in https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/8ea27e5faadead192baf0df9035a556ffcdd43a3

lightning/src/routing/router.rs
```diff
@@ -784,6 +800,12 @@ where L::Target: Logger {
 					// Verify the liquidity offered by this channel complies to the minimal contribution.
 					let contributes_sufficient_value = available_value_contribution_msat >= minimal_value_contribution_msat;
 
+
+					// Do not consider candidates that exceed the maximum total cltv expiry limit.
+					let max_total_cltv_expiry_delta = payee.max_total_cltv_expiry_delta;
+					let hop_total_cltv_delta = $next_hops_cltv_delta + $directional_info.cltv_expiry_delta as u32;
```
TheBlueMatt, line 806: Grr, sorry, this should be a checked_add, otherwise I believe it can overflow (which panics in debug)

tnull: Alright, is addressed by https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/f4e57f9ff08d930d58654ad8be1706e14f4d687c.

lightning/src/routing/router.rs
```diff
@@ -795,184 +813,190 @@ where L::Target: Logger {
...
-								next_hops_fee_msat: u64::max_value(),
-								hop_use_fee_msat: u64::max_value(),
-								total_fee_msat: u64::max_value(),
-								htlc_minimum_msat: $directional_info.htlc_minimum_msat,
-								path_htlc_minimum_msat,
-								path_penalty_msat: u64::max_value(),
-								was_processed: false,
-								#[cfg(any(test, feature = "fuzztarget"))]
-								value_contribution_msat,
+					if let Some(hop_total_cltv_delta) = ($next_hops_cltv_delta as u32).checked_add($directional_info.cltv_expiry_delta as u32) {
```
jkczyz, line 816: I'm touching a lot of this code in #1227. To avoid merge conflicts, could we do something like the following to keep the patch set minimal?

```rust
let hop_total_cltv_delta = ($next_hops_cltv_delta as u32)
	.checked_add($directional_info.cltv_expiry_delta as u32))
	.unwrap_or(u32::max_value());
```

lightning/src/routing/router.rs
```diff
@@ -784,6 +800,8 @@ where L::Target: Logger {
 					// Verify the liquidity offered by this channel complies to the minimal contribution.
 					let contributes_sufficient_value = available_value_contribution_msat >= minimal_value_contribution_msat;
 
+
+
```
jkczyz, line 804: nit: remove extra blank lines

TheBlueMatt (approved):

lightning/src/routing/router.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/commit/60b4bb75dfe4901ad89171e7e4ff3272097fc7e9.

lightning/src/routing/router.rs
tnull: Sure! I changed it (back) with https://github.com/lightningdevkit/rust-lightning/pull/1234/commits/367ed31dc980a099a5aa963a09c953ff80bbacd7.

TheBlueMatt (approved):

jkczyz (approved):
ACK 367ed31dc980a099a5aa963a09c953ff80bbacd7

PR #1243, opened 2022-01-17
Enable Dependabot

naveensrinivasan:
With the rise in supply chain attacks, it helps to get the dependency updates.

Even if the project decides not to update the dependency it helps to be informed. 

The dependabot settings have to be enabled in the security settings of the repository.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,36 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo" 
+    directory: "/" 
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning-background-processor" 
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning-block-sync" 
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning-invoice" 
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning-net-tokio"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning-persister"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions" 
+    directory: "/"
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo" 
+    directory: "/"
```
TheBlueMatt, line 6: Does this need to point to the subcrates we use, or is dependabot smart enough to figure out the root Cargo.toml?

naveensrinivasan: Good catch, it is not. I will update the PR to reflect that.

naveensrinivasan: Thanks! Took care of it.

On a different note I wasn't aware there were different crates. The README doesn't call out. How about including that within the README? What do you think?

.github/dependabot.yml
```diff
@@ -0,0 +1,36 @@
...
+  - package-ecosystem: "cargo" 
+    directory: "/lightning-persister"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo" 
+    directory: "/lightning"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions" 
+    directory: "/lightning"
```
TheBlueMatt, line 34: Our CI is in the root of the repo (in `.github`), so shouldn't this be "/"?

.github/dependabot.yml
TheBlueMatt: Sure, README could be updated with repo structure, sounds like a great idea (for a separate PR).

.github/dependabot.yml
naveensrinivasan: Yes, that is the default location that GitHub looks for actions , here is an example https://github.com/ossf/scorecard/blob/4837262895fa4ebc4438c94038facfee4529e4bf/.github/dependabot.yml#L10

.github/dependabot.yml
naveensrinivasan: OK, I will do a separate PR. Thanks!

Kixunil:
FIY [dependabot seems to have no special handling of MSRV](https://github.com/romanz/electrs/pull/666)

.github/dependabot.yml
TheBlueMatt: Right, but the directory here says "/lightning"? So wouldn't this look int he wrong place, then?

naveensrinivasan:
> FIY [dependabot seems to have no special handling of MSRV](https://github.com/romanz/electrs/pull/666)

I am new to Rust. Is this a blocker?

TheBlueMatt:
Its certainly very annoying, but I wouldn't call it a blocker. I assume if dependabot opens a PR and CI fails due to MSRV we can just close the PR and dependabot will shut up?

naveensrinivasan:
Yes and also we can tell dependabot to ignore that version and so going forward it will not open a PR for either that minor or major version.

TheBlueMatt:
LGTM, just seeking clarification on https://github.com/lightningdevkit/rust-lightning/pull/1243#discussion_r786276067

.github/dependabot.yml
naveensrinivasan: :face_palm

naveensrinivasan: I will fix this.

naveensrinivasan: Fixed it. Thanks

TheBlueMatt (approved):
LGTM. worst case its broken and nothing happens, best case we get automated dep-bump prs, most of which will be useless cause we have to adapt to API changes, but hey, notifications good, I guess.

naveensrinivasan:
> LGTM. worst case its broken and nothing happens, best case we get automated dep-bump prs, most of which will be useless cause we have to adapt to API changes, but hey, notifications good, I guess.

This will work only when the settings are turned on the security settings like this.

image

arik-so (approved):
LGTM

TheBlueMatt:
Looks like that was auto-enabled when we added the file to the main branch.

naveensrinivasan:
> Looks like that was auto-enabled when we added the file to the main branch.

That is surprising!

PR #1272, opened 2022-01-21
Improve KeysInterface::sign_invoice API

devrandom:
Split hrp from invoice data, to allow parsing, since there is no delimiter between the two parts in the flat preimage.


TheBlueMatt (approved):

fuzz/src/full_stack.rs
```diff
@@ -333,7 +334,7 @@ impl KeysInterface for KeyProvider {
 		))
 	}
 
-	fn sign_invoice(&self, _invoice_preimage: Vec<u8>) -> Result<RecoverableSignature, ()> {
+	fn sign_invoice(&self, hrp_bytes: &[u8], invoice_data: &[u5]) -> Result<RecoverableSignature, ()> {
```
valentinewallace, line 337: Warnings introduced, need leading underscores here and in `chanmon_consistency`

devrandom: fixed

devrandom:
unrelated error.  should probably disable rust self-update (?!) checks, although I'm not sure how.

valentinewallace (approved):

PR #1360, opened 2022-03-11
Avoid looping CLTV shadow routes.

tnull:
This PR is a follow-up to #1286: it avoids 'shadow route' random walks that loop back on the actual payment path, since such paths are implausible from an adversary's perspective.

With this I also simplified the code a bit and opted to always generate a random shadow route of a certain length. When no suitable offset can be retrieved from the (unvisited) public network graph, the hop offset is filled with the default of `MEDIAN_HOP_CLTV_EXPIRY_DELTA == 40` to simulate a partially private path.

As however already noted in #1286, such loopless random walks tend to be less diverse. For example, in the small test scenario generated by `build_graph()`, this implementation now basically always uses the fallback value since it's running out of options. It should also be noted that in order not to reduce the set of candidate paths even further, I now opted to account for visited nodes on per-path basis, i.e., crossing the streams is still possible for MPPs.

### DIFF
--- a/lightning/src/routing/router.rs
+++ b/lightning/src/routing/router.rs
@@ -1564,45 +1564,58 @@ fn add_random_cltv_offset(route: &mut Route, payment_params: &PaymentParameters,
 	for path in route.paths.iter_mut() {
 		let mut shadow_ctlv_expiry_delta_offset: u32 = 0;
 
-		// Choose the last publicly known node as the starting point for the random walk
-		if let Some(starting_hop) = path.iter().rev().find(|h| network_nodes.contains_key(&NodeId::from_pubkey(&h.pubkey))) {
-			let mut cur_node_id = NodeId::from_pubkey(&starting_hop.pubkey);
+		// Remember the last three nodes of the random walk and avoid looping back on them.
+		// Init with the last three nodes from the actual path, if possible.
+		let mut nodes_to_avoid: [NodeId; 3] = [NodeId::from_pubkey(&path.last().unwrap().pubkey),
+			NodeId::from_pubkey(&path.get(path.len().saturating_sub(2)).unwrap().pubkey),
+			NodeId::from_pubkey(&path.get(path.len().saturating_sub(3)).unwrap().pubkey)];
 
-			// Init PRNG with path nonce
-			let mut path_nonce = [0u8; 12];
-			path_nonce.copy_from_slice(&cur_node_id.as_slice()[..12]);
-			let mut prng = ChaCha20::new(random_seed_bytes, &path_nonce);
-			let mut random_path_bytes = [0u8; ::core::mem::size_of::<usize>()];
+		// Choose the last publicly known node as the starting point for the random walk.
+		let mut cur_hop: Option<NodeId> = None;
+		let mut path_nonce = [0u8; 12];
+		if let Some(starting_hop) = path.iter().rev()
+			.find(|h| network_nodes.contains_key(&NodeId::from_pubkey(&h.pubkey))) {
+				cur_hop = Some(NodeId::from_pubkey(&starting_hop.pubkey));
+				path_nonce.copy_from_slice(&cur_hop.unwrap().as_slice()[..12]);
+		}
 
-			// Pick a random path length in [1 .. 3]
-			prng.process_in_place(&mut random_path_bytes);
-			let random_walk_length = usize::from_be_bytes(random_path_bytes).wrapping_rem(3).wrapping_add(1);
+		// Init PRNG with the path-dependant nonce, which is static for private paths.
+		let mut prng = ChaCha20::new(random_seed_bytes, &path_nonce);
+		let mut random_path_bytes = [0u8; ::core::mem::size_of::<usize>()];
 
-			for _random_hop in 0..random_walk_length {
+		// Pick a random path length in [1 .. 3]
+		prng.process_in_place(&mut random_path_bytes);
+		let random_walk_length = usize::from_be_bytes(random_path_bytes).wrapping_rem(3).wrapping_add(1);
+
+		for random_hop in 0..random_walk_length {
+			// If we don't find a suitable offset in the public network graph, we default to
+			// MEDIAN_HOP_CLTV_EXPIRY_DELTA.
+			let mut random_hop_offset = MEDIAN_HOP_CLTV_EXPIRY_DELTA;
+
+			if let Some(cur_node_id) = cur_hop {
 				if let Some(cur_node) = network_nodes.get(&cur_node_id) {
-					// Randomly choose the next hop
+					// Randomly choose the next unvisited hop.
 					prng.process_in_place(&mut random_path_bytes);
-					if let Some(random_channel) = usize::from_be_bytes(random_path_bytes).checked_rem(cur_node.channels.len())
+					if let Some(random_channel) = usize::from_be_bytes(random_path_bytes)
+						.checked_rem(cur_node.channels.len())
 						.and_then(|index| cur_node.channels.get(index))
 						.and_then(|id| network_channels.get(id)) {
 							random_channel.as_directed_from(&cur_node_id).map(|(dir_info, next_id)| {
-								dir_info.direction().map(|channel_update_info|
-									shadow_ctlv_expiry_delta_offset = shadow_ctlv_expiry_delta_offset
-										.checked_add(channel_update_info.cltv_expiry_delta.into())
-										.unwrap_or(shadow_ctlv_expiry_delta_offset));
-								cur_node_id = *next_id;
+								if !nodes_to_avoid.iter().any(|x| x == next_id) {
+									nodes_to_avoid[random_hop] = *next_id;
+									dir_info.direction().map(|channel_update_info| {
+										random_hop_offset = channel_update_info.cltv_expiry_delta.into();
+										cur_hop = Some(*next_id);
+									});
+								}
 							});
 						}
 				}
 			}
-		} else {
-			// If the entire path is private, choose a random offset from multiples of
-			// MEDIAN_HOP_CLTV_EXPIRY_DELTA
-			let mut prng = ChaCha20::new(random_seed_bytes, &[0u8; 8]);
-			let mut random_bytes = [0u8; 4];
-			prng.process_in_place(&mut random_bytes);
-			let random_walk_length = u32::from_be_bytes(random_bytes).wrapping_rem(3).wrapping_add(1);
-			shadow_ctlv_expiry_delta_offset = random_walk_length * MEDIAN_HOP_CLTV_EXPIRY_DELTA;
+
+			shadow_ctlv_expiry_delta_offset = shadow_ctlv_expiry_delta_offset
+				.checked_add(random_hop_offset)
+				.unwrap_or(shadow_ctlv_expiry_delta_offset);
 		}
 
 		// Limit the total offset to reduce the worst-case locked liquidity timevalue


tnull:
Since the fuzzer was failing at a related point, I rebased this now off #1358.

TheBlueMatt:
Sorry about the fuzz failure, #1358 landed now so feel free to rebase.

tnull:
> Sorry about the fuzz failure, #1358 landed now so feel free to rebase.

No apology necessary, especially in this case... Rebased on main now.

lightning/src/routing/router.rs
```diff
@@ -1564,45 +1564,55 @@ fn add_random_cltv_offset(route: &mut Route, payment_params: &PaymentParameters,
 	for path in route.paths.iter_mut() {
 		let mut shadow_ctlv_expiry_delta_offset: u32 = 0;
 
-		// Choose the last publicly known node as the starting point for the random walk
-		if let Some(starting_hop) = path.iter().rev().find(|h| network_nodes.contains_key(&NodeId::from_pubkey(&h.pubkey))) {
-			let mut cur_node_id = NodeId::from_pubkey(&starting_hop.pubkey);
+		// Mark all nodes on the actual path as visited to avoid looping random walks.
+		let mut visited_nodes: HashSet<NodeId> = HashSet::new();
```
TheBlueMatt, line 1568: I'd think it'd be faster to keep a single vec and just iterate it, but at a higer level, do we care about avoiding taking any node that we preiviously took in the path, or do we only care about not going back on the last three hops of the path, that'd let us even keep a fixed length array of like 3 and not hit the heap at all.

tnull: > I'd think it'd be faster to keep a single vec and just iterate it, 

Hm, true, that might be faster since the routes are usually really not long. a26083ee24eb9c2371c3e53489b4ef099ef73a64 now utilizes a `Vec`.

> but at a higer level, do we care about avoiding taking any node that we preiviously took in the path, or do we only care about not going back on the last three hops of the path, that'd let us even keep a fixed length array of like 3 and not hit the heap at all.

Well, I'd argue that if we bother about looping shadow routes in the first place, we should do it 'properly' and avoid any loops. That said, if we don't want to hit the heap, we could also just use a fixed array with the maximum length of 20 to be on the safe side, no? (Which reminds me that I probably should limit the shadow route length so that the overall length of path + shadow route never exceeds 20, because otherwise it becomes implausible anyways).

tnull: I realize I had the fixed pre-TLV limit in mind, but limiting the shadow route length in conjunction with https://github.com/lightningdevkit/rust-lightning/issues/1371 is probably still a good idea.

TheBlueMatt: > Well, I'd argue that if we bother about looping shadow routes in the first place, we should do it 'properly' and avoid any loops.

Right, I'm not sure if its really "properly" - if you're a node one hop from the end of a 5 hop route, the fact that the route appears to loop back over the first hop is actually a signal that that first hop node is not in the path, whereas if we don't consider looping over the first hop an issue such a node gets no signal. Its a bit of an awkward balancing act, though - allowing looping removes signal for nodes late in the path, but not allowing looping removes a signal for nodes early in the path.

> That said, if we don't want to hit the heap, we could also just use a fixed array with the maximum length of 20 to be on the safe side, no?

I mean its not a *huge* concern, just always nice to avoid it where possible - its never free and I'm generally scarred from overuse of the heap/allocations/reallocations in Bitcoin Core that causes severe heap fragmentation and memory usage bloat in excess of what it really needs...something Core has never been in a position to materially improve :(.

TheBlueMatt: Yea, I think we're on the same page. I'd be happy to see you go to only the last three hops, I'll leave it up to you exactly how you want to proceed, though, wanted to get on the same page first.

tnull: With https://github.com/lightningdevkit/rust-lightning/pull/1360/commits/85dbc5ddf03b386c75cb587c0ae0c6c32a1ca704, I switched to a fixed-length array that remembers only the three most recent nodes, which is initialized from the path suffix if possible.

TheBlueMatt (dismissed):

lightning/src/routing/router.rs
```diff
@@ -1564,45 +1564,61 @@ fn add_random_cltv_offset(route: &mut Route, payment_params: &PaymentParameters,
...
+		let mut random_path_bytes = [0u8; ::core::mem::size_of::<usize>()];
 
-			// Init PRNG with path nonce
-			let mut path_nonce = [0u8; 12];
-			path_nonce.copy_from_slice(&cur_node_id.as_slice()[..12]);
-			let mut prng = ChaCha20::new(random_seed_bytes, &path_nonce);
-			let mut random_path_bytes = [0u8; ::core::mem::size_of::<usize>()];
+		// Pick a random path length in [1 .. 3]
+		prng.process_in_place(&mut random_path_bytes);
+		let random_walk_length = usize::from_be_bytes(random_path_bytes).wrapping_rem(3).wrapping_add(1);
```
jkczyz, line 1590: Think there was a similar comment in another PR, but no need for the wrapping variants here given the use of `usize` and `3`.

lightning/src/routing/router.rs
```diff
@@ -1564,45 +1564,61 @@ fn add_random_cltv_offset(route: &mut Route, payment_params: &PaymentParameters,
 	for path in route.paths.iter_mut() {
 		let mut shadow_ctlv_expiry_delta_offset: u32 = 0;
 
-		// Choose the last publicly known node as the starting point for the random walk
-		if let Some(starting_hop) = path.iter().rev().find(|h| network_nodes.contains_key(&NodeId::from_pubkey(&h.pubkey))) {
-			let mut cur_node_id = NodeId::from_pubkey(&starting_hop.pubkey);
+		// Remember the last three nodes of the random walk and avoid looping back on them.
+		// Init with the last three nodes from the actual path, if possible.
+		let last_id = NodeId::from_pubkey(&path.last().unwrap().pubkey);
```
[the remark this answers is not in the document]
jkczyz: Out of curiosity, any reason why we don't check the entire path plus any selected shadow nodes? Just wondering if the additional complexity is justified if paths are effectively bounded.

lightning/src/routing/router.rs
```diff
@@ -1564,45 +1564,61 @@ fn add_random_cltv_offset(route: &mut Route, payment_params: &PaymentParameters,
...
 						.and_then(|id| network_channels.get(id)) {
 							random_channel.as_directed_from(&cur_node_id).map(|(dir_info, next_id)| {
-								dir_info.direction().map(|channel_update_info|
-									shadow_ctlv_expiry_delta_offset = shadow_ctlv_expiry_delta_offset
-										.checked_add(channel_update_info.cltv_expiry_delta.into())
-										.unwrap_or(shadow_ctlv_expiry_delta_offset));
-								cur_node_id = *next_id;
+								if !nodes_to_avoid.iter().any(|x| x == next_id) {
+									nodes_to_avoid[nodes_ring_index] = *next_id;
+									nodes_ring_index = (nodes_ring_index + 1) % 3;
```
jkczyz, line 1608: Regarding the modulo operator and name `nodes_ring_index`, is it ever the case we'll loop back to the `0` index? Seems like we'll always loop use at most `3` times.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
jkczyz: Ah, I see. I guess with this approach, though, you may still loop over one of the last three hops since they are overwritten by shadow hops. Is that a concern?

There may be a way to avoid that using iterators chained with three `Option`s (or an array of them). For example, with one shadow hop it would be something like:

```rust
let shadown_hop = None;
// ...
path.iter().rev().take(3).chain(shadow_hop.iter()).any(...)
```

Though you may have to do the pubkey to node id conversion more than once with a `map` somewhere in there.

lightning/src/routing/router.rs
tnull: Yes, we had this discussion before, and I agree that 'normal' arithmetic can be more readable in many circumstances. However, in instances like this where already some function is applied I personally prefer the 'functional style' readability allowing to read the chained operations left to right over mixing the two like `(usize::from_be_bytes(random_path_bytes) % 3) + 1`.

lightning/src/routing/router.rs
tnull: Ah good point. In https://github.com/lightningdevkit/rust-lightning/pull/1360/commits/eb5a4c478aa2c98d0c04f38615e66296c673c20b I now simply use the previously unused `random_hop` index value of the `for` loop.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
tnull: > super-duper-tiny-nit(cause I hope LLVM will optimize this out, though it may not): no need to create three `NodeId`s and then copy them into the array, you can just create them directly in the array initialization and put the array init on a few lines.

Ah, I assumed the compiler would optimize this away. Addressed with https://github.com/lightningdevkit/rust-lightning/pull/1360/commits/eb5a4c478aa2c98d0c04f38615e66296c673c20b.

> Ah, I see. I guess with this approach, though, you may still loop over one of the last three hops since they are overwritten by shadow hops. Is that a concern?

No, from my side it was intentional to just remember the last three hops and avoid having the random walk loop back on them. This should exclude many trivially implausible shadow routes, while still giving the random walk some degrees of freedom.
 
> There may be a way to avoid that using iterators chained with three `Option`s (or an array of them). For example, with one shadow hop it would be something like:
> 
> ```rust
> let shadown_hop = None;
> // ...
> path.iter().rev().take(3).chain(shadow_hop.iter()).any(...)
> ```
> 
> Though you may have to do the pubkey to node id conversion more than once with a `map` somewhere in there.

I'm not entirely sure what you are proposing, but I first played around with a similar construction to initialize `nodes_to_avoid`, i.e., something like `path.iter().rev().take(3).map(|h| NodeId::from_pubkey(&h.pubkey)`, which I however found to be not easily `collect`ed into an array.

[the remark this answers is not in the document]
jkczyz: > > There may be a way to avoid that using iterators chained with three `Option`s (or an array of them). For example, with one shadow hop it would be something like:
> > ```rust
> > let shadown_hop = None;
> > // ...
> > path.iter().rev().take(3).chain(shadow_hop.iter()).any(...)
> > ```
...
> I'm not entirely sure what you are proposing,

The suggestion above was demonstrating how to avoid looping back on a single shadow hop and the original path's 3-suffix without allocations. It would require a little more work to support three shadow hops. And as note it would mean `NodeId::from_pubkey` would have to be called on each loop iteration. But given that `NodeId::from_pubkey` will allocate a `Vec`, maybe it's not worth it.

jkczyz:
Overall looks good. Fine with squashing and merging as is. Just want to note my observation that `Node::from_pubkey` calls `PublicKey::serialize`, which allocates a `Vec`. So while we avoid one allocation by using an array, there is still three allocations there.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
tnull: Ah, after re-reading I'm getting what you wrote. Sorry, first time my brain failed to parse. Chaining iterators seems like a neat alternative way to go about this!

tnull:
Rebased and squashed.

TheBlueMatt:
> Rebased and squashed.

Hmm, did you have to rebase? Please don't rebase unless there's an actual conflict as it means reviewers can't as easily compare to the previous state.

lightning/src/routing/router.rs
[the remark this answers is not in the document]
TheBlueMatt: It returns a fixed-length array, and passes the pointer to C, there's no allocation there - https://docs.rs/secp256k1/latest/src/secp256k1/key.rs.html#424-440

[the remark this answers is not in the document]
TheBlueMatt: Of course that said, `PublicKey::serialize` is *not* free, so if performance is the concern, we'd still want to avoid it, but its three calls so nbd.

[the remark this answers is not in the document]
jkczyz: Ah, whoops. You're right. I was looking at the wrong type / implementation.

https://docs.rs/bitcoin/0.27.1/bitcoin/util/psbt/serialize/trait.Serialize.html

TheBlueMatt (dismissed):

tnull:
> Looks like github actions are acting up. Could you run the following?

Done!

> Hmm, did you have to rebase? Please don't rebase unless there's an actual conflict as it means reviewers can't as easily compare to the previous state.

Ah, sorry, didn't realize that. Will keep in mind.

lightning/src/routing/router.rs
jkczyz: Ah, makes sense. Could also use the non-wrapping versions `rem` and `add`. No need to change though as I think we want to just merge now.

jkczyz (approved):

PR #1416, opened 2022-04-10
Add utils to persist scorer in BackgroundProcessor

jurvis:
Addresses https://github.com/lightningdevkit/rust-lightning/issues/1191

## Implementation
1. [Added](https://github.com/lightningdevkit/rust-lightning/compare/main...jurvis:jurvis/persist-scorer?expand=1#diff-b00a115e9303b98772e06346a86a0312aa9790d78a73904eb6a20f938cd6da0fR138-R154) a super-trait `WriteableScore` that makes `LockableScore` writeable.
2. Added a new parameter to `BackgroundProcessor.start` that takes in our new `WriteableScore`
3. Added a new `persist_scorer` function to `Persister`, and implemented it with `BufWriter` (placeholder; to update after https://github.com/lightningdevkit/rust-lightning/pull/1404) is merged

## Notes/Questions
(see diff comments)

### DIFF
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -278,8 +278,6 @@ impl BackgroundProcessor {
 						if let Err(e) = persister.persist_graph(handler.network_graph()) {
 							log_error!(logger, "Error: Failed to persist network graph, check your disk and permissions {}", e)
 						}
-						last_prune_call = Instant::now();
-						have_pruned = true;
 					}
 					if let Some(ref scorer) = scorer {
 						log_trace!(logger, "Persisting scorer");
@@ -287,6 +285,9 @@ impl BackgroundProcessor {
 							log_error!(logger, "Error: Failed to persist scorer, check your disk and permissions {}", e)
 						}
 					}
+
+					last_prune_call = Instant::now();
+					have_pruned = true;
 				}
 			}
 
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -18,6 +18,7 @@ use lightning::ln::channelmanager::ChannelManager;
 use lightning::ln::msgs::{ChannelMessageHandler, RoutingMessageHandler};
 use lightning::ln::peer_handler::{CustomMessageHandler, PeerManager, SocketDescriptor};
 use lightning::routing::network_graph::{NetworkGraph, NetGraphMsgHandler};
+use lightning::routing::scoring::WriteableScore;
 use lightning::util::events::{Event, EventHandler, EventsProvider};
 use lightning::util::logger::Logger;
 use lightning::util::persist::Persister;
@@ -151,6 +152,7 @@ impl BackgroundProcessor {
 	/// [`NetworkGraph`]: lightning::routing::network_graph::NetworkGraph
 	/// [`NetworkGraph::write`]: lightning::routing::network_graph::NetworkGraph#impl-Writeable
 	pub fn start<
+		'a,
 		Signer: 'static + Sign,
 		CA: 'static + Deref + Send + Sync,
 		CF: 'static + Deref + Send + Sync,
@@ -171,9 +173,11 @@ impl BackgroundProcessor {
 		NG: 'static + Deref<Target = NetGraphMsgHandler<G, CA, L>> + Send + Sync,
 		UMH: 'static + Deref + Send + Sync,
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
+		S: 'static + Deref<Target = SC> + Send + Sync,
+		SC: WriteableScore<'a>,
 	>(
 		persister: PS, event_handler: EH, chain_monitor: M, channel_manager: CM,
-		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L
+		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L, scorer: Option<S>
 	) -> Self
 	where
 		CA::Target: 'static + chain::Access,
@@ -187,7 +191,7 @@ impl BackgroundProcessor {
 		CMH::Target: 'static + ChannelMessageHandler,
 		RMH::Target: 'static + RoutingMessageHandler,
 		UMH::Target: 'static + CustomMessageHandler,
-		PS::Target: 'static + Persister<Signer, CW, T, K, F, L>
+		PS::Target: 'static + Persister<'a, Signer, CW, T, K, F, L, SC>,
 	{
 		let stop_thread = Arc::new(AtomicBool::new(false));
 		let stop_thread_clone = stop_thread.clone();
@@ -277,6 +281,12 @@ impl BackgroundProcessor {
 						last_prune_call = Instant::now();
 						have_pruned = true;
 					}
+					if let Some(ref scorer) = scorer {
+						log_trace!(logger, "Persisting scorer");
+						if let Err(e) = persister.persist_scorer(&scorer) {
+							log_error!(logger, "Error: Failed to persist scorer, check your disk and permissions {}", e)
+						}
+					}
 				}
 			}
 
@@ -285,10 +295,16 @@ impl BackgroundProcessor {
 			// ChannelMonitor update(s) persisted without a corresponding ChannelManager update.
 			persister.persist_manager(&*channel_manager)?;
 
+			// Persist Scorer on exit
+			if let Some(ref scorer) = scorer {
+				persister.persist_scorer(&scorer)?;
+			}
+
 			// Persist NetworkGraph on exit
 			if let Some(ref handler) = net_graph_msg_handler {
 				persister.persist_graph(handler.network_graph())?;
 			}
+
 			Ok(())
 		});
 		Self { stop_thread: stop_thread_clone, thread_handle: Some(handle) }
@@ -369,6 +385,7 @@ mod tests {
 	use std::path::PathBuf;
 	use std::sync::{Arc, Mutex};
 	use std::time::Duration;
+	use lightning::routing::scoring::{FixedPenaltyScorer};
 	use super::{BackgroundProcessor, FRESHNESS_TIMER};
 
 	const EVENT_DEADLINE: u64 = 5 * FRESHNESS_TIMER;
@@ -395,6 +412,7 @@ mod tests {
 		network_graph: Arc<NetworkGraph>,
 		logger: Arc<test_utils::TestLogger>,
 		best_block: BestBlock,
+		scorer: Arc<Mutex<FixedPenaltyScorer>>,
 	}
 
 	impl Drop for Node {
@@ -410,13 +428,14 @@ mod tests {
 	struct Persister {
 		graph_error: Option<(std::io::ErrorKind, &'static str)>,
 		manager_error: Option<(std::io::ErrorKind, &'static str)>,
+		scorer_error: Option<(std::io::ErrorKind, &'static str)>,
 		filesystem_persister: FilesystemPersister,
 	}
 
 	impl Persister {
 		fn new(data_dir: String) -> Self {
 			let filesystem_persister = FilesystemPersister::new(data_dir.clone());
-			Self { graph_error: None, manager_error: None, filesystem_persister }
+			Self { graph_error: None, manager_error: None, scorer_error: None, filesystem_persister }
 		}
 
 		fn with_graph_error(self, error: std::io::ErrorKind, message: &'static str) -> Self {
@@ -425,6 +444,10 @@ mod tests {
 
 		fn with_manager_error(self, error: std::io::ErrorKind, message: &'static str) -> Self {
 			Self { manager_error: Some((error, message)), ..self }
+		}
+
+		fn with_scorer_error(self, error: std::io::ErrorKind, message: &'static str) -> Self {
+			Self { scorer_error: Some((error, message)), ..self }
 		}
 	}
 
@@ -438,6 +461,12 @@ mod tests {
 
 			if key == "network_graph" {
 				if let Some((error, message)) = self.graph_error {
+					return Err(std::io::Error::new(error, message))
+				}
+			}
+
+			if key == "scorer" {
+				if let Some((error, message)) = self.scorer_error {
 					return Err(std::io::Error::new(error, message))
 				}
 			}
@@ -473,7 +502,8 @@ mod tests {
 			let net_graph_msg_handler = Some(Arc::new(NetGraphMsgHandler::new(network_graph.clone(), Some(chain_source.clone()), logger.clone())));
 			let msg_handler = MessageHandler { chan_handler: Arc::new(test_utils::TestChannelMessageHandler::new()), route_handler: Arc::new(test_utils::TestRoutingMessageHandler::new() )};
 			let peer_manager = Arc::new(PeerManager::new(msg_handler, keys_manager.get_node_secret(Recipient::Node).unwrap(), &seed, logger.clone(), IgnoringMessageHandler{}));
-			let node = Node { node: manager, net_graph_msg_handler, peer_manager, chain_monitor, persister, tx_broadcaster, network_graph, logger, best_block };
+			let scorer = Arc::new(Mutex::new(test_utils::TestScorer::with_penalty(0)));
+			let node = Node { node: manager, net_graph_msg_handler, peer_manager, chain_monitor, persister, tx_broadcaster, network_graph, logger, best_block, scorer };
 			nodes.push(node);
 		}
 
@@ -571,7 +601,7 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
 
 		macro_rules! check_persisted_data {
 			($node: expr, $filepath: expr) => {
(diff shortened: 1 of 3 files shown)


lightning-persister/src/lib.rs
```diff
@@ -117,6 +119,27 @@ impl FilesystemPersister {
...
+		let file = fs::File::create(&tmp_filename)?;
+		let write_res = scorer.write(&mut BufWriter::new(file));
+		if let Err(e) = write_res.and_then(|_| fs::rename(&tmp_filename, filename_with_path)) {
+			let _ = fs::remove_file(&tmp_filename);
+			Err(e)
+		} else {
+			Ok(())
+		}
+	}
+
```
jurvis, line 142: This is a placeholder implementation I borrowed from combining ideas from the sample and #1404. I will change it after that PR is merged. 

I reckon it is probably best to use `utils::write_to_file`, but that currently requires D to be `DiskWriteable`, which may conflict with our new trait `WriteableScore`. Any advice?

lightning-background-processor/src/lib.rs
```diff
@@ -180,15 +185,16 @@ impl BackgroundProcessor {
...
 		M: 'static + Deref<Target = ChainMonitor<Signer, CF, T, F, L, P>> + Send + Sync,
 		CM: 'static + Deref<Target = ChannelManager<Signer, CW, T, K, F, L>> + Send + Sync,
 		NG: 'static + Deref<Target = NetGraphMsgHandler<G, CA, L>> + Send + Sync,
 		UMH: 'static + Deref + Send + Sync,
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
+		S: 'static + Deref + Send + Sync,
 	>(
 		persister: PS, event_handler: EH, chain_monitor: M, channel_manager: CM,
-		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L
+		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L, scorer: S
```
TheBlueMatt, line 197: The scorer here should be `Option`al, I think.

lightning-background-processor/src/lib.rs
```diff
@@ -604,6 +633,35 @@ mod tests {
 			}
 		}
 
+		macro_rules! check_mutex_persisted_data {
+			($node: expr, $filepath: expr) => {
+				let mut expected_bytes = Vec::new();
+				loop {
+					expected_bytes.clear();
+					match $node.lock() {
```
TheBlueMatt, line 641: nit: generally I think almost all rust code either `unwrap`s or `expect`s mutex `lock()` calls - they can only fail if the lock had previously been held during a panic, and we generally strongly recommend LDK be built with `panic=abort` so it is actually unreachable (and we do so in tests IIRC).

TheBlueMatt (commented):
Thanks! Sorry for the review delay - been travelling for Bitcoin Miami.

lightning-background-processor/src/lib.rs
```diff
@@ -291,6 +301,10 @@ impl BackgroundProcessor {
 			if let Some(ref handler) = net_graph_msg_handler {
 				persister.persist_graph(handler.network_graph())?;
 			}
+
+			// Persist Scorer on exit
+			persister.persist_scorer(&scorer)?;
+
 			Ok(())
```
vincenzopalazzo, line 308:
```suggestion
			// Persist Scorer on exit
			persister.persist_scorer(&scorer)?;

			Ok(())
```

Why the failure is ignored and reported `Ok()` any way? not able to catch this, sorry

vincenzopalazzo (changes requested):

lightning-persister/src/lib.rs
TheBlueMatt: You can implement `DiskWriteable` for any `WriteableScore` to fix it. `DiskWriteable` is also removed in #1417 so you can also wait for that to go in first (but I'll let you and @johncantrell97 figure out which order y'all want to land these).

jurvis:
sounds good, thanks for your review. @johncantrell97 it looks like you have some good stuff in that PR so I'll let you merge that in first and build scorer persistence around your changes!

lightning-persister/src/lib.rs
jurvis: @johncantrell97 got it, thanks for clarifying!

TheBlueMatt:
Needs rebase now that #1417 has landed.

lightning/src/routing/scoring.rs
```diff
@@ -135,6 +135,23 @@ pub trait LockableScore<'a> {
 	fn lock(&'a self) -> Self::Locked;
 }
 
+/// Refers to a scorer that is accessible under lock and also writeable to disk
+///
+/// We need this trait to be able to pass in a scorer to `lightning-background-processor` that will enable us to
+/// use the Persister to persist it.
+pub trait WriteableScore<'a>: LockableScore<'a> {
+	/// Locks the LockableScore and writes it to disk
+	fn write<W: Writer>(&'a self, writer: &mut W) -> Result<(), io::Error>;
```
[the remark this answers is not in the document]
jurvis: hm, I just assumed from the sample app that we wanted to make sure that the scorer is lockable so we can do `.lock().write()`. If `WriteableScore` were to just only extend `Writeable`, we won't be able to lock it before writing it.

lightning/src/routing/scoring.rs
```diff
@@ -135,6 +135,23 @@ pub trait LockableScore<'a> {
...
+}
+
+impl<'a, U: Writeable, T: LockableScore<'a>> WriteableScore<'a> for T
+	where T::Locked: DerefMut<Target=U>
+{
+    fn write<W: Writer>(&'a self, writer: &mut W) -> Result<(), io::Error> {
```
TheBlueMatt, line 150: nit: tabs not spaces.

lightning/src/routing/scoring.rs
```diff
@@ -135,6 +135,23 @@ pub trait LockableScore<'a> {
 	fn lock(&'a self) -> Self::Locked;
 }
 
+/// Refers to a scorer that is accessible under lock and also writeable to disk
+///
+/// We need this trait to be able to pass in a scorer to `lightning-background-processor` that will enable us to
+/// use the Persister to persist it.
+pub trait WriteableScore<'a>: LockableScore<'a> {
```
TheBlueMatt, line 142: Nice, I think with this we can simplify some of the bindings changes we've been maintaining on previous versions. Don't need to bother with it here but I'll take a look as a followup after we land this.

TheBlueMatt (commented):
Nice! Thanks. Some relatively minor nits and minor comments.

lightning/src/routing/scoring.rs
```diff
@@ -135,6 +137,32 @@ pub trait LockableScore<'a> {
...
+///
+/// We need this trait to be able to pass in a scorer to `lightning-background-processor` that will enable us to
+/// use the Persister to persist it.
+pub trait WriteableScore<'a>: LockableScore<'a> + Writeable {}
+
+impl<'a, T> WriteableScore<'a> for T
+	where T: LockableScore<'a> + Writeable
+{}
+
+impl<'a, T> LockableScore<'a> for Arc<T>
```
TheBlueMatt, line 150: This can be blanket, no? Instead of implementing it only for `Arc` we can implement it for any `Deref<T> where T: LockableScore and T::Locked: Writeable`.

jurvis: touché, not sure why I didn't think of that

jurvis: @TheBlueMatt I tried implementing a blanket using what you suggested:

```
impl<'a, T, S> LockableScore<'a> for S
	where S: Deref<Target = T>,
		T: LockableScore<'a>,
		T::Locked: Writeable {
	type Locked = T::Locked;

	fn lock(&'a self) -> Self::Locked {
		self.deref().lock()
	}
}
```

but it looks like it may be too generic as it conflicts with our implementation of `LockableScore<'a>` for `Mutex` and `RefCell` a couple of lines below. thoughts?

TheBlueMatt: Ah, yea, lets not go that way - instead lets just make the `WriteableScore` we pass to the `BackgroundProcessor` actually a `Deref<WriteableScore>`, then we can drop both of the Arc implementations. You'll need to have two bounds - one for the `WriteableScore` and one for the `Deref` as you'll want to re-use the `WriteableScore` bound on the `Persister.

vincenzopalazzo (dismissed):

jurvis:
@TheBlueMatt fixed and squashed.

> either way lets not clean up previous commits in later ones.

I didn't know what you meant by this, but all the commits look like they are squashable so I did it anyways.

lightning/src/util/persist.rs
```diff
@@ -24,37 +25,47 @@ pub trait KVStorePersister {
...
 }
 
-impl<A: KVStorePersister, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> Persister<Signer, M, T, K, F, L> for A
+impl<'a, A: KVStorePersister, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref, S> Persister<'a, Signer, M, T, K, F, L, S> for A
 	where M::Target: 'static + chain::Watch<Signer>,
 		T::Target: 'static + BroadcasterInterface,
 		K::Target: 'static + KeysInterface<Signer = Signer>,
 		F::Target: 'static + FeeEstimator,
 		L::Target: 'static + Logger,
+		S: WriteableScore<'a>
```
jkczyz, line 53: nit: add trailing comma

lightning/src/util/persist.rs
```diff
@@ -24,37 +25,47 @@ pub trait KVStorePersister {
...
-/// Trait that handles persisting a [`ChannelManager`] and [`NetworkGraph`] to disk.
-pub trait Persister<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>
+/// Trait that handles persisting a [`ChannelManager`], [`NetworkGraph`], and [`WriteableScore`] to disk.
+pub trait Persister<'a, Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref, S>
 	where M::Target: 'static + chain::Watch<Signer>,
 		T::Target: 'static + BroadcasterInterface,
 		K::Target: 'static + KeysInterface<Signer = Signer>,
 		F::Target: 'static + FeeEstimator,
 		L::Target: 'static + Logger,
+		S: WriteableScore<'a>
```
jkczyz, line 35: nit: add trailing comma

lightning/src/routing/scoring.rs
```diff
@@ -137,6 +137,16 @@ pub trait LockableScore<'a> {
...
 
+/// Refers to a scorer that is accessible under lock and also writeable to disk
+///
+/// We need this trait to be able to pass in a scorer to `lightning-background-processor` that will enable us to
+/// use the Persister to persist it.
+pub trait WriteableScore<'a>: LockableScore<'a> + Writeable {}
+
+impl<'a, T> WriteableScore<'a> for T
+	where T: LockableScore<'a> + Writeable
+{}
```
jkczyz, line 148: This can fit on one line within a 100 characters, which is our aspirational line length limit. 🙂

lightning-background-processor/src/lib.rs
```diff
@@ -171,9 +173,11 @@ impl BackgroundProcessor {
 		NG: 'static + Deref<Target = NetGraphMsgHandler<G, CA, L>> + Send + Sync,
 		UMH: 'static + Deref + Send + Sync,
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
+		S: 'static + Deref<Target = SC> + Send + Sync,
+		SC: WriteableScore<'a>
```
jkczyz, line 177: nit: add trailing comma

lightning-background-processor/src/lib.rs
```diff
@@ -277,6 +285,12 @@ impl BackgroundProcessor {
 			if let Some(ref handler) = net_graph_msg_handler {
 				persister.persist_graph(handler.network_graph())?;
 			}
+
+			// Persist Scorer on exit
+			if let Some(ref scorer) = scorer {
```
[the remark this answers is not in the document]
jkczyz: Would it be worth doing the same in the loop?

lightning-background-processor/src/lib.rs
```diff
@@ -411,12 +426,13 @@ mod tests {
 		graph_error: Option<(std::io::ErrorKind, &'static str)>,
 		manager_error: Option<(std::io::ErrorKind, &'static str)>,
 		filesystem_persister: FilesystemPersister,
+		scorer_error: Option<(std::io::ErrorKind, &'static str)>
```
jkczyz, line 429: Let's move this before the prior field, which feels more logically consistent.

lightning-background-processor/src/lib.rs
```diff
@@ -632,7 +663,8 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
+		let scorer = Arc::new(Mutex::new(test_utils::TestScorer::with_penalty(0)));
```
jkczyz, line 666: Here and later tests, if the test doesn't explicitly check scoring, can we pass in `None` instead? Alternatively, add it the test `Node` struct so it only needs to be created in one place.

jkczyz (commented):
Looks good! Mostly just some trivial clean-ups / nits.

lightning-background-processor/src/lib.rs
jurvis: Done. I wasn't able to use just `None` because the compiler wasn't able to infer the type:

> cannot infer type for type parameter `S` declared on the associated function

so I added `None::<Arc<Mutex<FixedPenaltyScorer>>>`. 

I know `FixedPenaltyScorer` is deprecated, however, when I tried `None::<Arc<Mutex<ProbabilisticScorer<NetworkGraph, TestLogger>>>`, the compiler complains:

```
669 | ...e(), None::<Arc<Mutex<ProbabilisticScorer<NetworkGraph, Logger>>>>);
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Deref` is not implemented for `NetworkGraph`
```

is that a problem?

also, should we create a follow-up issue for updating `TestScorer` to return a `ProbabilisticScorer` instead of `FixedPenaltyScorer`?

jkczyz: > Done. I wasn't able to use just `None` because the compiler wasn't able to infer the type:
> 
> > cannot infer type for type parameter `S` declared on the associated function
> 
> so I added `None::<Arc<Mutex<FixedPenaltyScorer>>>`.
> 

Hmm... right, that's a bit verbose. I'd say add a `scorer` field to `Node`  and do `nodes[0].scorer.clone()`.

> I know `FixedPenaltyScorer` is deprecated, however, when I tried `None::<Arc<Mutex<ProbabilisticScorer<NetworkGraph, TestLogger>>>`, the compiler complains:
> 
> ```
> 669 | ...e(), None::<Arc<Mutex<ProbabilisticScorer<NetworkGraph, Logger>>>>);
>     |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Deref` is not implemented for `NetworkGraph`
> ```
> 
> is that a problem?
> 

With this approach you would need to use `&NetworkGraph` or `Arc<NetworkGraph>` since the type parameter must implement `Deref`.

> also, should we create a follow-up issue for updating `TestScorer` to return a `ProbabilisticScorer` instead of `FixedPenaltyScorer`?

No need. `Scorer` is deprecated not, `FixedPenaltyScorer`.

jurvis: @jkczyz gotcha. I guess if `FixedPenaltyScorer` is not deprecated I won't change it to `ProbabilisticScorer` then. Just added the scorer field to Node. should be ready for review again :) thanks!

lightning-background-processor/src/lib.rs
```diff
@@ -571,7 +600,8 @@ mod tests {
...
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(scorer.clone()));
```
jkczyz, line 604: Can use the scorer in `Node` here now.

jkczyz (approved):
Looks good modulo one comment.

lightning-background-processor/src/lib.rs
jurvis: oops, thanks for catching that.

lightning-background-processor/src/lib.rs
```diff
@@ -571,7 +600,8 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone());
+		let scorer = Arc::new(Mutex::new(test_utils::TestScorer::with_penalty(0)));
```
jkczyz, line 603: Will want to remove this and replace the usage below now. 🙂

jurvis: 🤦 I'm so sorry -- fixed

jkczyz (dismissed):

jurvis:
@TheBlueMatt good catch! resolved.

lightning-background-processor/src/lib.rs
```diff
@@ -277,6 +281,11 @@ impl BackgroundProcessor {
 						last_prune_call = Instant::now();
 						have_pruned = true;
 					}
+					if let Some(ref scorer) = scorer {
```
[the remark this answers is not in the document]
jkczyz: Looks like this predates this change. @jurvis Could you make a separate commit for this fix?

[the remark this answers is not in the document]
jurvis: sure, I can do that 😄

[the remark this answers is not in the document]
jurvis: @jkczyz done

jkczyz (approved):

PR #1425, opened 2022-04-15
Wumbo!

valentinewallace:
Wumbo channels! 🥳 

I'm not sure if we care about the `TOTAL_BITCOIN_SUPPLY` check, so I haven't tested it yet. Let me know if we want it and I'll add testing.


lightning/src/ln/channel.rs
```diff
@@ -1076,8 +1083,14 @@ impl<Signer: Sign> Channel<Signer> {
...
 		// Check sanity of message fields:
-		if msg.funding_satoshis >= MAX_FUNDING_SATOSHIS {
-			return Err(ChannelError::Close(format!("Funding must be smaller than {}. It was {}", MAX_FUNDING_SATOSHIS, msg.funding_satoshis)));
+		if !their_features.supports_wumbo() && msg.funding_satoshis >= MAX_FUNDING_SATOSHIS_NO_WUMBO {
+			return Err(ChannelError::Close(format!("Funding must be smaller than {}. It was {}", MAX_FUNDING_SATOSHIS_NO_WUMBO, msg.funding_satoshis)));
+		}
+		if msg.funding_satoshis >= config.peer_channel_config_limits.max_funding_satoshis {
+			return Err(ChannelError::Close(format!("Per our config, funding must be smaller than {}. It was {}", config.peer_channel_config_limits.max_funding_satoshis, msg.funding_satoshis)));
+		}
+		if msg.funding_satoshis > TOTAL_BITCOIN_SUPPLY_SATOSHIS {
```
ariard, line 1092: Note, here the check is superior only, when above it's superior-or-equal. I don't think it changes that much in practice but..

lightning/src/util/config.rs
```diff
@@ -151,6 +156,7 @@ impl Default for ChannelHandshakeLimits {
 	fn default() -> Self {
 		ChannelHandshakeLimits {
 			min_funding_satoshis: 0,
+			max_funding_satoshis: 1 << 24,
```
ariard, line 159: Maybe you can import `MAX_FUNDING_SATOSHIS_NO_WUMBO`.

lightning/src/ln/channel.rs
```diff
@@ -850,8 +854,11 @@ impl<Signer: Sign> Channel<Signer> {
 		let holder_signer = keys_provider.get_channel_signer(false, channel_value_satoshis);
 		let pubkeys = holder_signer.pubkeys().clone();
 
-		if channel_value_satoshis >= MAX_FUNDING_SATOSHIS {
-			return Err(APIError::APIMisuseError{err: format!("funding_value must be smaller than {}, it was {}", MAX_FUNDING_SATOSHIS, channel_value_satoshis)});
+		if !their_features.supports_wumbo() && channel_value_satoshis >= MAX_FUNDING_SATOSHIS_NO_WUMBO {
```
TheBlueMatt, line 857: Lets not bother with the feature check - it doesn't change anything to add it, but adds a good chunk of additional code and constant(s) to test for.

lightning/src/ln/channel.rs
```diff
@@ -734,9 +734,13 @@ pub const COMMITMENT_TX_WEIGHT_PER_HTLC: u64 = 172;
...
-/// Maximum `funding_satoshis` value, according to the BOLT #2 specification
-/// it's 2^24.
-pub const MAX_FUNDING_SATOSHIS: u64 = 1 << 24;
+/// Maximum `funding_satoshis` value according to the BOLT #2 specification, if
+/// `option_support_large_channel` (aka wumbo channels) is not supported.
+/// It's 2^24.
+pub const MAX_FUNDING_SATOSHIS_NO_WUMBO: u64 = 1 << 24;
+
+/// Total bitcoin supply in satoshis.
+pub const TOTAL_BITCOIN_SUPPLY_SATOSHIS: u64 = 21_000_000 * 1_0000_0000;
```
maxgiraldo, line 743: Just a tiny nitpick on naming (not blocking): I feel like "total" can be misinterpreted in various ways such as representing the total supply so far and not necessarily year 2140. If the intention is to communicate the maximum number of sats possible in any one channel, then maybe a name like `MAX_SATOSHI_LIMIT` would be more accurate?

lightning/src/util/config.rs
valentinewallace: Ended up getting rid of this constant

lightning/src/ln/channel.rs
valentinewallace: "Total bitcoin supply" is referred to quite a few times in the codebase as 21 million bitcoin, so I like that it's consistent with that

TheBlueMatt (dismissed):

vincenzopalazzo (dismissed):
:sparkler:

dunxen (approved):
ACK 451e096
Compared to spec and everything looks good!

lightning/src/ln/channel.rs
```diff
@@ -6334,9 +6336,8 @@ mod tests {
 	}
 
 	#[test]
-	fn test_max_funding_satoshis() {
-		assert!(MAX_FUNDING_SATOSHIS <= 21_000_000 * 100_000_000,
-		        "MAX_FUNDING_SATOSHIS is greater than all satoshis in existence");
+	fn test_total_btc_supply_const() {
+		assert_eq!(TOTAL_BITCOIN_SUPPLY_SATOSHIS, 21_000_000 * 100_000_000);
 	}
```
jkczyz, line 6341: Test shouldn't be needed any longer.

lightning/src/ln/functional_tests.rs
```diff
@@ -58,9 +58,12 @@ use ln::chan_utils::CommitmentTransaction;
 #[test]
 fn test_insane_channel_opens() {
 	// Stand up a network of 2 nodes
+	use ln::channel::TOTAL_BITCOIN_SUPPLY_SATOSHIS;
+	let mut cfg = UserConfig::default();
+	cfg.peer_channel_config_limits.max_funding_satoshis = TOTAL_BITCOIN_SUPPLY_SATOSHIS + 1;
```
jkczyz, line 63: Is there any need to use `TOTAL_BITCOIN_SUPPLY_SATOSHIS` here? Seems strange to use an invalid value when any valid value would suffice.

lightning/src/ln/channel.rs
```diff
@@ -850,8 +849,8 @@ impl<Signer: Sign> Channel<Signer> {
 		let holder_signer = keys_provider.get_channel_signer(false, channel_value_satoshis);
 		let pubkeys = holder_signer.pubkeys().clone();
 
-		if channel_value_satoshis >= MAX_FUNDING_SATOSHIS {
-			return Err(APIError::APIMisuseError{err: format!("funding_value must be smaller than {}, it was {}", MAX_FUNDING_SATOSHIS, channel_value_satoshis)});
+		if channel_value_satoshis >= TOTAL_BITCOIN_SUPPLY_SATOSHIS {
+			return Err(APIError::APIMisuseError { err: format!("funding_value must be smaller than the total bitcoin supply, it was {}", channel_value_satoshis) });
```
jkczyz, line 853: Should we also return an error if our counterparty doesn't support wumbo but we try to open a channel >= 2^24?

lightning/src/util/config.rs
```diff
@@ -100,6 +100,13 @@ pub struct ChannelHandshakeLimits {
 	///
 	/// Default value: 0.
 	pub min_funding_satoshis: u64,
+	/// Maximum allowed satoshis when a channel is funded, this is supplied by the sender and so
+	/// only applies to inbound channels.
+	///
+	/// Default value: 2^24.
+	//  Note that this default was chosen because it was the maximum funding amount prior to wumbo
+	//  channels being introduced.
+	pub max_funding_satoshis: u64,
```
jkczyz, line 109: Why have a config parameter? If we advertise support for wumbo, wouldn't it be surprising to a peer if they couldn't open a channel >= 2^24?

lightning/src/util/config.rs
```diff
@@ -100,6 +100,13 @@ pub struct ChannelHandshakeLimits {
 	///
 	/// Default value: 0.
 	pub min_funding_satoshis: u64,
+	/// Maximum allowed satoshis when a channel is funded, this is supplied by the sender and so
```
jkczyz, line 103: nit: s/funded, this/funded. This

lightning/src/ln/channel.rs
TheBlueMatt: I don't see why we should? Its extra code that doesn't protect against anything. Not sure what it'd accomplish.

lightning/src/util/config.rs
TheBlueMatt: Users are likely to implement it anyway via the `accept_channel` hooks, plus its a really important security parameter - what level of risk are you willing to tolerate if the channel funds somehow get lost because you went offline too long?

lightning/src/ln/channel.rs
jkczyz: Mainly because it's a MUST in the spec:

```
The sending node:
- if both nodes advertised option_support_large_channel:
  - MAY set funding_satoshis greater than or equal to 2^24 satoshi.
- otherwise:
  - MUST set funding_satoshis to less than 2^24 satoshi.
```

jkczyz: Isn't this code the sending side for `open_channel`? `ChannelManager::create_channel` call `Channel::new_outbound` prior to generating `events::MessageSendEvent::SendOpenChannel`.

lightning/src/ln/functional_tests.rs
valentinewallace: It allows us to cover both the cases of `config.max_funding_satoshis` and the `TOTAL_BITCOIN_SUPPLY` check without recreating the channels with a new config. If it were lower, we'd never get to the `TOTAL_BITCOIN_SUPPLY` check I think

lightning/src/ln/channel.rs
TheBlueMatt: Oh! I'm an idiot! Yes, we should have a check on the sending side, IMO.

vincenzopalazzo (approved):

dunxen (dismissed):

lightning/src/ln/functional_tests.rs
```diff
@@ -113,6 +116,25 @@ fn test_insane_channel_opens() {
...
+	// Test that if a peer does not support wumbo channels, we'll refuse to open a wumbo channel to
+	// them.
+	use ln::channel::MAX_FUNDING_SATOSHIS_NO_WUMBO;
+	let chanmon_cfgs = create_chanmon_cfgs(2);
+	let mut node_cfgs = create_node_cfgs(2, &chanmon_cfgs);
+	node_cfgs[1].features = InitFeatures::known().clear_wumbo();
+	let node_chanmgrs = create_node_chanmgrs(2, &node_cfgs, &[None, None]);
+	let nodes = create_network(2, &node_cfgs, &node_chanmgrs);
+
+	match nodes[0].node.create_channel(nodes[1].node.get_our_node_id(), MAX_FUNDING_SATOSHIS_NO_WUMBO + 1, 0, 42, None) {
```
jkczyz, line 130: The `+ 1` should not be necessary as non-wumbo channels must be strictly less than `MAX_FUNDING_SATOSHIS_NO_WUMBO` according to BOLT 2. Maybe we should redefine the constant as `(1 << 24) - 1` and change the checks accordingly for consistency and so the error messages are accurate.

lightning/src/util/config.rs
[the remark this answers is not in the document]
valentinewallace: IMO it's unimportant, just a reasonable default.. If there's a strong preference for using the constant I can switch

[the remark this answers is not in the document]
jkczyz: Figure if we change the constant to `(1 << 24) - 1` and use `>` instead of `>=` in the checks, then the "max" naming of this field and the constant would be accurate. Currently, they are off by one. Then might as well use the constant (or another with the same value) here rather than have the new default be one more than the old default, which would occur if the above recommended changes are made.

jkczyz (dismissed):

valentinewallace:
Squashed!

jkczyz (approved):

PR #1432, opened 2022-04-19
Onion messages API 

valentinewallace:
This PR's purpose is to gather initial feedback on the proposed onion messages API, both external and internal. After garnering ACKs to crystallize the docs, structs, and method signatures, it will be closed in favor of #1503 with filled in methods and tests.

Note that the initial PR will exclude some features: sending custom TLVs (receiving custom TLVs is supported), reply paths, and sending to blinded routes.

In a few cases, I included some small code-level changes if I thought they made it more obvious why an API change is being suggested.


lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,61 @@
+/// The payload of an inbound onion message.
+pub struct Payload {
+	/// Onion messages may contain custom TLVs for the recipient.
+	custom_tlvs: Vec<CustomTlv>,
+	// Coming soon: path_id, reply_path
+}
+
+/// A sender, receiver and forwarder of onion messages. In upcoming releases, this object will be
+/// used to retrieve invoices and fulfill invoice requests from offers.
+pub struct OnionMessager<Signer: Sign, K: Deref>
```
TheBlueMatt, line 10: Should this be `OnionMessager` or `InvoiceOnionMessager`?

lightning/src/ln/msgs.rs
```diff
@@ -909,6 +920,13 @@ pub trait RoutingMessageHandler : MessageSendEventsProvider {
 	fn handle_query_short_channel_ids(&self, their_node_id: &PublicKey, msg: QueryShortChannelIds) -> Result<(), LightningError>;
 }
 
+/// A trait to describe an object that can receive onion messages.
+pub trait OnionMessageHandler {
```
TheBlueMatt, line 924: Doesn't this need to be `: MessageSendEventsProvider`?

lightning/src/ln/onion_utils.rs
```diff
@@ -199,6 +231,8 @@ pub(super) fn route_size_insane(payloads: &Vec<msgs::OnionHopData>) -> bool {
 }
 
 /// panics if route_size_insane(paylods)
+// NOTE: I believe since onion messages can be variable size, we'll have to remove the panic
+// mentioned above^
```
TheBlueMatt, line 235: `route_size_insane` handles that just fine, you just have to check it at the callsite before calling this.

lightning/src/util/ser_macros.rs
```diff
@@ -165,6 +165,10 @@ macro_rules! decode_tlv {
 	}};
 }
 
+/// Decode a TLV stream that contains custom TLVs that are unknown to LDK but may be known to the
+/// user.
+macro_rules! decode_tlv_stream_with_custom {}
```
TheBlueMatt, line 170: Just to check - because onion message TLVs are split into "control types" and "message types", this is split, too? ie it takes a list of TLVs to decode like we normally decode TLVs, and then also reads anything higher than a certain constant as "custom TLVs".

lightning/src/ln/onion_utils.rs
```diff
@@ -506,15 +549,25 @@ pub(super) fn process_onion_failure<T: secp256k1::Signing, L: Deref>(secp_ctx: &
...
+	/// This payload was for an incoming onion message.
+	Message(msgs::OnionMsgPayload),
+}
+
 /// Data decrypted from the onion payload.
-pub(crate) enum Hop {
-	/// This onion payload was for us, not for forwarding to a next-hop. Contains information for
-	/// verifying the incoming payment.
-	Receive(msgs::OnionHopData),
+pub(crate) enum HopPayload {
```
TheBlueMatt, line 563: This feels super awkward. Now the caller of `decode_next_hop` (which knows the type of onion it expects) has to check that the received onion type matches what it expects? Maybe lets add trivial wrapper functions here and make `decode_next_hop` private so that the caller can still say "decode(payment_hash)" and always get back a payment onion.

lightning/src/util/ser_macros.rs
TheBlueMatt: Also, is it at all practical to have some kind of matching on the onion type and use a different decoder per-type? Would be really nice to not have to allocate a vec of TLV decode entries for onion message decodes we implement ourselves.

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,69 @@
...
+	where K::Target: KeysInterface<Signer = Signer>,
+{
+	/// Constructs a new `OnionMessager` to send, forward, and receive onion messages.
+	pub fn new(keys_manager: K) -> Self {}
+
+	/// Send an empty onion message to `recipient`, routing it through `intermediate_nodes`.
+	///
+	/// LDK will not make peer connections for you, so you must be connected to the first hop of
+	/// `intermediate_nodes` for sending to work.
+	// NOTE: sending custom TLVs, reply paths, and sending to blinded routes aren't included atm
```
TheBlueMatt, line 47: Are blinded routes going to end up having to redo the onion_utils changes anyway? Like, is implementing blinded routes going to mean all the work put into ensuring we re-use the same code here going to end up getting thrown away?

lightning/src/ln/msgs.rs
```diff
@@ -724,6 +724,17 @@ enum EncodingType {
 	Uncompressed = 0x00,
 }
 
+/// An onion message to be sent or received from a peer
+#[derive(Clone, Debug, PartialEq)]
+pub struct OnionMessage {
+	/// The blinding point used in the shared secret that is used to decrypt the onion message's
+	/// `encrypted_data` field.
+	pub blinding_point: PublicKey,
```
TheBlueMatt, line 732: Does this need to be a `Result<PublicKey...`? ie if we receive an onion message on the wire we shouldn't disconnect our peer just because the point is invalid (I mean our peers shouldnt do it, but still).

TheBlueMatt (commented):
You know I'm gonna complain if I find any cases where we can drop a Vec and replace it without allocatiing. May be hard in a few cases but I see a lot of Vecs in message handling here :(

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl<Signer: Sign, K: Deref> OnionMessager<Signer, K>
+	where K::Target: KeysInterface<Signer = Signer>,
+{
+	/// Constructs a new `OnionMessager` to send, forward, and receive onion messages.
+	pub fn new(keys_manager: K) -> Self {}
+
+	/// Send an empty onion message to `recipient`, routing it through `intermediate_nodes`.
+	///
+	/// LDK will not make peer connections for you, so you must be connected to the first hop of
```
ariard, line 45: That sounds weird at first but we're doing the same with channel establishment. So we expect a user to fetch the first `PublicKey` of intermediate node, query `NetworkGraph` to get a `NetAddress` and establish a connection ? Though I expect if the user failed to do so we're going to fail fast and return a `APIError::ConnectionUnavailable`.

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,69 @@
+/// The payload of an inbound onion message.
+pub struct Payload {
```
ariard, line 2: Maybe prefix names with `Inbound` ?

lightning/src/ln/onion_message.rs
valentinewallace: >  I guess its more of an intermediary struct that handles p2p onion messages, decrypts them, then hands them off to various other handler traits that actually handle the messages?

That'd make sense to me. For inbound OMs, we'd have an `InvoiceRequestHandler` (this would be another object that marshalls an invoice and gives it back to the `OnionMessager`) and an `InvoiceHandler` (this would probably be `ChannelManager` under the hood, to pay the invoice?) And a `CustomHandler` for custom onion messages. 

For outbound InvoiceRequests, `OnionMessager` could have a `request_invoice` method?

lightning/src/ln/onion_message.rs
valentinewallace: Now this struct is for inbound and outbound :)

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,96 @@
...
+pub enum Destination {
+	/// We're sending this onion message to a node.
+	Node(PublicKey),
+	/// We're sending this onion message to a blinded route.
+	BlindedRoute(BlindedRoute),
+}
+
+/// A sender, receiver and forwarder of onion messages. In upcoming releases, this object will be
+/// used to retrieve invoices and fulfill invoice requests from offers.
+pub struct OnionMessager<Signer: Sign, K: Deref>
```
jkczyz, line 124: How about either `OnionMessaging` or `OnionMessenger`? "messager" is French for "messenger" lol

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,96 @@
...
+	pub blinded_hops: Vec<BlindedNode>,
+}
+
+impl BlindedRoute {
+	/// Create a blinded route to be forwarded along `hops`.
+	pub fn new<Signer: Sign, K: Deref>(hops: Vec<PublicKey>, keys_manager: K) -> Result<Self, secp256k1::Error>
+		where K::Target: KeysInterface<Signer = Signer> {}
+}
+
+/// The final destination of an onion message.
```
jkczyz, line 114: nit: drop "final"?

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,96 @@
...
+// /// `InvoiceHandler` handles incoming invoice requests, invoices, and invoice errors.
+// pub trait InvoiceHandler {
+//   /// Handles an incoming invoice request.
+//   fn handle_invoice_request(invoice_request: InvoiceRequest) -> Result<Invoice, ()>;
+//   // Handles an incoming invoice.
+//   fn handle_invoice(invoice: Invoice) -> Result<(), ()>;
+//   // Handles an incoming invoice error.
+//   fn handle_invoice_error(invoice_error: InvoiceError);
+//   // Sends a payment.
+//   fn send_payment(..);
```
jkczyz, line 144: High-level questions: What do you expect the parameters and intended use to be for this? Would `handle_invoice` call this?

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,96 @@
...
+impl<Signer: Sign, K: Deref> OnionMessager<Signer, K>
+	where K::Target: KeysInterface<Signer = Signer>,
+{
+	/// Constructs a new `OnionMessager` to send, forward, and delegate received onion messages to
+	/// their respective handlers.
+	pub fn new(keys_manager: K) -> Self {}
+
+	/// Send an empty onion message to `recipient`, routing it through `intermediate_nodes`.
+	// NOTE: sending custom TLVs and reply paths aren't included atm
+	pub fn send_onion_message(&self, recipient: Destination, intermediate_nodes: Vec<PublicKey>) -> Result<(), APIError> {}
```
jkczyz: I suppose routing would play a role in determining `intermediate_nodes`. Would it be a single-path route and do you need more than the `PublicKey` of each hop?

lightning/src/ln/onion_message.rs
```diff
@@ -154,6 +163,9 @@ impl<Signer: Sign, K: Deref> OnionMessager<Signer, K>
 
 impl OnionMessageHandler for OnionMessager {
 	fn handle_onion_message(&self, peer_node_id: &PublicKey, msg: &msgs::OnionMessage) {
+		// calls:
+		// * onion_utils::decode_next_message_hop
+		// * onion_utils::next_hop_packet_pubkey
 	}
```
jkczyz, line 169: I take it users would typically just implement `InvoiceHandler` but we're giving the option to implement `OnionMessageHandler` to handle custom onion messages? Should we have a separate hook for that?

Also, is there a case where users would need custom logic outside of handling custom messages? Just wondering if it makes sense for users to ever do the decoding / forwarding. Maybe this akin to `ChannelMessageHandler` and `ChannelManager`, where the latter is almost always used.

lightning/src/ln/onion_message.rs
valentinewallace: Yeah, single-path route. Only the `PublicKey` is needed!

lightning/src/ln/onion_message.rs
valentinewallace: I love OnionMessenger lol, will switch to that if that sounds good

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,200 @@
...
+/// by outside observers and thus can be used to hide the identity of the recipient.
+pub struct BlindedNode {
+	/// The blinded node id of this hop in a blinded route.
+	pub blinded_node_id: PublicKey,
+	/// The encrypted payload intended for this hop in a blinded route.
+	// If we're sending to this blinded route, this payload will later be encoded into the
+	// [`EncryptedTlvs`] for the hop when constructing the onion packet for sending.
+	//
+	// [`EncryptedTlvs`]: EncryptedTlvs
+	pub encrypted_payload: Vec<u8>,
```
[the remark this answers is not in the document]
valentinewallace: These are supplied by the receiver, as the encrypted blobs of the blinded route, so they'll always be plain encrypted bytes I think?

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,200 @@
...
+		// calls:
+		// * ChaCha20Poly1305RFC::decrypt_in_place
+	}
+}
+
+// Coming soon:
+// enum Message {
+// 	InvoiceRequest(InvoiceRequest),
+// 	Invoice(Invoice),
+//	InvoiceError(InvoiceError),
```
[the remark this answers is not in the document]
valentinewallace: Does the `T` mean only one type of custom message will be allowed? Confused what that is

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,96 @@
...
+	/// message's next hop and forward it along.
+	///
+	/// [`encrypted_payload`]: BlindedNode::encrypted_payload
+	pub introduction_node_id: PublicKey,
+	/// Creators of blinded routes supply the introduction node id's `blinding_point`, which the
+	/// introduction node will use in decrypting its [`encrypted_payload`] to forward the onion
+	/// message.
+	///
+	/// [`encrypted_payload`]: BlindedNode::encrypted_payload
+	pub blinding_point: PublicKey,
```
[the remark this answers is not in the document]
valentinewallace: No, the blinding_point is used as part of generating the shared secret that decrypts the `EncryptedTlvs`. The `blinded_node_id` is the hop's node_id multiplied by a blinding factor

lightning/src/ln/onion_message.rs
valentinewallace: This would take `intermediate_ndoes`, a `Destination`, and an amount. It would then send an `invoice_request` onion message (OM) to the destination, who would respond with an `invoice` OM, and then pay the invoice probably using an `InvoicePayer` under the hood.

lightning/src/ln/onion_message.rs
valentinewallace: > Maybe this akin to ChannelMessageHandler and ChannelManager, where the latter is almost always used.

I think this is the case. I don't think users would want to do the decoding themselves, seems likely they'd only want to implement handling of custom onion messages if anything

lightning/src/ln/onion_message.rs
valentinewallace: Okay, so IIUC `OnionMessenger` will have a pointer to an `InvoiceHandler` that has a `send_payment` method that sends out an `invoice_request` and pays the invoice. I'm wondering, how does this `InvoiceHandler` communicate to the `OnionMessenger` when it has an invoice request it wants to send out?

lightning/src/ln/onion_message.rs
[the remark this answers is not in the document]
valentinewallace: I think their size can vary a lot especially with custom tlvs shoved in there :(

valentinewallace:
Updated to the latest API/struct changes and reworked the commit history

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,264 @@
...
+	// * `message: Message` field
+	// * `reply_path: Option<BlindedRoute>` field
+}
+
+// Coming soon:
+// enum Message {
+// 	InvoiceRequest(InvoiceRequest),
+// 	Invoice(Invoice),
+//	InvoiceError(InvoiceError),
+//	CustomMessage<T>,
```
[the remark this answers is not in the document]
valentinewallace: Similar to the usages of `onion_utils::` and `msgs::` across the codebase, the intent is for usages external to the `onion_message` module to utilize the `onion_message::` prefix, e.g. `onion_message::Packet`. Does this address this comment and the below?

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,264 @@
...
+	where K::Target: KeysInterface<Signer = Signer>,
+				L::Target: Logger,
+{
+	keys_manager: K,
+	logger: L,
+	pending_msg_events: Mutex<Vec<MessageSendEvent>>,
+	secp_ctx: Secp256k1<secp256k1::All>,
+	// Coming soon:
+	// invoice_handler: InvoiceHandler,
+	// custom_handler: CustomHandler, // handles custom onion messages
```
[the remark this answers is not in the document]
valentinewallace: Without thinking too hard about the details, the high level concept of using OMs to relay an on-chain transaction to a tx broadcaster seems reasonable to me. However, unless the OMs are providing chain data too, it seems you'll still need a chain source, and that chain source is likely to provide broadcasting anyway?

lightning/src/ln/onion_message.rs
```diff
@@ -8,3 +8,95 @@
...
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), io::Error> {}
+}
+
+impl Readable for ControlTlvs {
+	fn read<R: Read>(mut r: &mut R) -> Result<Self, DecodeError> {}
+}
+
+/// Used to construct the blinded hops portion of a blinded route. These hops cannot be identified
+/// by outside observers and thus can be used to hide the identity of the recipient.
+pub struct BlindedNode {
```
jkczyz, line 123: Could we call this `BlindedHop` to be consistent with our `Route` / `RouteHop` terminology?

lightning/src/ln/onion_message.rs
```diff
@@ -0,0 +1,264 @@
...
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Onion Messages: sending, receiving, forwarding, and ancillary utilities live here
+
+pub(crate) const SMALL_PACKET_HOP_DATA_LEN: usize = 1300;
+pub(crate) const BIG_PACKET_HOP_DATA_LEN: usize = 32768;
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct Packet {
```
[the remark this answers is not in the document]
jkczyz: I'm somewhat indifferent on this. Have heard arguments both ways. Internally, there is no conflict. Externally, you could always use `as` to rename the import or qualify with the module name, FWIW. Probably fine as is.

lightning/src/ln/onion_message.rs
```diff
@@ -9,6 +9,25 @@
 
 //! Onion Messages: sending, receiving, forwarding, and ancillary utilities live here
 
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct Packet {
+	pub(crate) version: u8,
+	/// We don't want to disconnect a peer just because they provide a bogus public key, so we hold a
+	/// Result instead of a PublicKey as we'd like.
+	pub(crate) public_key: Result<PublicKey, secp256k1::Error>,
```
jkczyz, line 20: Haven't looked at the full context of where this is constructed / used, but could that return a `Err` if the public key is bad?

jkczyz (commented):
Concept ACK. Will move on to reviewing #1518 and #1503.

For documentation, I'd say having the bulk it as module-level docs for `onion_message` describing the public structs and use by other modules, cross-referencing as needed, should be fine.

lightning/src/ln/onion_message.rs
```diff
@@ -8,3 +8,95 @@
 // licenses.
 
 //! Onion Messages: sending, receiving, forwarding, and ancillary utilities live here
+
+/// Onion messages have "control" TLVs and "data" TLVs. Control TLVs are used to control the
+/// direction and routing of an onion message from hop to hop, whereas data TLVs contain the onion
+/// message content itself.
+pub(crate) enum ControlTlvs {
```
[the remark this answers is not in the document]
valentinewallace: Hmm, I'm a bit confused which ones aren't already outside-in? Packet>Payload>>EncryptedTlvs>ControlTlvs is outside-in

lightning/src/ln/onion_message.rs
valentinewallace: Here's the context: https://github.com/lightningdevkit/rust-lightning/pull/1432#discussion_r854129322 lmk if more clarification is desired

lightning/src/ln/onion_message.rs
```diff
@@ -8,3 +8,95 @@
...
+#[allow(unused_assignments)]
+#[inline]
+fn construct_keys_callback<T: secp256k1::Signing + secp256k1::Verification, FType: FnMut(PublicKey, SharedSecret, [u8; 32], PublicKey, SharedSecret)> (secp_ctx: &Secp256k1<T>, unblinded_path: &Vec<PublicKey>, session_priv: &SecretKey, mut callback: FType) -> Result<(), secp256k1::Error> {}
+
+/// Construct keys for constructing a blinded route along the given `unblinded_path`.
+///
+/// Returns: `(encrypted_tlvs_keys, blinded_node_ids)`
+/// where the encrypted tlvs keys are used to encrypt the blinded route's blinded payloads and the
+/// blinded node ids are used to set the [`blinded_node_id`]s of the [`BlindedRoute`].
+fn construct_blinded_route_keys<T: secp256k1::Signing + secp256k1::Verification>(
```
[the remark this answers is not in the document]
jkczyz: By "double", I meant "Construct" and "constructing" in the documentation (not across types). So I'd say used `create_` in both names. But feel free to keep as-is if `construct_` is common already. Figured `create_` is fairly standard but we probably haven't been consistent.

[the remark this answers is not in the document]
valentinewallace: Ah, these are consistent with the analogous `construct_*_keys` method names in `onion_utils`. Searching the codebase, it seems that `construct_*` tends to be used for creating internal intermediate structs, and `create_*` tends to be used for creating user-facing things like channels and invoices. 

Rephrasing to "Construct keys for creating a blinded route" for now, but no strong opinion

valentinewallace:
I believe I've incorporated everything except https://github.com/lightningdevkit/rust-lightning/pull/1432#pullrequestreview-1009664596 on #1503, so closing this and marking 1503 as ready for review. Thanks for all the conceptual review here.

PR #1433, opened 2022-04-19
Allow indication to BackgroundProcessor that graph sync is pending

arik-so:
When running a graph sync, the network graph should not be modified from other sources until graph sync completes to avoid inconsistencies or developer/user confusion. Similarly, the network graph should not be prematurely pruned, as graph sync allows very large timestamp deltas in graph snapshot comparisons.

To that end, I add an argument to the BackgroundProcessor to indicate that network-graph-related processed should be avoided pending the graph sync completion, which can be signaled by calling a new method.

This PR depends on #1155. Thanks to @jkczyz for the logic and idiomaticness improvement suggestions!

### DIFF
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -79,6 +79,11 @@ const PING_TIMER: u64 = 1;
 /// Prune the network graph of stale entries hourly.
 const NETWORK_PRUNE_TIMER: u64 = 60 * 60;
 
+#[cfg(all(not(test), debug_assertions))]
+const SCORER_PERSIST_TIMER: u64 = 30;
+#[cfg(test)]
+const SCORER_PERSIST_TIMER: u64 = 1;
+
 #[cfg(not(test))]
 const FIRST_NETWORK_PRUNE_TIMER: u64 = 60;
 #[cfg(test)]
@@ -214,6 +219,7 @@ impl BackgroundProcessor {
 			let mut last_freshness_call = Instant::now();
 			let mut last_ping_call = Instant::now();
 			let mut last_prune_call = Instant::now();
+			let mut last_scorer_persist_call = Instant::now();
 			let mut have_pruned = false;
 
 			loop {
@@ -307,12 +313,16 @@ impl BackgroundProcessor {
 					} else {
 						log_trace!(logger, "Not pruning network graph, either due to pending rapid gossip sync or absence of a prunable graph.");
 					}
+				}
+
+				if last_scorer_persist_call.elapsed().as_secs() > SCORER_PERSIST_TIMER {
 					if let Some(ref scorer) = scorer {
 						log_trace!(logger, "Persisting scorer");
 						if let Err(e) = persister.persist_scorer(&scorer) {
 							log_error!(logger, "Error: Failed to persist scorer, check your disk and permissions {}", e)
 						}
 					}
+					last_scorer_persist_call = Instant::now();
 				}
 			}
 
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -9,6 +9,7 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 #[macro_use] extern crate lightning;
+extern crate lightning_rapid_gossip_sync;
 
 use lightning::chain;
 use lightning::chain::chaininterface::{BroadcasterInterface, FeeEstimator};
@@ -22,6 +23,7 @@ use lightning::routing::scoring::WriteableScore;
 use lightning::util::events::{Event, EventHandler, EventsProvider};
 use lightning::util::logger::Logger;
 use lightning::util::persist::Persister;
+use lightning_rapid_gossip_sync::RapidGossipSync;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::thread;
@@ -142,6 +144,12 @@ impl BackgroundProcessor {
 	/// functionality implemented by other handlers.
 	/// * [`NetGraphMsgHandler`] if given will update the [`NetworkGraph`] based on payment failures.
 	///
+	/// # Rapid Gossip Sync
+	///
+	/// If rapid gossip sync is meant to run at startup, pass an optional [`RapidGossipSync`]
+	/// to `rapid_gossip_sync` to indicate to [`BackgroundProcessor`] not to prune the
+	/// [`NetworkGraph`] instance until the [`RapidGossipSync`] instance completes its first sync.
+	///
 	/// [top-level documentation]: BackgroundProcessor
 	/// [`join`]: Self::join
 	/// [`stop`]: Self::stop
@@ -175,9 +183,11 @@ impl BackgroundProcessor {
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
 		S: 'static + Deref<Target = SC> + Send + Sync,
 		SC: WriteableScore<'a>,
+		RGS: 'static + Deref<Target = RapidGossipSync<G>> + Send
 	>(
 		persister: PS, event_handler: EH, chain_monitor: M, channel_manager: CM,
-		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L, scorer: Option<S>
+		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L, scorer: Option<S>,
+		rapid_gossip_sync: Option<RGS>
 	) -> Self
 	where
 		CA::Target: 'static + chain::Access,
@@ -272,12 +282,30 @@ impl BackgroundProcessor {
 				// pruning their network graph. We run once 60 seconds after startup before
 				// continuing our normal cadence.
 				if last_prune_call.elapsed().as_secs() > if have_pruned { NETWORK_PRUNE_TIMER } else { FIRST_NETWORK_PRUNE_TIMER } {
-					if let Some(ref handler) = net_graph_msg_handler {
-						log_trace!(logger, "Pruning network graph of stale entries");
-						handler.network_graph().remove_stale_channels();
-						if let Err(e) = persister.persist_graph(handler.network_graph()) {
+					// The network graph must not be pruned while rapid sync completion is pending
+					log_trace!(logger, "Assessing prunability of network graph");
+					let graph_to_prune = match rapid_gossip_sync.as_ref() {
+						Some(rapid_sync) => {
+							if rapid_sync.is_initial_sync_complete() {
+								Some(rapid_sync.network_graph())
+							} else {
+								None
+							}
+						},
+						None => net_graph_msg_handler.as_ref().map(|handler| handler.network_graph())
+					};
+
+					if let Some(network_graph_reference) = graph_to_prune {
+						network_graph_reference.remove_stale_channels();
+
+						if let Err(e) = persister.persist_graph(network_graph_reference) {
 							log_error!(logger, "Error: Failed to persist network graph, check your disk and permissions {}", e)
 						}
+
+						last_prune_call = Instant::now();
+						have_pruned = true;
+					} else {
+						log_trace!(logger, "Not pruning network graph, either due to pending rapid gossip sync or absence of a prunable graph.");
 					}
 					if let Some(ref scorer) = scorer {
 						log_trace!(logger, "Persisting scorer");
@@ -285,9 +313,6 @@ impl BackgroundProcessor {
 							log_error!(logger, "Error: Failed to persist scorer, check your disk and permissions {}", e)
 						}
 					}
-
-					last_prune_call = Instant::now();
-					have_pruned = true;
 				}
 			}
 
@@ -370,7 +395,7 @@ mod tests {
 	use lightning::chain::transaction::OutPoint;
 	use lightning::get_event_msg;
 	use lightning::ln::channelmanager::{BREAKDOWN_TIMEOUT, ChainParameters, ChannelManager, SimpleArcChannelManager};
-	use lightning::ln::features::InitFeatures;
+	use lightning::ln::features::{ChannelFeatures, InitFeatures};
 	use lightning::ln::msgs::{ChannelMessageHandler, Init};
 	use lightning::ln::peer_handler::{PeerManager, MessageHandler, SocketDescriptor, IgnoringMessageHandler};
 	use lightning::routing::network_graph::{NetworkGraph, NetGraphMsgHandler};
@@ -385,8 +410,10 @@ mod tests {
 	use std::fs;
 	use std::path::PathBuf;
 	use std::sync::{Arc, Mutex};
+	use std::sync::mpsc::SyncSender;
 	use std::time::Duration;
 	use lightning::routing::scoring::{FixedPenaltyScorer};
+	use lightning_rapid_gossip_sync::RapidGossipSync;
 	use super::{BackgroundProcessor, FRESHNESS_TIMER};
 
 	const EVENT_DEADLINE: u64 = 5 * FRESHNESS_TIMER;
@@ -414,6 +441,7 @@ mod tests {
 		logger: Arc<test_utils::TestLogger>,
 		best_block: BestBlock,
 		scorer: Arc<Mutex<FixedPenaltyScorer>>,
+		rapid_gossip_sync: Option<Arc<RapidGossipSync<Arc<NetworkGraph>>>>
 	}
 
 	impl Drop for Node {
@@ -428,6 +456,7 @@ mod tests {
 
 	struct Persister {
 		graph_error: Option<(std::io::ErrorKind, &'static str)>,
+		graph_persistence_notifier: Option<SyncSender<()>>,
 		manager_error: Option<(std::io::ErrorKind, &'static str)>,
 		scorer_error: Option<(std::io::ErrorKind, &'static str)>,
 		filesystem_persister: FilesystemPersister,
@@ -436,11 +465,15 @@ mod tests {
 	impl Persister {
 		fn new(data_dir: String) -> Self {
 			let filesystem_persister = FilesystemPersister::new(data_dir.clone());
-			Self { graph_error: None, manager_error: None, scorer_error: None, filesystem_persister }
+			Self { graph_error: None, graph_persistence_notifier: None, manager_error: None, scorer_error: None, filesystem_persister }
 		}
 
 		fn with_graph_error(self, error: std::io::ErrorKind, message: &'static str) -> Self {
 			Self { graph_error: Some((error, message)), ..self }
+		}
+
+		fn with_graph_persistence_notifier(self, sender: SyncSender<()>) -> Self {
+			Self { graph_persistence_notifier: Some(sender), ..self }
 		}
 
 		fn with_manager_error(self, error: std::io::ErrorKind, message: &'static str) -> Self {
@@ -461,6 +494,10 @@ mod tests {
 			}
 
 			if key == "network_graph" {
+				if let Some(sender) = &self.graph_persistence_notifier {
+					sender.send(()).unwrap();
+				};
+
 				if let Some((error, message)) = self.graph_error {
 					return Err(std::io::Error::new(error, message))
 				}
@@ -504,7 +541,8 @@ mod tests {
 			let msg_handler = MessageHandler { chan_handler: Arc::new(test_utils::TestChannelMessageHandler::new()), route_handler: Arc::new(test_utils::TestRoutingMessageHandler::new() )};
 			let peer_manager = Arc::new(PeerManager::new(msg_handler, keys_manager.get_node_secret(Recipient::Node).unwrap(), &seed, logger.clone(), IgnoringMessageHandler{}));
 			let scorer = Arc::new(Mutex::new(test_utils::TestScorer::with_penalty(0)));
-			let node = Node { node: manager, net_graph_msg_handler, peer_manager, chain_monitor, persister, tx_broadcaster, network_graph, logger, best_block, scorer };
+			let rapid_gossip_sync = None;
+			let node = Node { node: manager, net_graph_msg_handler, peer_manager, chain_monitor, persister, tx_broadcaster, network_graph, logger, best_block, scorer, rapid_gossip_sync };
 			nodes.push(node);
 		}
 
@@ -602,7 +640,7 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(),  nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 
 		macro_rules! check_persisted_data {
 			($node: expr, $filepath: expr) => {
@@ -667,7 +705,7 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 		loop {
 			let log_entries = nodes[0].logger.lines.lock().unwrap();
 			let desired_log = "Calling ChannelManager's timer_tick_occurred".to_string();
@@ -690,7 +728,7 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir).with_manager_error(std::io::ErrorKind::Other, "test"));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 		match bg_processor.join() {
 			Ok(_) => panic!("Expected error persisting manager"),
 			Err(e) => {
@@ -707,7 +745,7 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir).with_graph_error(std::io::ErrorKind::Other, "test"));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 
 		match bg_processor.stop() {
 			Ok(_) => panic!("Expected error persisting network graph"),
@@ -725,7 +763,7 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir).with_scorer_error(std::io::ErrorKind::Other, "test"));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(),  nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 
 		match bg_processor.stop() {
 			Ok(_) => panic!("Expected error persisting scorer"),
@@ -748,7 +786,7 @@ mod tests {
 		let event_handler = move |event: &Event| {
 			sender.send(handle_funding_generation_ready!(event, channel_value)).unwrap();
 		};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 
 		// Open a channel and check that the FundingGenerationReady event was handled.
 		begin_open_channel!(nodes[0], nodes[1], channel_value);
@@ -773,7 +811,7 @@ mod tests {
 		let (sender, receiver) = std::sync::mpsc::sync_channel(1);
 		let event_handler = move |event: &Event| sender.send(event.clone()).unwrap();
 		let persister = Arc::new(Persister::new(data_dir));
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
 
 		// Force close the channel and check that the SpendableOutputs event was handled.
 		nodes[0].node.force_close_channel(&nodes[0].node.list_channels()[0].channel_id, &nodes[1].node.get_our_node_id()).unwrap();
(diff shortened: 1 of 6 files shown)


lightning-background-processor/src/lib.rs
```diff
@@ -217,15 +220,21 @@ impl BackgroundProcessor {
...
 
 			loop {
+				// If we're awaiting graph sync to be completed, we do not want to modify the network graph
+				let is_currently_awaiting_graph_sync = is_awaiting_graph_sync_completion_clone.load(Ordering::Acquire);
+
 				peer_manager.process_events(); // Note that this may block on ChannelManager's locking
-				channel_manager.process_pending_events(&event_handler);
-				chain_monitor.process_pending_events(&event_handler);
+				if !is_currently_awaiting_graph_sync {
+					channel_manager.process_pending_events(&event_handler);
```
TheBlueMatt, line 228: Hmm, we definitely cant skip processing events while we're waiting on a download, I think - does this cause problems?

lightning-background-processor/src/lib.rs
```diff
@@ -324,6 +336,13 @@ impl BackgroundProcessor {
 		self.stop_and_join_thread()
 	}
 
+	/// Signal to `BackgroundProcessor` that graph sync has completed
+	/// This function can only be called usefully once, so there is an implicit understanding
+	/// that running graph sync multiple times after startup is API misuse.
```
TheBlueMatt, line 341: This is confusingly worded, I think - if a client application stays open for a few hours, they should totally keep their graph up to date whereas this reads to me as "you should only sync the graph once at start". Maybe we should name the function `initial_graph_sync_complete` or so?

lightning-background-processor/src/lib.rs
```diff
@@ -188,7 +189,7 @@ impl BackgroundProcessor {
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
 	>(
 		persister: PS, event_handler: EH, chain_monitor: M, channel_manager: CM,
-		net_graph_msg_handler: Option<NG>, peer_manager: PM, logger: L
+		net_graph_msg_handler: Option<NG>, peer_manager: PM, await_graph_sync_completion: bool, logger: L
```
TheBlueMatt, line 192: Need method-level docs that explains what the graph sync bool means - something about not pruning.

TheBlueMatt (commented):
Needs docs, but, yea, I guess this is probably where we should go. Also needs docs in `NetworkGraph::remove_stale_channels`.

lightning-background-processor/src/lib.rs
jkczyz: `ProbabilisticScorer` will take a read lock on `NetworkGraph`, which would block, IIUC.

TheBlueMatt: Is that an issue? The update processing logic should take a few hundred ms, which is fine for the event processing time, I'd assumed most of the "awaiting" time here is just "waiting on download to complete", or was hoping so.

arik-so: The download + graph syncing time should indeed only take a few hundred ms max, but it needs a write lock to the network graph. So given the low time constraint, the delay seemed to be an acceptable trade-off.

jkczyz: Actually, it looks like a write lock is taken per update, not overall, so maybe event processing would be fine?

arik-so: You mean by the graph sync process itself? That's true.

arik-so: My concern though remains that I don't want any of the handled events to trigger a change to the network graph, which definitely seems like a possibility. I guess an alternative would be propagating the awaiting boolean into the DecoratingEventHandler and filtering network events right there.

jkczyz: That gets tricky because they need to be handled before we persist. If you really don't want it to change I think it would make more sense to take a write lock at the start of processing. That will block any event handling needing the lock, though.

arik-so: If network graph changes need to be handled before we can persist non-graph-related changes such as channel monitors, that once again brings us back to the point we're at right now, where we need to delay processing until graph sync completes, which is not an issue except in high-latency conditions.

TheBlueMatt: I don't really get why handling things which change the network graph before or after the initial sync changes anything? Like, we can always get HTLC failures that result in the network graph being changed, if that's going to throw off the next sync then ithat's a problem with the sync, not with the events. It'll either throw off the current sync or the next one, not much we can do about that.

arik-so: In that case, do we even really need this PR? Considering that the graph will settle into something reasonable quickly anyway, if a temporary partial sync disruption is acceptable, it might be a NO_OP.

TheBlueMatt: Yes - the concern is mass removal based on timeouts when we just have a stale timer for all our data. The concern over events is just a concern over removing things from the graph based on HTLC failures - that's a totally legit reason to remove something from the graph, even if the sync logic ends up with a duplicate removal later.

arik-so: In that case, I think the easiest path forward is modifying this PR to merely delay pruning the graph. Frankly, I think it's already a non-issue to begin with, considering the pruning starts only a minute in.

TheBlueMatt: Its not super likely to be an issue, sure, but its mobile - "shouldnt happen as long as the internet kinda works and can manage to fetch a file within a minute" is.....a dubious assumption :(

arik-so: Agreed. I think we might benefit from a battery that can detect faulty internet connectivity, because in high latency situations, the other events would take a while to show up, too, despite there perhaps already being a push notification. And user confusion/perceived loss/non-receipt of funds is the argument you mentioned above for not delaying event processing.

lightning-background-processor/src/lib.rs
```diff
@@ -764,34 +758,44 @@ mod tests {
...
 		};
 		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), true, nodes[0].logger.clone());
 
-		// Open a channel and check that the FundingGenerationReady event was handled.
-		begin_open_channel!(nodes[0], nodes[1], channel_value);
-		let reception_result = receiver
-			.recv_timeout(Duration::from_secs(EVENT_DEADLINE));
-		assert!(reception_result.is_err());
+		for i in 0..EVENT_DEADLINE {
+			std::thread::sleep(Duration::from_secs(1));
```
jkczyz, line 779: Hmm... we should avoid sleeps in tests if at all possible. I'd imagine you could have a persister that signals completion using `std::sync::mpsc::sync_channel` similar to the event handler above.

lightning-background-processor/src/lib.rs
```diff
@@ -764,34 +758,44 @@ mod tests {
...
-			.recv_timeout(Duration::from_secs(EVENT_DEADLINE))
-			.expect("FundingGenerationReady not handled within deadline");
 
-		end_open_channel!(nodes[0], nodes[1], temporary_channel_id, funding_tx);
-
-		assert!(bg_processor.stop().is_ok());
+		let mut seconds_slept = 0;
+		loop {
+			assert!(seconds_slept < 3);
+			std::thread::sleep(Duration::from_secs(1));
```
jkczyz, line 789: Likewise here.

lightning-background-processor/src/lib.rs
```diff
@@ -764,34 +758,44 @@ mod tests {
...
+		let network_graph = nodes[0].network_graph.clone();
+		let features = ChannelFeatures::empty();
+		network_graph.update_channel_from_partial_announcement(42, 53, features, nodes[0].node.get_our_node_id(), nodes[1].node.get_our_node_id());
+		let original_graph_description = network_graph.to_string();
+		assert!(original_graph_description.contains("42: features: 0000, node_one:"));
 
-		// Set up a background event handler for FundingGenerationReady events.
 		let (sender, receiver) = std::sync::mpsc::sync_channel(1);
 		let event_handler = move |event: &Event| {
 			sender.send(handle_funding_generation_ready!(event, channel_value)).unwrap();
```
jkczyz, line 774: You can use an empty event handler since you don't need to test that logic here.

arik-so: Oops, good point.

lightning-background-processor/src/lib.rs
arik-so: Unfortunately, I can't really because the network pruning is not calling the event handler. All it does is call a method on the network graph without informing anybody. It's especially tricky because I wanna make sure that more time has passed than FIRST_NETWORK_PRUNE_TIMER.

lightning-background-processor/src/lib.rs
```diff
@@ -224,8 +227,7 @@ impl BackgroundProcessor {
 				// We wait up to 100ms, but track how long it takes to detect being put to sleep,
 				// see `await_start`'s use below.
 				let await_start = Instant::now();
-				let updates_available =
-					channel_manager.await_persistable_update_timeout(Duration::from_millis(100));
+				let updates_available = channel_manager.await_persistable_update_timeout(Duration::from_millis(100));
```
jkczyz, line 230: Unrelated change can be removed.

lightning-background-processor/src/lib.rs
```diff
@@ -744,6 +768,38 @@ mod tests {
...
+		network_graph.update_channel_from_partial_announcement(42, 53, features, nodes[0].node.get_our_node_id(), nodes[1].node.get_our_node_id());
+		let original_graph_description = network_graph.to_string();
+		assert!(original_graph_description.contains("42: features: 0000, node_one:"));
+
+		let event_handler = |_: &_| {};
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), true, nodes[0].logger.clone());
+
+		let reception_result = receiver
+			.recv_timeout(Duration::from_secs(EVENT_DEADLINE));
+		assert!(reception_result.is_err());
```
jkczyz, line 789: Hmm... this is a 5-second sleep in practice. Seems like you need at least a second for the prune logic, though. 😕 Ideally, we could use the `Time` trait being refactored out in #1418. But even then you can't be guaranteed that the code has been executed yet in the other thread. May have to add logging and check for it similarly as done in `test_timer_tick_called`.

lightning-background-processor/src/lib.rs
```diff
@@ -744,6 +768,38 @@ mod tests {
...
+	#[test]
+	fn test_not_pruning_network_graph_until_graph_sync_completion() {
+		let mut nodes = create_nodes(2, "test_not_pruning_network_graph_until_graph_sync_completion".to_string());
+		let channel_value = 100000;
+		let data_dir = nodes[0].persister.get_data_dir();
+		let (sender, receiver) = std::sync::mpsc::sync_channel(1);
+		let persister = Persister::new(data_dir.clone()).with_graph_sender(sender);
+		let network_graph = nodes[0].network_graph.clone();
+		let features = ChannelFeatures::empty();
+		network_graph.update_channel_from_partial_announcement(42, 53, features, nodes[0].node.get_our_node_id(), nodes[1].node.get_our_node_id());
```
jkczyz, line 780: Could you address the build warnings?

lightning-background-processor/src/lib.rs
arik-so: this comment isn't showing me the affected line numbers for some reason?

lightning-background-processor/src/lib.rs
arik-so: Probably something for a future PR, but should these loop-based unit tests perhaps have a timeout exit condition to avoid potentially stalling a failing test indefinitely?

jkczyz: > Probably something for a future PR,

Just replacing this with a loop like `test_timer_tick_called` should be pretty trivial, no? The larger refactor can wait for a future PR, of course, but no need adding 5 seconds to our tests.

>but should these loop-based unit tests perhaps have a timeout exit condition to avoid potentially stalling a failing test indefinitely?

Yeah, definitely!

lightning-background-processor/src/lib.rs
jkczyz: Line 236 on the latest push.

lightning-background-processor/src/lib.rs
jkczyz: > Just replacing this with a loop like `test_timer_tick_called` should be pretty trivial, no? The larger refactor can wait for a future PR, of course, but no need adding 5 seconds to our tests.
> 

Ah, nevermind. Just saw you already added the loop. Thanks!

lightning-background-processor/src/lib.rs
arik-so: lol now it's showing me the lines, but saying it's 230

jkczyz: Yeah, 230 is on the earlier commit, I believe.

lightning-background-processor/src/lib.rs
jkczyz: `GraphSync` wouldn't necessarily need a reference to `NetworkGraph` -- that can still be taken as a parameter to `read_network_graph`. It would mainly be used to encapsulate the bool such that `read_network_graph` sets it as required instead of relying on the user to do it. Also, `BackgroundProcessor` just needs to read it, so we shouldn't allow for `BackgroundProcessor` to modify it by passing a bool.

As for the dependency, we can avoid it by having `BackgroundProcessor` take a closure instead of either a `GraphSync` reference as mentioned here or an `AtomicBool` as is currently done. Then the user would pass in `GraphSync::is_processing`.

arik-so: yeah, I do think it's cleaner to take a dependency, however, I think the current approach is very similar to the chain_sync_complete() method, and the interface is quite straightforward.

jkczyz: I disagree. The current approach requires the caller to explicitly signal to `BackgroundProcessor` that the sync is complete whereas my alternative doesn't place that burden on the user.

lightning-background-processor/src/lib.rs
```diff
@@ -195,6 +204,11 @@ impl BackgroundProcessor {
 	{
 		let stop_thread = Arc::new(AtomicBool::new(false));
 		let stop_thread_clone = stop_thread.clone();
+		let is_initial_sync_complete = if let Some(rapid_sync) = rapid_gossip_sync {
```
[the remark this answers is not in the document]
arik-so: honestly, I might actually instead change it to taking a reference to a gossip sync. Absolutely no point in holding the entire sync struct imo.

lightning-background-processor/src/lib.rs
```diff
@@ -56,7 +58,7 @@ use std::ops::Deref;
 #[must_use = "BackgroundProcessor will immediately stop on drop. It should be stored until shutdown."]
 pub struct BackgroundProcessor {
 	stop_thread: Arc<AtomicBool>,
-	thread_handle: Option<JoinHandle<Result<(), std::io::Error>>>,
+	thread_handle: Option<JoinHandle<Result<(), std::io::Error>>>
```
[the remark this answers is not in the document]
arik-so: grrr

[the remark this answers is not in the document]
arik-so: seems like my IDE removed it in an auto-format

arik-so:
Which latest timestamp synced value are you referring to – the timestamp included in the rapid sync, the local timestamp when the rapid sync was applied, or the timestamp the latest regular sync update was applied?

lightning-background-processor/src/lib.rs
```diff
@@ -79,6 +79,11 @@ const PING_TIMER: u64 = 1;
 /// Prune the network graph of stale entries hourly.
 const NETWORK_PRUNE_TIMER: u64 = 60 * 60;
 
+#[cfg(all(not(test), debug_assertions))]
+const SCORER_PERSIST_TIMER: u64 = 30;
+#[cfg(test)]
+const SCORER_PERSIST_TIMER: u64 = 1;
+
```
jkczyz, line 86: When you squash commits, could you make sure this and related changes remain in a separate commit with an appropriate commit message?

lightning-background-processor/src/lib.rs
```diff
@@ -151,6 +153,13 @@ impl BackgroundProcessor {
 	/// [`Persister::persist_graph`]: lightning::util::persist::Persister::persist_graph
 	/// [`NetworkGraph`]: lightning::routing::network_graph::NetworkGraph
 	/// [`NetworkGraph::write`]: lightning::routing::network_graph::NetworkGraph#impl-Writeable
+	///
+	/// # Rapid Gossip Sync
+	///
+	/// If a rapid gossip sync is meant to run at startup, pass an option with a set
```
jkczyz, line 164: s/an option with a set/an optional

lightning-background-processor/src/lib.rs
```diff
@@ -151,6 +153,13 @@ impl BackgroundProcessor {
...
+	/// [`RapidGossipSync`] reference to `rapid_gossip_sync` to indicate to [`BackgroundProcessor`]
+	/// not to prune the [`NetworkGraph`] instance until the [`RapidGossipSync`] instance
+	/// completes its first sync.
```
jkczyz, line 167: Could you move this above the links?

lightning-background-processor/src/lib.rs
```diff
@@ -602,7 +639,8 @@ mod tests {
 		let data_dir = nodes[0].persister.get_data_dir();
 		let persister = Arc::new(Persister::new(data_dir));
 		let event_handler = |_: &_| {};
-		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(), nodes[0].logger.clone(), Some(nodes[0].scorer.clone()));
+		let rapid_gossip_sync: Option<Arc<RapidGossipSync<Arc<NetworkGraph>>>> = None;
```
jkczyz, line 651: Could we make this part of `Node` so this line isn't needed in every test?

lightning-background-processor/src/lib.rs
```diff
@@ -791,6 +835,69 @@ mod tests {
...
+		let data_dir = nodes[0].persister.get_data_dir();
+		let (sender, receiver) = std::sync::mpsc::sync_channel(1);
+		let persister = Arc::new(Persister::new(data_dir.clone()).with_graph_persistence_notifier(sender));
+		let network_graph = nodes[0].network_graph.clone();
+		let rapid_sync = Arc::new(RapidGossipSync::new(network_graph.clone()));
+		let features = ChannelFeatures::empty();
+		network_graph.add_channel_from_partial_announcement(42, 53, features, nodes[0].node.get_our_node_id(), nodes[1].node.get_our_node_id())
+			.expect("Failed to update channel from partial announcement");
+		let original_graph_description = network_graph.to_string();
+		assert!(original_graph_description.contains("42: features: 0000, node_one:"));
```
jkczyz, line 879: This assertion doesn't seem necessary.

lightning-background-processor/src/lib.rs
```diff
@@ -791,6 +835,69 @@ mod tests {
...
+		let _ = receiver
+			.recv_timeout(Duration::from_secs(super::FIRST_NETWORK_PRUNE_TIMER * 2))
+			.expect("Network graph not pruned within deadline");
+		let current_graph_description = network_graph.to_string();
+
+		background_processor.stop().unwrap();
+
+		assert_ne!(current_graph_description, original_graph_description);
+		assert!(!current_graph_description.contains("node_one:"));
+		assert!(!current_graph_description.contains("node_two:"));
```
jkczyz, line 924: Likewise, is this needed if we later assert there are no channels?

lightning-background-processor/src/lib.rs
```diff
@@ -175,9 +184,11 @@ impl BackgroundProcessor {
 		PM: 'static + Deref<Target = PeerManager<Descriptor, CMH, RMH, L, UMH>> + Send + Sync,
 		S: 'static + Deref<Target = SC> + Send + Sync,
 		SC: WriteableScore<'a>,
+		RGS: 'static + Deref<Target = RapidGossipSync<G>> + Send
```
jkczyz, line 192: Will this need to be `Sync`, too?

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -68,32 +74,55 @@ pub mod error;
...
-/// `sync_path`: Path to the file where the gossip update data is located
-///
-pub fn sync_network_graph_with_file_path(
-	network_graph: &network_graph::NetworkGraph,
-	sync_path: &str,
-) -> Result<u32, GraphSyncError> {
-	let mut file = File::open(sync_path)?;
-	processing::update_network_graph_from_byte_stream(&network_graph, &mut file)
+/// Rapid Gossip Sync holder struct
+/// todo: ask reviewers for more interesting things to say here
```
jkczyz: I would explain how this differs from P2P gossip syncing.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -68,32 +74,55 @@ pub mod error;
...
+pub struct RapidGossipSync<NG: Deref<Target=NetworkGraph>> {
+	network_graph: NG,
+	is_initial_sync_complete: Arc<AtomicBool>
+}
+
+impl<NG: Deref<Target=NetworkGraph>> RapidGossipSync<NG> {
+	/// Instantiate a new RapidGossipSync holder
+	/// todo: same as above
```
jkczyz: Wouldn't refer to this as a holder, FWIW.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -68,32 +74,55 @@ pub mod error;
...
+	///
+	pub fn sync_network_graph_with_file_path(
+		&self,
+		sync_path: &str,
+	) -> Result<u32, GraphSyncError> {
+		let mut file = File::open(sync_path)?;
+		self.update_network_graph_from_byte_stream(&mut file)
+	}
+
+	/// Bool indicating whether a rapid gossip sync has completed at least once
```
jkczyz, line 110: s/Bool indicating/Returns

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -152,11 +181,12 @@ mod tests {
 		let graph_sync_test_file = sync_test.get_test_file_path();
 
 		let block_hash = genesis_block(Network::Bitcoin).block_hash();
-		let network_graph = NetworkGraph::new(block_hash);
+		let network_graph = Arc::new(NetworkGraph::new(block_hash));
```
jkczyz, line 185: Can't you simply pass a reference in these tests?

lightning-background-processor/src/lib.rs
arik-so: It helps assert that channel 42 is actually there.

lightning-background-processor/src/lib.rs
arik-so: It helps assert that channel 42 is actually there.

lightning-background-processor/src/lib.rs
arik-so: Doesn't appear to be.

lightning-rapid-gossip-sync/src/lib.rs
arik-so: The code's outdated.

lightning-rapid-gossip-sync/src/lib.rs
arik-so: This code's outdated.

lightning-background-processor/src/lib.rs
```diff
@@ -791,6 +838,83 @@ mod tests {
...
 
+	#[test]
+	fn test_scorer_persistence() {
+		let nodes = create_nodes(2, "test_scorer_persistence".to_string());
+		let data_dir = nodes[0].persister.get_data_dir();
+		let persister = Arc::new(Persister::new(data_dir));
+		let event_handler = |_: &_| {};
+		let bg_processor = BackgroundProcessor::start(persister, event_handler, nodes[0].chain_monitor.clone(), nodes[0].node.clone(), nodes[0].net_graph_msg_handler.clone(), nodes[0].peer_manager.clone(),  nodes[0].logger.clone(), Some(nodes[0].scorer.clone()), nodes[0].rapid_gossip_sync.clone());
+
+		loop {
```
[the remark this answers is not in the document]
arik-so: Not sure I get the question. I can wait until is_initial_sync_complete is true, but there are no mpsc messages being sent or other events being fired. Whatever I do, it's gonna be busy polling. Are you thinking of the stuff I did on the server?

arik-so:
Happy to squash into separate commits, just lmk preferred grouping.

lightning-background-processor/src/lib.rs
```diff
@@ -899,7 +906,7 @@ mod tests {
 		assert_eq!(network_graph.read_only().channels().len(), 3);
 
 		let _ = receiver
-			.recv_timeout(Duration::from_secs(super::FIRST_NETWORK_PRUNE_TIMER * 2))
+			.recv_timeout(Duration::from_secs(super::FIRST_NETWORK_PRUNE_TIMER * 5))
```
jkczyz, line 909: This should go in the previous commit.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -74,17 +73,20 @@ pub mod error;
 pub mod processing;
 
 /// Rapid Gossip Sync holder struct
+/// See [module-level documentation] for usage.
```
jkczyz, line 76: All the changes to this file in this commit should be in the previous commit.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -68,19 +72,43 @@ pub mod error;
...
-pub fn sync_network_graph_with_file_path(
-	network_graph: &network_graph::NetworkGraph,
-	sync_path: &str,
-) -> Result<u32, GraphSyncError> {
-	let mut file = File::open(sync_path)?;
-	processing::update_network_graph_from_byte_stream(&network_graph, &mut file)
+/// [module-level documentation]: crate
+pub struct RapidGossipSync<NG: Deref<Target=NetworkGraph>> {
+	network_graph: NG,
+	is_initial_sync_complete: AtomicBool
```
jkczyz, line 81: Is this redundant with `NetworkGraph::last_rapid_gossip_sync_timestamp`? Would be nice if `last_rapid_gossip_sync_timestamp` was in `RapidGossipSync` instead given it's not always needed with `NetworkGraph`.

lightning-background-processor/src/lib.rs
```diff
@@ -286,10 +292,9 @@ impl BackgroundProcessor {
 						log_trace!(logger, "Assessing prunability of network graph");
 
 						// The network graph must not be pruned while rapid sync completion is pending
-						let should_prune = if let Some(rapid_sync) = rapid_gossip_sync.as_ref() {
-							rapid_sync.is_initial_sync_complete()
-						} else {
-							true
+						let should_prune = match rapid_gossip_sync.as_ref(){
+							Some(rapid_sync) => { rapid_sync.is_initial_sync_complete() }
+							None => { true }
```
jkczyz, line 297: Wrong commit. 😢

lightning-background-processor/src/lib.rs
```diff
@@ -273,21 +289,38 @@ impl BackgroundProcessor {
...
 				if last_prune_call.elapsed().as_secs() > if have_pruned { NETWORK_PRUNE_TIMER } else { FIRST_NETWORK_PRUNE_TIMER } {
 					if let Some(ref handler) = net_graph_msg_handler {
-						log_trace!(logger, "Pruning network graph of stale entries");
-						handler.network_graph().remove_stale_channels();
-						if let Err(e) = persister.persist_graph(handler.network_graph()) {
-							log_error!(logger, "Error: Failed to persist network graph, check your disk and permissions {}", e)
+						log_trace!(logger, "Assessing prunability of network graph");
+
+						// The network graph must not be pruned while rapid sync completion is pending
+						let should_prune = match rapid_gossip_sync.as_ref(){
```
jkczyz, line 295: Isn't it the case that user of `RapidGossipSync` won't also use `NetGraphMsgHandler`? If so, there would need to be mutually exclusive persistence paths. Otherwise, the graph will never be persisted.

lightning-background-processor/src/lib.rs
```diff
@@ -273,21 +289,38 @@ impl BackgroundProcessor {
...
+							Some(rapid_sync) => { rapid_sync.is_initial_sync_complete() }
+							None => { true }
```
jkczyz, line 297: nit: drop braces

lightning-rapid-gossip-sync/src/lib.rs
arik-so: no, it's in network graph for easier serialization and persistence. Admittedly a bit of a hack, but one that users will appreciate

lightning-background-processor/src/lib.rs
jkczyz: But if they don't opt to use the p2p updates (i.e., when `net_graph_msg_handler` is `None`), won't the `NetworkGraph` never be persisted?

arik-so: Yes, that is true. They do already have the downloaded snapshot, which actually takes less time to convert into a network graph than a conventionally serialized network graph, so I'm not sure that's really a problem.

jkczyz: Should we not bother serializing the timestamp then? That way it can go inside `RapidGossipSync` instead of `NetworkGraph`. (cc: @TheBlueMatt)

arik-so: We should definitely bother serializing the timestamp. I think the real question is whether we should expose the NetworkGraph access from the RapidGossipSync struct for persistence purposes.

lightning/src/routing/network_graph.rs
```diff
@@ -1078,6 +1078,18 @@ impl NetworkGraph {
 		}
 	}
 
+	/// The unix timestamp in UTC provided by the most recent rapid gossip sync.
```
jkczyz, line 1081: Drop "in UTC".

lightning/src/routing/network_graph.rs
```diff
@@ -1078,6 +1078,18 @@ impl NetworkGraph {
...
+	/// It will be set by the rapid sync process after every sync completion.
+	pub fn get_last_rapid_gossip_sync_timestamp(&self) -> Option<u32> {
+		self.last_rapid_gossip_sync_timestamp.lock().unwrap().clone()
+	}
+
+	/// Update the UTC timestamp provided by the most recent rapid gossip sync.
```
jkczyz, line 1087: s/UTC/unix

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -68,19 +72,43 @@ pub mod error;
...
+}
+
+impl<NG: Deref<Target=NetworkGraph>> RapidGossipSync<NG> {
+	/// Instantiate a new RapidGossipSync instance
```
jkczyz, line 85: nit: ticks

lightning-background-processor/src/lib.rs
```diff
@@ -288,37 +288,32 @@ impl BackgroundProcessor {
...
-						None => true
+					let graph_to_prune = match rapid_gossip_sync.as_ref() {
+						Some(rapid_sync) => {
+							if rapid_sync.is_initial_sync_complete() {
+								Some(rapid_sync.network_graph())
+							} else {
+								None
+							}
+						},
+						None => net_graph_msg_handler.as_ref().map_or(None, |handler| Some(handler.network_graph()))
```
jkczyz, line 301: Just use `map`, it's simpler.

lightning-background-processor/src/lib.rs
```diff
@@ -288,37 +288,32 @@ impl BackgroundProcessor {
...
-							last_prune_call = Instant::now();
-							have_pruned = true;
+						if let Err(e) = persister.persist_graph(network_graph_reference) {
+							log_error!(logger, "Error: Failed to persist network graph, check your disk and permissions {}", e)
 						}
+
+						last_prune_call = Instant::now();
+						have_pruned = true;
 					} else {
 						log_trace!(logger, "Not pruning network graph due to pending gossip sync");
```
jkczyz, line 314: Or because neither sync was given?

lightning-rapid-gossip-sync/src/processing.rs
```diff
@@ -23,203 +26,207 @@ const GOSSIP_PREFIX: [u8; 4] = [76, 68, 75, 1];
...
-		let current_node_id = Readable::read(read_cursor)?;
-		node_ids.push(current_node_id);
+impl<NG: Deref<Target=NetworkGraph>> RapidGossipSync<NG> {
+	/// Update network graph from binary data.
+	/// Returns the last sync timestamp to be used the next time rapid sync data is queried.
+	///
+	/// `network_graph`: network graph to be updated
+	///
+	/// `update_data`: `&[u8]` binary stream that comprises the update data
+	pub fn update_network_graph(&self, update_data: &[u8], ) -> Result<u32, GraphSyncError> {
```
jkczyz, line 36: nit: remove trailing `, ` in parameter list.

lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -68,19 +72,43 @@ pub mod error;
 /// Core functionality of this crate
 pub mod processing;
 
-/// Sync gossip data from a file
-/// Returns the last sync timestamp to be used the next time rapid sync data is queried.
+/// Rapid Gossip Sync struct
+/// See [module-level documentation] for usage.
```
jkczyz, line 76: s/module/crate

lightning/src/routing/network_graph.rs
```diff
@@ -1078,6 +1078,18 @@ impl NetworkGraph {
 		}
 	}
 
+	/// The unix timestamp in UTC provided by the most recent rapid gossip sync.
+	/// It will be set by the rapid sync process after every sync completion.
+	pub fn get_last_rapid_gossip_sync_timestamp(&self) -> Option<u32> {
```
jkczyz, line 1083: Do we describe anywhere how this should be used?

lightning-background-processor/src/lib.rs
```diff
@@ -272,22 +288,42 @@ impl BackgroundProcessor {
...
+					log_trace!(logger, "Assessing prunability of network graph");
+					let graph_to_prune = match rapid_gossip_sync.as_ref() {
+						Some(rapid_sync) => {
+							if rapid_sync.is_initial_sync_complete() {
+								Some(rapid_sync.network_graph())
+							} else {
+								None
+							}
+						},
+						None => net_graph_msg_handler.as_ref().map_or(None, |handler| Some(handler.network_graph()))
```
jkczyz, line 301: FWIW, the interface is prone to misuse since the user could potentially pass in both syncs with two different `NetworkGraph`s, but only one would be persisted. Not sure if we really care but it would be a reason to prefer an enum over two `Option`s.

lightning-rapid-gossip-sync/src/lib.rs
arik-so: what do you mean?

lightning-background-processor/src/lib.rs
arik-so: Phrased it differently

lightning/src/routing/network_graph.rs
arik-so: Yeah, the rapid gossip sync crate docs talk about the timestamp

lightning-background-processor/src/lib.rs
arik-so: True, but with an enum, you cannot use both a rapid sync and a regular message handler. I can make an extra check verifying that the network graph references are the same somehow? Is there a way to do that that will work with Arc?

lightning-background-processor/src/lib.rs
jkczyz: > As for persisting the timestamp, I don't think we need to worry about persisting it or not, kinda who cares? And, indeed, someone may want to move from rapid to P2P or back, so might as well persist it.

Hmm... if it doesn't matter if we persist it, but including it in `NetworkGraph` so we can persist it is a hack, why do it? Seems to just add complexity while exposing an API that only some users would care about but largely only exists for the `lightning-rapid-gossip-sync` crate's internal usage.

Not persisting it but keeping it internal to `RapidGossipSync` only affects the restart case. Or would fetching new rapid sync data be so infrequent that it would only happen on a restart (i.e., user app is mostly closed)? And thus persisting the timestamp would be important.

arik-so: The main use case for rapid sync is mobile devices, where the most common state of the app will be "launched within the previous 30 seconds." Persisting the timestamp allows them to fetch differential snapshots whose size will be one or a couple hundred kB instead of 2MB. It can shave off a couple seconds on each app start.

lightning-background-processor/src/lib.rs
jkczyz: > True, but with an enum, you cannot use both a rapid sync and a regular message handler.

How useful is this for a mobile? Wouldn't the app only be open infrequently? Just trying to understand the use case of having both.

> I can make an extra check verifying that the network graph references are the same somehow? Is there a way to do that that will work with Arc?

Probably some pointer comparison though not sure if we want to do so.

lightning-rapid-gossip-sync/src/lib.rs
jkczyz: Put ticks around the identifiers in rust docs so they are formatted.

lightning-background-processor/src/lib.rs
arik-so: > How useful is this for a mobile? Wouldn't the app only be open infrequently? Just trying to understand the use case of having both.

That depends on whether mobile implementations will choose to obtain gossip data in between rapid syncs. If some apps envision being open for up to a minute or two, it's a relatively cheap thing to do in the background, and allows to minimize trust after the initial fast forward. But we're all just guessing here, I can survey some folks who'll be using it.

jkczyz: Just occurred to me that we currently use `NetGraphMsgHandler` to handle the `network_update` field of `Event::PaymentPathFailed`. We'll have to do something similar in the rapid-sync-only case.

jkczyz (approved):
As discussed offline, let's leave any refactors for a follow-up.

PR #1464, opened 2022-05-03
Fix renamed `broken_intra_doc_links` lint.

tnull:
This fixes some warnings during `cargo test`. 

I tried this before in #1249, but it only became feasible since MSRV is 1.14.1.


tnull:
Nvm, I was mistaken and this was only added in 1.52.0...

PR #1474, opened 2022-05-10
lightning-invoice/utils: Actually add expiry to invoices

### LINKED ISSUE #1411: Include expiry in invoices

dunxen:
My bad.

I've also added the expiry to non-phantom invoice utilities.

This should now fix #1411


lightning-invoice/src/utils.rs
```diff
@@ -931,10 +940,11 @@ mod test {
 		];
 
 		let description_hash = crate::Sha256(Hash::hash("Description hash phantom invoice".as_bytes()));
-		let invoice = ::utils::create_phantom_invoice_with_description_hash::<EnforcingSigner,&test_utils::TestKeysInterface>(Some(payment_amt), None, 3600, description_hash, route_hints, &nodes[1].keys_manager, Currency::BitcoinTestnet).unwrap();
+		let invoice = ::utils::create_phantom_invoice_with_description_hash::<EnforcingSigner,&test_utils::TestKeysInterface>(Some(payment_amt), None, 4200, description_hash, route_hints, &nodes[1].keys_manager, Currency::BitcoinTestnet).unwrap();
```
tnull, line 943: tiny nit: missing whitespace before `&test_utils`.

Also, is there any particular reason why the expiry is set to `4200` here and `3600` everywhere else? If so, maybe use a variable to clarify the difference in the variable name?

tnull (approved):
Tiny nit, otherwise LGTM.

lightning-invoice/src/utils.rs
dunxen: Thanks!
So I just made it different to the "default" expiry to make it pretty clear it is set. (But seems pretty useless since the expiry is definitely required now).

So I can make it 3600 again or make a variable. Don't think it matters anymore.

lightning-invoice/src/utils.rs
```diff
@@ -300,13 +307,14 @@ where
...
 		),
-		duration_since_epoch,
+		duration_since_epoch, invoice_expiry_delta_secs
 	)
 }
 
 fn _create_invoice_from_channelmanager_and_duration_since_epoch<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref>(
 	channelmanager: &ChannelManager<Signer, M, T, K, F, L>, keys_manager: K, network: Currency,
 	amt_msat: Option<u64>, description: InvoiceDescription, duration_since_epoch: Duration,
+	invoice_expiry_delta_secs: u32
```
jkczyz, line 317: We'll need to pass this to `create_inbound_payment` below instead of `DEFAULT_EXPIRY_TIME`. Otherwise, `ChannelManager` will not honor the expiry.

lightning-invoice/src/utils.rs
```diff
@@ -227,7 +231,7 @@ where
 	let duration = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)
 		.expect("for the foreseeable future this shouldn't happen");
 	create_invoice_from_channelmanager_and_duration_since_epoch(
-		channelmanager, keys_manager, network, amt_msat, description, duration
+		channelmanager, keys_manager, network, amt_msat, description, duration, invoice_expiry_delta_secs
```
jkczyz, line 234: I realize we haven't been consistent in this file, but could you wrap parameter and argument lists at 100 characters here and other places that you touched?

jkczyz (commented):
Thanks for the quick turnaround! We should have caught this in review, so no worries.

lightning-invoice/src/utils.rs
```diff
@@ -931,10 +946,23 @@ mod test {
...
+			EnforcingSigner,
+			&test_utils::TestKeysInterface,
+		>(
+			Some(payment_amt),
+			None,
+			non_default_invoice_expiry_secs,
+			description_hash,
+			route_hints,
+			&nodes[1].keys_manager,
+			Currency::BitcoinTestnet,
```
dunxen, line 960: Let me know if this is a bit overkill for wrapping. Just modified what rustfmt did. I know it doesn't play too nicely with this repo in general so can wrap more like above.

jkczyz: A good middle ground is to wrap at 100 characters as you are but also do so for parameter lists such that the parameters lists are packed into that width instead of having a parameter on each line. I'd leave off the trailing comma when doing so vs keeping it when there is one parameter per line.

dunxen: Cool. Will just fix this one up then to match the others :)

lightning-invoice/src/utils.rs
dunxen: Yes, will split them up :)

dunxen:
Rebasing

dunxen:
> Needs rebase, it appears?

Beat me to it. Was quite behind on my local.

lightning-invoice/src/utils.rs
```diff
@@ -811,6 +812,7 @@ mod test {
 		assert_eq!(invoice.min_final_cltv_expiry(), MIN_FINAL_CLTV_EXPIRY as u64);
 		assert_eq!(invoice.description(), InvoiceDescription::Direct(&Description("test".to_string())));
 		assert_eq!(invoice.route_hints().len(), 2);
+		assert_eq!(invoice.expiry_time(), Duration::from_secs(3600));
```
jkczyz, line 827: We may want to use something other than `DEFAULT_EXPIRY_TIME` in tests since `Invoice::expiry_time` will return that if not set (i.e., just updating the test and not the implementation here would have caused this assertion to pass).

lightning-invoice/src/utils.rs
```diff
@@ -529,10 +540,11 @@ mod test {
 		create_unannounced_chan_between_nodes_with_value(&nodes, 0, 1, 100000, 10001, InitFeatures::known(), InitFeatures::known());
 		let invoice = create_invoice_from_channelmanager_and_duration_since_epoch(
 			&nodes[1].node, nodes[1].keys_manager, Currency::BitcoinTestnet, Some(10_000), "test".to_string(),
-			Duration::from_secs(1234567)).unwrap();
+			Duration::from_secs(1234567), 3600).unwrap();
 		assert_eq!(invoice.amount_pico_btc(), Some(100_000));
 		assert_eq!(invoice.min_final_cltv_expiry(), MIN_FINAL_CLTV_EXPIRY as u64);
 		assert_eq!(invoice.description(), InvoiceDescription::Direct(&Description("test".to_string())));
+		assert_eq!(invoice.expiry_time(), Duration::from_secs(3600));
```
jkczyz, line 547: Similarly here.

jkczyz (commented):
Looks good modulo some test changes.

lightning-invoice/src/utils.rs
dunxen: Just have a power cut in 2 min for the next two hours so might be after that only.

lightning-invoice/src/utils.rs
```diff
@@ -228,7 +230,8 @@ where
 	let duration = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)
 		.expect("for the foreseeable future this shouldn't happen");
 	create_invoice_from_channelmanager_and_duration_since_epoch(
-		channelmanager, keys_manager, network, amt_msat, description, duration
+		channelmanager, keys_manager, network, amt_msat,
+		description, duration, invoice_expiry_delta_secs
```
jkczyz, line 234: nit: would just wrap at 100 chars instead of making them evenly distributed to ease future automation

jkczyz (approved):

lightning-invoice/src/utils.rs
```diff
@@ -799,8 +813,14 @@ mod test {
 		} else {
 			None
 		};
+		let non_default_invoice_expiry_secs = 4200;
 
-		let invoice = ::utils::create_phantom_invoice::<EnforcingSigner, &test_utils::TestKeysInterface>(Some(payment_amt), payment_hash, "test".to_string(), 3600, route_hints, &nodes[1].keys_manager, Currency::BitcoinTestnet).unwrap();
+		let invoice = ::utils::create_phantom_invoice::<
+			EnforcingSigner, &test_utils::TestKeysInterface
+		>(
```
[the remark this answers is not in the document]
dunxen: I'll get to these in a follow-up today :) (also Jeff's)

lightning-invoice/src/utils.rs
jkczyz: Ah, I was actually wrong about this. We wouldn't need `Time` to test this because the expiry is checked against the block header time.

https://github.com/lightningdevkit/rust-lightning/blob/65920818db58880f6576fd50c3ea5df273912978/lightning/src/ln/channelmanager.rs#L5399

dunxen: Oh, interesting. So it should be pretty straightforward to test.

jkczyz: Yeah, though, would probably need to modify `functional_test_utils` to set the time in a more realistic manner. Currently, it just uses the block height.

PR #1479, opened 2022-05-12
Pass `counterparty_node_id` to `ChannelManager` functions

ViktorTigerstrom:
After #105 and #422, channels will be stored in the `ChannelManager::per_peer_state` which requires the `counterparty_node_id` to access them. #1278 also requires this updated storage design. This will require that the user passes the `counterparty_node_id` to some of the public API functions of the `ChannelManager` for which only the `channel_id` is passed currently. 

This PR updates the parameters required in those functions, to include the `counterparty_node_id`. 

I figured it's probably better to separate the API changes from rest of the channel storage changes. Let me know if you're of a different opinion, and I'll close this PR.

Pushing this as a draft for now, but let me know if you think that better to merge this before the other functionality is ready (#105 and #422), which I will base on this.

Finally, when updating these functions, I also noticed that lots of other functions that has the `counterparty_node_id` as an input parameter, name the `counterparty_node_id` differently. That could be a bit confusing in my opinion. Examples that I have found are: `their_node_id`, `their_network_key`, `peer_node_id`. Let me know if you'd like me to update the name this PR is using, or if I should update the name in the other functions to make it uniform.

### DIFF
--- a/lightning/src/ln/channelmanager.rs
+++ b/lightning/src/ln/channelmanager.rs
@@ -2686,8 +2686,9 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 
 	/// Handles the generation of a funding transaction, optionally (for tests) with a function
 	/// which checks the correctness of the funding transaction given the associated channel.
-	fn funding_transaction_generated_intern<FundingOutput: Fn(&Channel<Signer>, &Transaction) -> Result<OutPoint, APIError>>
-			(&self, temporary_channel_id: &[u8; 32], funding_transaction: Transaction, find_funding_output: FundingOutput) -> Result<(), APIError> {
+	fn funding_transaction_generated_intern<FundingOutput: Fn(&Channel<Signer>, &Transaction) -> Result<OutPoint, APIError>>(
+		&self, temporary_channel_id: &[u8; 32], _counterparty_node_id: &PublicKey, funding_transaction: Transaction, find_funding_output: FundingOutput
+	) -> Result<(), APIError> {
 		let (chan, msg) = {
 			let (res, chan) = match self.channel_state.lock().unwrap().by_id.remove(temporary_channel_id) {
 				Some(mut chan) => {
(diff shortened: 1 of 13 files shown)


ViktorTigerstrom:
Sorry for the failed checks, will address them tomorrow.

lightning/src/ln/channelmanager.rs
```diff
@@ -2757,7 +2756,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 	///
 	/// [`Event::FundingGenerationReady`]: crate::util::events::Event::FundingGenerationReady
 	/// [`Event::ChannelClosed`]: crate::util::events::Event::ChannelClosed
-	pub fn funding_transaction_generated(&self, temporary_channel_id: &[u8; 32], funding_transaction: Transaction) -> Result<(), APIError> {
+	pub fn funding_transaction_generated(&self, temporary_channel_id: &[u8; 32], counterparty_node_id: &PublicKey, funding_transaction: Transaction) -> Result<(), APIError> {
```
ViktorTigerstrom, line 2759: Should we add the `counterparty_node_id` to the `FundingGenerationReady` event?

ViktorTigerstrom:
Addressed the feedback and the failed checks. Let me know if you think it's worth adding the extra `counterparty_node_id` check to `create_funding_transaction` (commit a85a63e), and I'll squash it with the `FundingGenerationReady` commit! Else I will drop that specific commit.

lightning/src/ln/channelmanager.rs
```diff
@@ -4130,6 +4131,9 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 				if !channel.get().inbound_is_awaiting_accept() {
 					return Err(APIError::APIMisuseError { err: "The channel isn't currently awaiting to be accepted.".to_owned() });
 				}
+				if *counterparty_node_id != channel.get().get_counterparty_node_id() {
```
ViktorTigerstrom, line 4134: Let me know if you want me to add a test case for this. Though given that it shouldn't be a thing after #105, perhaps it isn't really worth it.

ViktorTigerstrom:
Thanks @TheBlueMatt!  Squashed the commits without changes. The only one I'm not entirely sure about it ded76da. As it's an update to the docs of the `OpenChannelRequest`, do you prefer it to be a separate commit, or should I squash it with the `Pass counterparty_node_id to accept_inbound_channel` commit as that's the reason for the doc update?

lightning/src/ln/channelmanager.rs
```diff
@@ -1769,7 +1769,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		});
 	}
 
-	fn close_channel_internal(&self, channel_id: &[u8; 32], target_feerate_sats_per_1000_weight: Option<u32>) -> Result<(), APIError> {
+	fn close_channel_internal(&self, channel_id: &[u8; 32], _counterparty_node_id: &PublicKey, target_feerate_sats_per_1000_weight: Option<u32>) -> Result<(), APIError> {
```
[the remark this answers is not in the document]
ViktorTigerstrom: Yeah that was my reasoning behind that choice, but changed that now just in case you'd like to merge this PR before the next PR is completed :).

ViktorTigerstrom:
Addressed the latest feedback, and marked this as an open PR in case you think it's worth merging this before the functionality for #105 is completed :)

arik-so (approved):

PR #1490, opened 2022-05-19
Fix rust beta docs for lightning-invoice crate.

arik-so:


lightning-invoice/src/lib.rs
```diff
@@ -796,19 +796,20 @@ impl SignedRawInvoice {
...
 ///
-/// let elements = vec![A(1), A(2), B(3), A(4)]
+/// let elements = vec![A(1), A(2), B(3), A(4)];
 ///
-/// assert_eq!(find_extract!(elements.iter(), Enum::B(ref x), x), Some(3u16))
+/// assert_eq!(find_extract!(elements.iter(), Enum::B(ref x), x), Some(3u16));
 /// ```
 macro_rules! find_extract {
 	($iter:expr, $enm:pat, $enm_var:ident) => {
+
```
jkczyz, line 812: nit: remove blank line

jkczyz (dismissed):

jkczyz:
> Oof, looks like we need it to use defined types....matbe lets just not make it a code doc?

I think `use Enum::*;` is the problem. IIUC, you can't do this for enum variants. They need to be fully qualified as `Enum::A`, etc.

arik-so:
Oh lol, I have already started figuring out how to fix this

lightning-invoice/src/lib.rs
```diff
@@ -796,17 +796,18 @@ impl SignedRawInvoice {
 ///
 /// The following example would extract the first B.
 /// ```
-/// use Enum::*;
+/// use lightning_invoice::{find_extract, find_all_extract};
```
jkczyz, line 799: nit: only need to import one of them here and below.

lightning-invoice/src/lib.rs
```diff
@@ -796,17 +796,18 @@ impl SignedRawInvoice {
...
 /// 	B(u16)
 /// }
 ///
-/// let elements = vec![A(1), A(2), B(3), A(4)];
+/// let elements = vec![Enum::A(1), Enum::A(2), Enum::B(3), Enum::A(4)];
 ///
-/// assert_eq!(find_extract!(elements.iter(), Enum::B(ref x), x), Some(3u16));
+/// assert_eq!(find_extract!(elements.iter(), Enum::B(x), x), Some(3u16));
 /// ```
+#[macro_export]
```
jkczyz, line 810: Not sure if we want to export these though... Is there a way to do it conditionally for tests?

lightning-invoice/src/lib.rs
arik-so: not technically true. Both need to be imported here, but find_all_extract is indeed sufficient below.

lightning-invoice/src/lib.rs
arik-so: Tried `#[cfg_attr(test, macro_export)]`, but that's not working. It'd be a shame to make the code not be evaluated, as it's verifying the accuracy of the comment though.

arik-so: It's even wilder. When I explicitly say `#[cfg_attr(not(test), macro_export)]` as a sanity check, it works!

jkczyz: How about `doctest`?

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html?highlight=doctest#include-items-only-when-collecting-doctests

arik-so: Ah, you've found the attribute Google wouldn't show me. Yet alas, also no.

arik-so: So question. If the docs show how to use the macro, wouldn't that require the macro be public? Otherwise the docs are simply inaccurate.

jkczyz: I think it was more to test that the macros works as expected and to show internal developers how to use it. It's possible to generate docs for private items, too.

Suppose it could have be a normal test. Maybe already covered by other tests, though. I'd say just remove the ticks as Matt suggested.

arik-so: will do

arik-so: Ok, I gave up on the macros and reverted everything except the semicolons.

lightning-invoice/src/lib.rs
```diff
@@ -795,18 +795,18 @@ impl SignedRawInvoice {
...
 /// 	B(u16)
 /// }
 ///
-/// let elements = vec![A(1), A(2), B(3), A(4)]
+/// let elements = vec![A(1), A(2), B(3), A(4)];
 ///
-/// assert_eq!(find_extract!(elements.iter(), Enum::B(ref x), x), Some(3u16))
-/// ```
+/// assert_eq!(find_extract!(elements.iter(), Enum::B(ref x), x), Some(3u16));
+/// ``
```
jkczyz, line 809: Does double ticking do anything? Thought that would just be the empty string formatted in markdown.

lightning-invoice/src/lib.rs
```diff
@@ -795,18 +795,18 @@ impl SignedRawInvoice {
 /// variant. If no element was found `None` gets returned.
 ///
 /// The following example would extract the first B.
-/// ```
-/// use Enum::*
+/// ``
+/// use Enum::*;
```
jkczyz, line 799: May as well have kept the fully-qualified enum variants as I don't think this was ever valid rust. 😛

jkczyz (commented):
LGTRM - please squash!

jkczyz (approved):

PR #1499, opened 2022-05-27
Fix 'unreachable' warning during tests.

tnull:
This fixes an `unreachable` warning during tests that was introduced with #1155.


lightning-rapid-gossip-sync/src/lib.rs
```diff
@@ -197,6 +197,7 @@ mod tests {
 				println!("{}", error_string);
 				return;
 			}
+			#[cfg(require_route_graph_test)]
```
arik-so, line 200: Is this necessary? If `require_route_graph_test` is not active, this line would already not be getting hit to begin with.

arik-so: Lol nvm, gotcha.

arik-so (approved):

PR #1527, opened 2022-06-07
Expose method to update HTLC relay policy

wpaulino:
A new `update_htlc_relay_policy` method is exposed on the `ChannelManger` to update the HTLC relay policy for a set of channels atomically. New ChannelUpdate events are generated for each eligible channel.

Note that as currently implemented, a buggy and/or auto-policy-management client could spam the network with updates as there is no rate-limiting in place. This could already be done with `broadcast_node_announcement`, though users are less inclined to update that as frequently as its data is mostly static.

Fixes https://github.com/lightningdevkit/rust-lightning/issues/216.

I noticed the previous PR attempting this wanted to refactor the HTLC relay parameters out of `ChannelConfig`. I went with the most minimal diff here -- if that is still desired then I can work on a parent branch to address that separately.

PR #1529, opened 2022-06-08
Move ChannelConfig static fields to ChannelHandshakeConfig

wpaulino:
In the near future, we plan to allow users to update their `ChannelConfig` after the initial channel handshake. In order to reuse the same struct and expose it to users, we opt to move out all static fields that cannot be updated after the initial channel handshake.

`announced_channel` and `commit_upfront_shutdown_pubkey` represent these fields. We now store `announced_channel` within the `Channel` struct, making sure we're backwards compatible to allow users to downgrade if necessary. `commit_upfront_shutdown_pubkey` doesn't need to be stored since it only determines whether a
`Channel` stores a `shutdown_scriptpubkey` or not prior to a `Channel` being stored.

Motivated by https://github.com/lightningdevkit/rust-lightning/pull/1527#discussion_r892512704.


lightning/src/util/config.rs
```diff
@@ -265,18 +276,8 @@ pub struct ChannelConfig {
...
-	/// This should only be set to true for nodes which expect to be online reliably.
-	///
-	/// As the node which funds a channel picks this value this will only apply for new outbound
-	/// channels unless [`ChannelHandshakeLimits::force_announced_channel_preference`] is set.
-	///
-	/// This cannot be changed after the initial channel handshake.
-	///
-	/// Default value: false.
-	pub announced_channel: bool,
+	/// Deprecated, use [`ChannelHandshakeConfig::announced_channel`] instead.
```
tnull, line 294: Hm, if you deprecate this here, could you update the rest of the codebase to directly use the new location?

Also, instead of a comment, you probably want to mark this as deprecated:
```
#[deprecated(
	since = "0.0.107",
	note = "ChannelHandshakeConfig::announced_channel should be used instead.",
)]
```
(Yes, if you cannot update all instances where it is used immediately, this will produce annoying warning messages, which is however exactly the idea.)

lightning/src/util/config.rs
```diff
@@ -278,18 +293,8 @@ pub struct ChannelConfig {
...
-	///
-	/// This is set to true by default to provide a slight increase in security, though ultimately
-	/// any attacker who is able to take control of a channel can just as easily send the funds via
-	/// lightning payments, so we never require that our counterparties support this option.
-	///
-	/// This cannot be changed after a channel has been initialized.
-	///
-	/// Default value: true.
-	pub commit_upfront_shutdown_pubkey: bool,
+	/// Deprecated, use [`ChannelHandshakeConfig::commit_upfront_shutdown_pubkey`] instead.
```
tnull, line 296: Same here, you probably want to use:

```
#[deprecated(
	since = "0.0.107",
	note = "ChannelHandshakeConfig::commit_upfront_shutdown_pubkey should be used instead.",
)]
```

lightning/src/ln/channel.rs
```diff
@@ -928,9 +929,17 @@ impl<Signer: Sign> Channel<Signer> {
 			}
 		}
 
+		// Also update `ChannelConfig::announced_channel` to make sure we still allow users to
+		// downgrade versions after the fact.
+		//
+		// TODO: This can be removed once we prevent downgrades to v0.0.107.
```
tnull, line 935: Not 100% sure what exactly you mean here with "prevent downgrades to v0.0.107". Do you basically mean "when we're ready to break backwards compat with versions prior to v0.0.107"?

lightning/src/ln/channel.rs
```diff
@@ -495,6 +495,7 @@ pub(super) struct Channel<Signer: Sign> {
 	#[cfg(not(any(test, feature = "_test_utils")))]
 	config: ChannelConfig,
 
+	announced: bool,
```
tnull, line 498: Why is this third instance of an `announced` bool (besides the new `ChannelHandshakeConfig::announced_channel` and the now-deprecated `ChannelConfig::announced_channel`) necessary? Since it seems to be instantiated with `own_channel_config.announced_channel`, couldn't we just use this or move it entirely to `Channel`?

tnull (commented):
Some comments and questions.

Moreover: Not sure if its only me, but I find the current naming of variables in `channel.rs` vs. their type unnecessarily confusing: there are many instances of variables simply named `config`, which may refer to `ChannelConfig`, `UserConfig`, etc. depending on the context. Then there is the `channel_options` field and corresponding variables, which is however a `ChannelConfig`. Moreover, there are fields like `own_channel_config`, which would make you think it would be a `ChannelConfig`, but no, it's actually a `ChannelHandshakeConfig`.

TLDR: Not sure if it needs to happen in this PR, but I personally would appreciate a refactoring that brings the naming of variables more in alignment with their types to improve readability.

lightning/src/util/config.rs
tnull: Mh, I agree that in that timeframe it may be come a nuisance without much benefit. Maybe we could create something like a `TODO.md` in which we keep track of TODOs that cannot be completed right away and where we define until when they are deferred? Or immediately create issues for them with a 'defer' tag and assign an estimated milestone some time in the future?

wpaulino: > In any case, I think we can do better here. Instead of making these fields pub(crate), because we don't really care about users being able to serialize ChannelConfig objects, I think, we can move the serialization to some struct LegacyChannelConfig { config: ChannelConfig, announced_channel: bool, ... } object instead. That way, we can read it all the same, but the public interface remains clean and you can construct a full ChannelConfig without using default() and then modifying. Its a good bit more work cause we can't use the impl_writeable_tlv_based!() macro anymore and have to write out the deserialization manually, but it should be doable.

Sounds good, I assume this means we can reset the serialization schema for the new `ChannelConfig` then? Otherwise, we would want to document that the types 4 (`announced_channel`) & 6 (`commit_upfront_shutdown_key`) cannot be reused as it would break `LegacyChannelConfig` serialization.

> Maybe we could create something like a TODO.md in which we keep track of TODOs that cannot be completed right away and where we define until when they are deferred? Or immediately create issues for them with a 'defer' tag and assign an estimated milestone some time in the future?

:+1: I would prefer the latter.

lightning/src/ln/channel.rs
wpaulino: Correct, I can reword to your suggestion if it's not clear as is.

lightning/src/ln/channel.rs
wpaulino: `ChannelHandshakeConfig` doesn't need to be stored anywhere (and currently isn't) as it's mostly ephemeral, but the `announced` bit should definitely continue to be stored, so I opted to have it within the `Channel` struct.

lightning/src/util/config.rs
```diff
@@ -265,30 +291,6 @@ pub struct ChannelConfig {
 	///
 	/// [`MIN_CLTV_EXPIRY_DELTA`]: crate::ln::channelmanager::MIN_CLTV_EXPIRY_DELTA
 	pub cltv_expiry_delta: u16,
-	/// Set to announce the channel publicly and notify all nodes that they can route via this
```
tnull, line 268: nit: If you split up the commits, you may want to move the removal of the moved parts to the 'move' commits. (Nevermind, if you end up squashing it all).

lightning/src/util/config.rs
```diff
@@ -337,23 +339,83 @@ impl Default for ChannelConfig {
...
-	(4, announced_channel, required),
-	(6, commit_upfront_shutdown_pubkey, required),
-	(8, forwarding_fee_base_msat, required),
-});
+/// Legacy version of [`ChannelConfig`] that stored the static
+/// [`ChannelHandshakeConfig::announced_channel`] and
+/// [`ChannelHandshakeConfig::commit_upfront_shutdown_pubkey`] fields.
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct LegacyChannelConfig {
+	pub(crate) mutable: ChannelConfig,
```
tnull, line 353: Not too sure about this field name. Is mutability its core characteristic/function?

tnull (approved):
LGTM, just two remarks, feel free to ignore.

PR #1534, opened 2022-06-10
Check release build in CI

tnull:
So far, CI did not check the code in the `release` build profile, which could result in some things not getting caught (cf. #1533). 

To fix this, we now implement a new CI jobs that runs checks in the `release` profile.


tnull:
Confirmed that `check_release` does what it is supposed to and fails due to #1533. Will rebase after it is merged.

tnull:
Rebased.

.github/workflows/build.yml
```diff
@@ -275,6 +275,27 @@ jobs:
       - name: For each commit, run cargo check (including in fuzz)
         run: ci/check-each-commit.sh upstream/main
 
+  check_release:
+    runs-on: ubuntu-latest
+    env:
+      TOOLCHAIN: 1.57.0
```
jkczyz, line 281: How do we decide what toolchain to use? I see fuzzing use 1.58.0 and linting 1.47.0.

tnull: Good question, I now switched to `stable` to avoid giving a specific version.

jkczyz: Not quite sure about fuzzing, tbh. `stable` seems fine for `check_release`. We typically only use `nightly` if we need experimental features, IIUC.

tnull: > `stable` seems fine for `check_release`. We typically only use `nightly` if we need experimental features, IIUC.

Yeah, that's what I assumed.

tnull:
Btw, I currently fail to see why some jobs are marked as 'required' and others are not. Maybe `check_release` should be required?

tnull:
> I'm not gonna bother making it required for a while, though - if you do it immediately than all old PRs that don't have this commit in them will fail CI until rebased, which sucks. After a while all new PRs will have the commit so we can make it required.

Ah, thats unfortunate.

PR #1540, opened 2022-06-13
Rename UserConfig and LegacyChannelConfig fields

wpaulino:
The current names aren't very clear to what each field represents, this commit aims to improve that.

Follow-up from #1529.


tnull (approved):
LGTM, thank you very much!

PR #1543, opened 2022-06-14
Look-up functions for `ReadOnlyNetworkGraph`

jkczyz:
`ReadOnlyNetworkGraph` uses `BTreeMap` to store its nodes and channels, but these data structures are not supported by C bindings. Expose look-up functions on these maps in lieu of such support.

### DIFF
--- a/lightning/src/routing/gossip.rs
+++ b/lightning/src/routing/gossip.rs
@@ -1602,11 +1602,21 @@ impl ReadOnlyNetworkGraph<'_> {
 		&*self.channels
 	}
 
+	/// Returns information on a channel with the given id.
+	pub fn channel(&self, short_channel_id: u64) -> Option<&ChannelInfo> {
+		self.channels.get(&short_channel_id)
+	}
+
 	/// Returns all known nodes' public keys along with announced node info.
 	///
 	/// (C-not exported) because we have no mapping for `BTreeMap`s
 	pub fn nodes(&self) -> &BTreeMap<NodeId, NodeInfo> {
 		&*self.nodes
+	}
+
+	/// Returns information on a node with the given id.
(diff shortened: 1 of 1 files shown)


jkczyz:
> Hmmmm, yuckkkk. That's a ton of copying, so much so I'd bet we'll see some nontrivial issues with anyone trying to use this on mobile. I'd much rather expose some method to query for individual entries rather than expose the full set.

Yeah, wasn't really happy about the cloning. But to confirm my understanding of bindings, (1) could we use references instead and (2) is `HashMap` unsupported?

jkczyz:
> > (1) could we use references instead
> 
> References to what? In theory references are supported, we could implement a getter that returns a reference, but most of the language bindings would just clone on the return, which is fine if we're just returning info for one channel.

Was thinking either `Vec<&(NodeId, NodeInfo)>` or `Vec<(NodeId, &NodeInfo)>`. But a look-up function may be preferable, as you noted.

lightning/src/routing/gossip.rs
```diff
@@ -1602,13 +1602,23 @@ impl ReadOnlyNetworkGraph<'_> {
...
+
 	/// Returns all known nodes' public keys along with announced node info.
 	///
 	/// (C-not exported) because we have no mapping for `BTreeMap`s
 	pub fn nodes(&self) -> &BTreeMap<NodeId, NodeInfo> {
 		&*self.nodes
 	}
 
+	/// Returns information on a node with the given id.
+	pub fn get_node(&self, node_id: &NodeId) -> Option<&NodeInfo> {
```
tnull, line 1618: nit: Same here.

tnull (dismissed):
LGTM, mod two style nits.

tnull (approved):

PR #1550, opened 2022-06-18
Allow nodes to be avoided during pathfinding

tnull:
Users may want to - for whatever reasons - prevent payments to be routed over certain nodes. 

This PR therefore allows to add `NodeId`s to a list of banned nodes, which then will be avoided during path finding.


lightning/src/routing/scoring.rs
```diff
@@ -655,6 +668,10 @@ impl<G: Deref<Target = NetworkGraph<L>>, L: Deref, T: Time> Score for Probabilis
 	fn channel_penalty_msat(
 		&self, short_channel_id: u64, source: &NodeId, target: &NodeId, usage: ChannelUsage
 	) -> u64 {
+		if self.banned_nodes.contains(source) || self.banned_nodes.contains(target) {
+			return u64::max_value();
+		}
```
jkczyz, line 673: Could you add a unit test?

lightning/src/routing/scoring.rs
```diff
@@ -1055,7 +1072,8 @@ impl<G: Deref<Target = NetworkGraph<L>>, L: Deref, T: Time> Writeable for Probab
 	#[inline]
 	fn write<W: Writer>(&self, w: &mut W) -> Result<(), io::Error> {
 		write_tlv_fields!(w, {
-			(0, self.channel_liquidities, required)
+			(0, self.channel_liquidities, required),
+			(1, self.banned_nodes, vec_type),
```
jkczyz, line 1076: Do we need/want to serialize these? I'd imagine they can be explicitly added after parsing or passed in as an arg instead. Presumably a list of such nodes is maintained elsewhere.

lightning/src/routing/scoring.rs
```diff
@@ -300,6 +300,7 @@ where L::Target: Logger {
 	logger: L,
 	// TODO: Remove entries of closed channels.
 	channel_liquidities: HashMap<u64, ChannelLiquidity<T>>,
+	banned_nodes: Vec<NodeId>,
```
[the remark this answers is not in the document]
tnull: I now use a `HashSet` with https://github.com/lightningdevkit/rust-lightning/pull/1550/commits/1382696a4a45139553565e06afe9cc7493f97243.

lightning/src/routing/scoring.rs
```diff
@@ -408,6 +410,17 @@ impl<G: Deref<Target = NetworkGraph<L>>, L: Deref, T: Time> ProbabilisticScorerU
 		self
 	}
 
+	/// Marks the node with the given `node_id` as banned, i.e.,
+	/// it will be avoided during path finding.
+	pub fn set_banned(&mut self, node_id: &NodeId) {
```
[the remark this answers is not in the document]
tnull: Added in https://github.com/lightningdevkit/rust-lightning/pull/1550/commits/1382696a4a45139553565e06afe9cc7493f97243.

lightning/src/routing/scoring.rs
```diff
@@ -1069,14 +1087,17 @@ ReadableArgs<(ProbabilisticScoringParameters, G, L)> for ProbabilisticScorerUsin
 	) -> Result<Self, DecodeError> {
 		let (params, network_graph, logger) = args;
 		let mut channel_liquidities = HashMap::new();
+		init_tlv_field_var!(banned_nodes, vec_type);
```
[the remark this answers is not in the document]
tnull: Because of @jkczyz's comment below, I now removed serialization of banned nodes. Makes things easier.

lightning/src/routing/scoring.rs
tnull: Will do!

lightning/src/routing/scoring.rs
tnull: Removed serialization now with https://github.com/lightningdevkit/rust-lightning/pull/1550/commits/1382696a4a45139553565e06afe9cc7493f97243.

lightning/src/routing/scoring.rs
tnull: Added with https://github.com/lightningdevkit/rust-lightning/pull/1550/commits/c1671a5c90a66324d025268c069b502cb6ed97ce.

lightning/src/routing/scoring.rs
```diff
@@ -413,12 +413,17 @@ impl<G: Deref<Target = NetworkGraph<L>>, L: Deref, T: Time> ProbabilisticScorerU
 	/// Marks the node with the given `node_id` as banned, i.e.,
 	/// it will be avoided during path finding.
 	pub fn set_banned(&mut self, node_id: &NodeId) {
```
jkczyz, line 415: How about `add_banned`?

lightning/src/routing/scoring.rs
jkczyz: I'd say keep the methods but leave it out of the config, as you have it, since its more of a dynamic setting. I guess the question would then be if it should be included in the `ReadableArgs` parameter.

lightning/src/routing/scoring.rs
tnull: Done in https://github.com/lightningdevkit/rust-lightning/pull/1550/commits/1e1302e5f4a8b0fa2316ba04fd52c509fe4b727c.

tnull:
Squashed commits for now. Let me know if I should add some serialization of the banned nodes list, or if it's fine as it is.

jkczyz:
> Squashed commits for now. Let me know if I should re-add some serialization of the banned nodes list, or if it's fine as it is.

Argument against serializing the banned nodes is that a crash before serialization would result in the list being inaccurate. So I think it should be maintained independently.

Regarding using the config for the list (https://github.com/lightningdevkit/rust-lightning/pull/1550#discussion_r900967225), we couldn't use a `HashSet` there since it's not supported in bindings. Using a `Vec` would work but would require converting to a `HashMap` internally, so would be odd to have two containers where either the nodes are duplicated or one is empty.

For `new` / `read` (https://github.com/lightningdevkit/rust-lightning/pull/1550#discussion_r901000809), we should probably be consistent in what they accept. So if we don't allow passing a banned list in `new`, then we shouldn't in `read` either.

tnull:
> I would really like to see the map in the config struct - I think it can improve ergonomics a good bit especially for rust users, but being able to set the whole thing up before deserialization instead of after is kinda nice. Its not a big deal though, if we feel strongly to not do that I can live with it. Looks basically good either way.

Alright, I have no strong opinion on this. 
@jkczyz, do you have any strong objection to this? Otherwise I'd just move it to the config? 🤷‍♂️

tnull:
So, just to be on the same page about the way forward:
1. I'll move the `HashSet` as well as the `add`/`remove`/`clear` methods to a field in `UserConfig`. As the `HashSet` initialized empty and will only be filled during runtime, `read` and `new` will be consistent.
2.  I'll add a `add_all` method that can make bindings usage easier and will otherwise tag `HashSet` `(C-not exported)`.

Does that sound about right? Objections/corrections?

jkczyz:
> So, just to be on the same page about the way forward:
> 
> 1. I'll move the `HashSet` as well as the `add`/`remove`/`clear` methods to a field in `UserConfig`. As the `HashSet` initialized empty and will only be filled during runtime, `read` and `new` will be consistent.

`ProbabilisticScoringParameters`, not `UserConfig`, IIUC. Also, the methods need to remain on the scorer since the user won't have access to the params once the scorer is initialized.

tnull:
> > So, just to be on the same page about the way forward:
> > 
> > 1. I'll move the `HashSet` as well as the `add`/`remove`/`clear` methods to a field in `UserConfig`. As the `HashSet` initialized empty and will only be filled during runtime, `read` and `new` will be consistent.
> 
> `ProbabilisticScoringParameters`, not `UserConfig`, IIUC. Also, the methods need to remain on the scorer since the user won't have access to the params once the scorer is initialized.

lol, thanks, I was starting to wonder what was referred to as 'config' and how I'd access `UserConfig`. The `params` struct makes more sense. Will update accordingly.

lightning/src/routing/scoring.rs
```diff
@@ -300,14 +300,13 @@ where L::Target: Logger {
...
 	// TODO: Remove entries of closed channels.
 	channel_liquidities: HashMap<u64, ChannelLiquidity<T>>,
-	banned_nodes: HashSet<NodeId>,
 }
 
 /// Parameters for configuring [`ProbabilisticScorer`].
 ///
 /// Used to configure base, liquidity, and amount penalties, the sum of which comprises the channel
 /// penalty (i.e., the amount in msats willing to be paid to avoid routing through the channel).
-#[derive(Clone, Copy)]
```
tnull, line 309: Unfortunately, adding the `HashSet` here makes this only moveable.

tnull: Yes, I of course left `Clone`. Sorry if that was confusing.

lightning/src/routing/scoring.rs
```diff
@@ -693,7 +706,7 @@ impl<G: Deref<Target = NetworkGraph<L>>, L: Deref, T: Time> Score for Probabilis
 			.get(&short_channel_id)
 			.unwrap_or(&ChannelLiquidity::new())
 			.as_directed(source, target, capacity_msat, liquidity_offset_half_life)
-			.penalty_msat(amount_msat, self.params)
+			.penalty_msat(amount_msat, self.params.clone())
```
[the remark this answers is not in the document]
tnull: Done with https://github.com/lightningdevkit/rust-lightning/pull/1550/commits/c69f400f2d22d3f34aaa4d74d5936217e5d4e922.

tnull:
> Looks like this needs rebase?

Yup, on it.

tnull:
Rebased!

tnull:
Squashed.

jkczyz (approved):

PR #1572, opened 2022-06-26
Separate funding signing into BroadcastFunding client event

### LINKED ISSUE #1571: Move funding signing after commitment
An external signer wants to see that the commitment transaction was signed by the counterparty before releasing the funding signature.  Without the commitment counter-signature, the signer cannot be sure that the channel funds won't be locked forever.

Since the commitment tx depends on the funding tx, we need to separate funding generation from funding signing:

1. generate the funding transaction
2. generate the commitment transaction
3. have the commitment transaction signed by the counterparty
4. provide the signature to the signer via `validate_holder_commitment`
5. sign the funding transaction

This could be done with a new type of `Event` callback.

devrandom:
fixes #1571


devrandom:
Maybe in that case it would be best to have a layer-1 signer?  as part of `KeysInterface`?  or separate?  and fold `spend_spendable_outputs` into it?

devrandom:
> Ideally I think that'd be `Sign` not `KeysInterface`...

An advantage to having the new call in `KeysInterface` is that it is a more natural API for one funding transaction to fund multiple channels.

devrandom:
Either way, there needs to be some node-level coordination in the signer when doing a batch funding, to make sure all of the outputs are understood and accounted for.  Also related is the handling the signing part of `spend_spendable_outputs` - it's at the node level, with some dispatching to channels.

For example, in VLS nodes and channels point to each-other for this kind of coordination.  Checking a funding outpoint is just a matter of looking the outpoint up in the channel list.

BTW, I'm envisioning that `funding_transaction_generated` would get a collection of temp channel IDs to DRY the submission of a multi-channel funding tx.

PR #1575, opened 2022-06-27
derive Hash for Invoice

NicolaLS:
It would be nice to have Hash derived for Invoice. The problem was that `RecoverableSignature` does not implement Hash but I also opened a [PR](https://github.com/rust-bitcoin/rust-secp256k1/pull/462) there so the 'by hand' implementation for `InvoiceSignature` could be removed.

### DIFF
--- a/lightning-invoice/src/lib.rs
+++ b/lightning-invoice/src/lib.rs
@@ -239,7 +239,7 @@ pub struct InvoiceBuilder<D: tb::Bool, H: tb::Bool, T: tb::Bool, C: tb::Bool, S:
 ///  1. using `InvoiceBuilder`
 ///  2. using `Invoice::from_signed(SignedRawInvoice)`
 ///  3. using `str::parse::<Invoice>(&str)`
-#[derive(Eq, PartialEq, Debug, Clone)]
+#[derive(Eq, PartialEq, Debug, Clone, Hash)]
 pub struct Invoice {
 	signed_invoice: SignedRawInvoice,
 }
@@ -263,7 +263,7 @@ pub enum InvoiceDescription<'f> {
 ///
 /// # Invariants
 /// The hash has to be either from the deserialized invoice or from the serialized `raw_invoice`.
-#[derive(Eq, PartialEq, Debug, Clone)]
+#[derive(Eq, PartialEq, Debug, Clone, Hash)]
 pub struct SignedRawInvoice {
 	/// The rawInvoice that the signature belongs to
 	raw_invoice: RawInvoice,
@@ -286,7 +286,7 @@ pub struct SignedRawInvoice {
 /// De- and encoding should not lead to information loss but may lead to different hashes.
 ///
 /// For methods without docs see the corresponding methods in `Invoice`.
-#[derive(Eq, PartialEq, Debug, Clone)]
+#[derive(Eq, PartialEq, Debug, Clone, Hash)]
 pub struct RawInvoice {
 	/// human readable part
 	pub hrp: RawHrp,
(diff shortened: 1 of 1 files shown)


lightning-invoice/src/lib.rs
```diff
@@ -1415,6 +1415,15 @@ impl Deref for SignedRawInvoice {
 	}
 }
 
+impl std::hash::Hash for InvoiceSignature {
+	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+		let (recovery_id, sig) = self.0.serialize_compact();
+		let recovery_id = recovery_id.to_i32();
+		state.write(&sig);
+		state.write_i32(recovery_id);
```
NicolaLS, line 1423: I'm not sure if it is ok to do it like this. In secp256k1 they already have a Hash impl for the contained type in ffi `pub struct RecoverableSignature(ffi::RecoverableSignature);` but they don't derive Hash on that. If ffi was public the correct way would be `(&self.0.0[..]).hash(state)` but it is not possible

lightning-invoice/src/lib.rs
```diff
@@ -1415,6 +1415,15 @@ impl Deref for SignedRawInvoice {
 	}
 }
 
+impl core::hash::Hash for InvoiceSignature {
```
jkczyz, line 1418: Looks like clippy doesn't like this

```
error: you are implementing `Hash` explicitly but have derived `PartialEq`
    --> lightning-invoice/src/lib.rs:1418:1
     |
1418 | / impl core::hash::Hash for InvoiceSignature {
1419 | |     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
1420 | |         let (recovery_id, sig) = self.0.serialize_compact();
1421 | |         let recovery_id = recovery_id.to_i32();
...    |
1424 | |     }
1425 | | }
     | |_^
     |
     = note: `#[deny(clippy::derive_hash_xor_eq)]` on by default
note: `PartialEq` implemented here
    --> lightning-invoice/src/lib.rs:456:28
     |
456  | #[derive(Clone, Debug, Eq, PartialEq)]
     |                            ^^^^^^^^^
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq
     = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

NicolaLS: Ok I'll leave it open, thank you

elsirion: I'd argue clippy isn't dumb here. The contract of `Hash` states that hashes have to be equal if `Eq` returns equal for the hashed objects. With this implementation we'd rely on libsecp256k1 to not use the higher bits of the recovery id byte for other things for that contract to hold. Related to https://github.com/rust-bitcoin/rust-secp256k1/issues/463. The `Hash` impl from `secp256k1` would be guaranteed to be at least internally consistent (just not between versions I guess).

NicolaLS:
I think it's ready [now](https://github.com/rust-bitcoin/rust-secp256k1/commit/3f99ae8f1f0ae345c428ea10ae09ad36434e1b1d) @TheBlueMatt

NicolaLS:
> I think it's ready [now](https://github.com/rust-bitcoin/rust-secp256k1/commit/3f99ae8f1f0ae345c428ea10ae09ad36434e1b1d) @TheBlueMatt

Oh wait it's until ldk updates to the version, not only when it's released so nvm

dunxen:
@NicolaLS, the upstream PRs are in and I see rust-bitcoin 0.29 does include https://github.com/rust-bitcoin/rust-bitcoin/pull/1066. So looks like we're unblocked completely now :)

dunxen (approved):
Thanks, pretty neat now!

NicolaLS:
> Thanks, pretty neat now!

Is it a problem that some CI failed ?

dunxen:
> Is it a problem that some CI failed ?

Those are unrelated to the project. It’s a crates registry issue (I think).

PR #1591, opened 2022-07-03
Implement `Readable` and `Writeable` traits for `Invoice`

meryacine:
So that invoices could be sent over wires.

### DIFF
--- a/lightning-invoice/src/lib.rs
+++ b/lightning-invoice/src/lib.rs
@@ -50,6 +50,9 @@ use lightning::ln::features::InvoiceFeatures;
 use lightning::routing::gossip::RoutingFees;
 use lightning::routing::router::RouteHint;
 use lightning::util::invoice::construct_invoice_preimage;
+use lightning::util::ser::{Readable, Writeable, Writer};
+use lightning::ln::msgs;
+use lightning::io::Read;
 
 use secp256k1::PublicKey;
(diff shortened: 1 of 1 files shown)


meryacine:
> What's the use-case for this?

Sending invoices inside lightning messages, custom lightning messages.
> Do you have something where you need a generic across different LDK types?

Sorry, not getting your question :(

meryacine:
> Sorry, my question was "does it make sense for you to do the string serialization and then shove it into wherever needs writeable or does it need to be Readable/Writeable".

This would work fine. I was thinking it looks cleaner though to have an invoice struct instead.

> I'm somewhat uneasy implementing Readable/Writeable for a lightning type where we're not doing something standardized - what happens if there's some future extension to the BOLTs that mandates BOLT 11 invoices be encoded....some other way (eg not bech32)?

ATM there is no standard for BOLT 11 messages, so whoever gonna use it as readable/writeable will be for their custom lightning messages. If for some reason in the future BOLT 11 was included in a standard lightning message and had to have a standard encoding, I don't think it will be a problem to just change the encoding to the standard one (at the end, users only care about their messages being sendable and not the underlying encoding). The peers using BOLT 11 in their custom messages will need to pin the same lightning-invoice version to keep their custom messages compatible.

meryacine:
Closing for possible backward compatibility violation.

PR #1596, opened 2022-07-05
Make ChannelMonitor aware of counterparty's node id

jurvis:
In #1403, we want to be able to send an `HTLCHandlingFailed` event, which has an enum field, `next_hop_destination: HTLCDestination`, that defines the reason for the error. 

One of the variants is `HTLCDestination::NextHopChannel`, which describes the scenario where we tried forwarding to a channel but failed to do so. 

When processing a channel monitor's events, we can possibly run into a scenario where we are unable to process an HTLC because the channel has closed. `ChannelManager` will also not be aware of the channel as a result, and we will not be able to look it up to say who the HTLC was meant for. 

Therefore, the only way to define the channel counterparty in `HTLCDestination::NextHopChannel` is by making `ChannelMonitor` aware of `counterparty_node_id`, and piping it through `MonitorEvent` for it to be used.

Related discussion: https://github.com/lightningdevkit/rust-lightning/pull/1403#discussion_r908520656

### DIFF
--- a/lightning/src/chain/channelmonitor.rs
+++ b/lightning/src/chain/channelmonitor.rs
@@ -722,6 +722,9 @@ pub(crate) struct ChannelMonitorImpl<Signer: Sign> {
 	// the full block_connected).
 	best_block: BestBlock,
 
+	/// The node_id of our counterparty
+	counterparty_node_id: Option<PublicKey>,
+
 	secp_ctx: Secp256k1<secp256k1::All>, //TODO: dedup this a bit...
 }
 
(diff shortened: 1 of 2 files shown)


jurvis:
@TheBlueMatt saw that I used a type number that is not consistent in read/write_tlv_fields, so I fixed that and repushed

jkczyz:
Could you update the commit message / PR description for the reason for this change?

dunxen (approved):
ACK 2d4933

jurvis:
@jkczyz done :)

jkczyz (approved):

jurvis:
@TheBlueMatt gotcha, my apologies. Will note that for next time :)

Issue #1498, opened 2022-05-26
make channel reserve configurable


moneyball:
Do we allow 0-1%? I think so because we accept 0 channel reserve for counterparty, which implies we expect and support such a use case (with associated security tradeoffs).

PR #1619, opened 2022-07-15
Add configuration support for 'their_channel_reserve_satoshis' [#1498]

G8XSU:
Add configuration support for 'their_channel_reserve_satoshis' [#1498] while channel open/accept
```
`their_channel_reserve_satoshis` is the minimum amount that the other node is to keep as a direct payment.
This ensures that if our counterparty broadcasts a revoked state, we can punish them by
claiming at least this value on chain.

`their_channel_reserve_proportion_millionth` is for us to configure counterparty's channel reserve
for both outgoing and incoming channels.

channel_reserve values greater than 30% could be considered highly unreasonable, since that
amount can never be used for payments.
Also, if our selected channel_reserve for counterparty and counterparty's selected channel_reserve
for us, sum upto or greater than channel_value, channel negotiations will fail.

`their_channel_reserve_proportion_millionth` is millionths of a channel_value to use as `their_channel_reserve_satoshis`.

Default Value is 1% of channel_value, configured as 10,000 millionth
Lower Bound on `their_channel_reserve_satoshis` is 1000 sats provided as safe implementation limit.
Max `their_channel_reserve_satoshis` can be equal to channel_value i.e 1000,000 millionth
but negotiations will fail in that case.
```


lightning/src/util/config.rs
```diff
@@ -126,6 +126,26 @@ pub struct ChannelHandshakeConfig {
 	///
 	/// [`KeysInterface::get_shutdown_scriptpubkey`]: crate::chain::keysinterface::KeysInterface::get_shutdown_scriptpubkey
 	pub commit_upfront_shutdown_pubkey: bool,
+
+	/// `their_channel_reserve_satoshis` is the minimum amount that the other node is to keep as a direct payment.
```
ViktorTigerstrom, line 130: I personally find it a bit confusing that this text starts by explaining another value than the one explicitly configured, even though it's related to the value that's configured. Would it make sense to start this text with something like:

Sets the proportion that of the `channel_value` that `their_channel_reserve_satoshis` is set to for both outgoing and incoming channels. `their_channel_reserve_satoshis` is the minimum amount ...

lightning/src/util/config.rs
```diff
@@ -126,6 +126,26 @@ pub struct ChannelHandshakeConfig {
...
+	/// for both outgoing and incoming channels.
+	///
+	/// channel_reserve values greater than 30% could be considered highly unreasonable, since that
+	/// amount can never be used for payments.
+	/// Also, if our selected channel_reserve for counterparty and counterparty's selected channel_reserve
+	/// for us, sum upto or greater than channel_value, channel negotiations will fail.
+	///
+	/// `their_channel_reserve_proportion_millionth` is millionths of a channel_value to use as `their_channel_reserve_satoshis`.
+	///
+	/// Default Value is 1% of channel_value, configured as 10,000 millionth
```
ViktorTigerstrom, line 144: The text above should mention that any new channels created with the non default value will cause versions of LDK prior to 0.0.104 to refuse to read the `ChannelManager`, similar to how it's mentioned in the explanation for the `max_inbound_htlc_value_in_flight_percent_of_channel` config option.

The reason why that's the case is that `holder_selected_channel_reserve_satoshis` was added with an even number (4) in the tlv fields, and was added in 0.0.104 (though, of course this shouldn't be mentioned in the docs for `their_channel_reserve_proportion_millionth` :) )

lightning/src/util/config.rs
```diff
@@ -126,6 +126,26 @@ pub struct ChannelHandshakeConfig {
...
+	/// Lower Bound on `their_channel_reserve_satoshis` is 1000 sats provided as safe implementation limit.
```
ViktorTigerstrom, line 145: I think this clearer should explain what happens if this value is set to a lower value than 1000, i.e.:
any values less than 1000 will be treated as 1000 instead.
This makes the config text more align with how other bounded values are explained (such as `cltv_expiry_delta` and `max_inbound_htlc_value_in_flight_percent_of_channel`).

Nit:
I also recommend formatting similar to how it's formatted in `cltv_expiry_delta` and `max_inbound_htlc_value_in_flight_percent_of_channel` i.e:
Default value: 10,000 ....
Minimum value: 1000 sats provided as ...
Maximum value: 1000,000 ....

lightning/src/util/config.rs
```diff
@@ -126,6 +126,26 @@ pub struct ChannelHandshakeConfig {
...
+	/// Max `their_channel_reserve_satoshis` can be equal to channel_value i.e 1000,000 millionth
```
ViktorTigerstrom, line 146: I find it very strange that the range is upper bound to value that is guaranteed to fail. If that's the case, then why is it even upper bound at all?
If we want to have an upper bound, wouldn't we at least want make the bound a value that's more likely to succeed?

lightning/src/util/config.rs
```diff
@@ -126,6 +126,26 @@ pub struct ChannelHandshakeConfig {
...
+	/// but negotiations will fail in that case.
+	pub their_channel_reserve_proportion_millionth: u32,
```
ViktorTigerstrom, line 148: A general question:
Do we really need the extreme precision that representing this value as a millionth gives us?
I think it's quite likely that users will find it quite confusing to set this value, as setting values as millionths is not commonly used.

Would instead using something like basis points provide us with enough precision or not? I think that using basis points is likely less confusing from the users perspective as it's more commonly used.

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
...
-	/// required by us.
+	/// required by us according to configured or default `their_channel_reserve_proportion_millionth`
 	///
 	/// Guaranteed to return a value no larger than channel_value_satoshis
 	///
-	/// This is used both for new channels and to figure out what reserve value we sent to peers
-	/// for channels serialized before we included our selected reserve value in the serialized
-	/// data explicitly.
-	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64) -> u64 {
+	/// This is used both for outgoing and incoming channels and has lower bound of `MIN_THEIR_CHAN_RESERVE_SATOSHIS`
```
ViktorTigerstrom, line 853: Nit:
```suggestion
	/// This is used both for outbound and inbound channels and has lower bound of `MIN_THEIR_CHAN_RESERVE_SATOSHIS`
```

lightning/src/ln/channel.rs
```diff
@@ -912,7 +923,7 @@ impl<Signer: Sign> Channel<Signer> {
 		if holder_selected_contest_delay < BREAKDOWN_TIMEOUT {
 			return Err(APIError::APIMisuseError {err: format!("Configured with an unreasonable our_to_self_delay ({}) putting user funds at risks", holder_selected_contest_delay)});
 		}
-		let holder_selected_channel_reserve_satoshis = Channel::<Signer>::get_holder_selected_channel_reserve_satoshis(channel_value_satoshis);
+		let holder_selected_channel_reserve_satoshis = Channel::<Signer>::get_holder_selected_channel_reserve_satoshis(channel_value_satoshis, config);
 		if holder_selected_channel_reserve_satoshis < MIN_CHAN_DUST_LIMIT_SATOSHIS {
```
ViktorTigerstrom, line 927: This and the other similar `MIN_CHAN_DUST_LIMIT_SATOSHIS` check is now unreachable as the `holder_selected_channel_reserve_satoshis` is lower bound to 1000 now. I therefore suggest removing the two checks.

ViktorTigerstrom (commented):
Nice, generally the code looks good :)! I've yet to review the tests though.
I've left a few suggestions of improvements below.

lightning/src/util/config.rs
G8XSU: While basis points might provide right/enough precision for this use case, i think we should have a standard way of representing proportions. Millionths was being used at some of other places and i didn't see basis points being used in code. 
I agree on the less confusing part.

lightning/src/ln/channel.rs
```diff
@@ -795,6 +796,8 @@ pub const MAX_CHAN_DUST_LIMIT_SATOSHIS: u64 = MAX_STD_OUTPUT_DUST_LIMIT_SATOSHIS
 /// See https://github.com/lightning/bolts/issues/905 for more details.
 pub const MIN_CHAN_DUST_LIMIT_SATOSHIS: u64 = 354;
 
+pub const MIN_THEIR_CHAN_RESERVE_SATOSHIS: u64 = 1000;
```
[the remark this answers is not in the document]
G8XSU: I don't have an exact answer for this, current minimum was 1000, i just extracted it to constant and made it bit more explicit.
I think its just a reasonable value greater than dust limit.

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
 	}
 
 	/// Returns a minimum channel reserve value the remote needs to maintain,
-	/// required by us.
+	/// required by us according to configured or default `their_channel_reserve_proportion_millionth`
```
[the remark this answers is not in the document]
G8XSU: done 👍

lightning/src/ln/channel.rs
G8XSU: i think those two are separate things and different safety checks. 
MIN_THEIR_CHAN_RESERVE_SATOSHIS is just a safe implementation lower bound, can be changed later. 
Check for MIN_CHAN_DUST_LIMIT_SATOSHIS is protocol limit, and must have a check for it.

lightning/src/ln/channel.rs
```diff
@@ -46,6 +46,7 @@ use io;
 use prelude::*;
 use core::{cmp,mem,fmt};
 use core::ops::Deref;
+use std::ops::Mul;
```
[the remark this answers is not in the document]
G8XSU: thanks, will keep it in mind

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
...
 	///
 	/// Guaranteed to return a value no larger than channel_value_satoshis
 	///
-	/// This is used both for new channels and to figure out what reserve value we sent to peers
-	/// for channels serialized before we included our selected reserve value in the serialized
-	/// data explicitly.
-	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64) -> u64 {
+	/// This is used both for outgoing and incoming channels and has lower bound of `MIN_THEIR_CHAN_RESERVE_SATOSHIS`
+	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64, config: &UserConfig) -> u64 {
+		let their_channel_reserve_proportion = (config.channel_handshake_config.their_channel_reserve_proportion_millionth as f64) / 1000_000.0;
```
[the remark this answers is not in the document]
G8XSU: done

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
...
+		let calculated_reserve = their_channel_reserve_proportion.mul(channel_value_satoshis as f64) as u64;
```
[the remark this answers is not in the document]
G8XSU: will be using 
```
let calculated_reserve = channel_value_satoshis.saturating_mul(config.channel_handshake_config.their_channel_reserve_proportion_millionth as u64) / 1000_000;
```
to avoid floating point arithmetic.

lightning/src/util/config.rs
G8XSU: done 👍

lightning/src/util/config.rs
G8XSU: yes the idea was to handle unreasonable reserve configs by docs, because technically even a 99.99% reserve might work. and 60-30 channel_reserve b/w parties is unreasonable but valid.

lightning/src/util/config.rs
ViktorTigerstrom: Good point :)!

lightning/src/util/config.rs
ViktorTigerstrom: Ok I still find it a bit strange to bound the value to an upper bound that's guaranteed to fail, instead of just not having any bound at all and just let the channel negotiations fail if the user sets a too high value. But if others are ok with it, I am as well :).

lightning/src/util/config.rs
```diff
@@ -126,6 +126,29 @@ pub struct ChannelHandshakeConfig {
 	///
 	/// [`KeysInterface::get_shutdown_scriptpubkey`]: crate::chain::keysinterface::KeysInterface::get_shutdown_scriptpubkey
 	pub commit_upfront_shutdown_pubkey: bool,
+
+	/// `their_channel_reserve_proportion_millionth` is proportion of channel_value
```
ViktorTigerstrom, line 130: Nit: 
```suggestion
	/// `their_channel_reserve_proportion_millionth` is the proportion of the channel_value
```

Pardon me if the current text is correct grammar already, as english isn't my native language. Feel free to ignore this nit if thats the case :).

lightning/src/util/config.rs
```diff
@@ -126,6 +126,29 @@ pub struct ChannelHandshakeConfig {
...
+	/// for us, sum upto or greater than channel_value, channel negotiations will fail.
+	///
+	/// Note: Versions of LDK earlier than v0.0.104 will fail to read channels with any channel_reserve
+	/// different than default value.
+	///
+	/// Default value: is 1% of channel_value, configured as 10,000 millionth
+	/// Minimum value: If calculated proportional value is < 1000 sats, will be treated as 1000 sats instead,
+	/// 				1000 sats is just a safe implementation-specific lower bound.
+	/// Maximum value: 1000,000, any values larger than 1Million will be treated as 1Million (or 100%) instead,
+	/// although channel negotiations will probably fail in that case.
```
ViktorTigerstrom, line 150: Nit:
```suggestion
	/// although channel negotiations will fail in that case.
```

It's guaranteed to fail as the `channel_reserve` is required to be set to the at least the dust limit by both parties.

lightning/src/ln/channel.rs
```diff
@@ -6962,6 +6974,63 @@ mod tests {
...
+	#[test]
+	fn test_configured_holder_selected_channel_reserve_satoshis() {
+
+		// Test that `new_outbound` and `new_from_req` creates a channel with the correct value for
+		// `their_channel_reserve_proportion`, when configured with a valid percentage value,
+		// which is set to (2%) of the `channel_value` and is greater than lower bound `MIN_THEIR_CHAN_RESERVE_SATOSHIS`.
+		test_self_and_counterparty_channel_reserve(10000000, 0.02, 0.02);
+
+		// Test with valid but unreasonably high channel reserve
+		// Requesting and accepting party have requested for 49%-49% and 60%-40% channel reserve
```
ViktorTigerstrom, line 6986: Nit:
```suggestion
		// Requesting and accepting party have requested for 49%-49% and 60%-30% channel reserve
```

ViktorTigerstrom (approved):
Tests looks good!
LGTM except a few nits.

lightning/src/util/config.rs
G8XSU: i think "`their_channel_reserve_proportion_millionth` is proportion of the channel_value" makes sense acc. to casual english.

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	///
+	/// `their_channel_reserve_satoshis` is the minimum amount that the other node is to keep as a
+	/// direct payment.
+	/// This ensures that if our counterparty broadcasts a revoked state, we can punish them by
+	/// claiming at least this value on chain.
+	///
+	/// channel_reserve values greater than 30% could be considered highly unreasonable, since that
+	/// amount can never be used for payments.
+	/// Also, if our selected channel_reserve for counterparty and counterparty's selected
+	/// channel_reserve for us, sum upto or greater than channel_value, channel negotiations will fail.
```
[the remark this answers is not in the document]
G8XSU: the check is already present in accept_channel which behaves similarly, channel.rs:2022
```
if msg.channel_reserve_satoshis > self.channel_value_satoshis - self.holder_selected_channel_reserve_satoshis {
			return Err(ChannelError::Close(format!("Bogus channel_reserve_satoshis ({}). Must not be greater than channel value minus our reserve ({})",
				msg.channel_reserve_satoshis, self.channel_value_satoshis - self.holder_selected_channel_reserve_satoshis)));
		}
```

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
 	///
 	/// [`KeysInterface::get_shutdown_scriptpubkey`]: crate::chain::keysinterface::KeysInterface::get_shutdown_scriptpubkey
 	pub commit_upfront_shutdown_pubkey: bool,
+
+	/// `their_channel_reserve_proportion_millionth` is proportion of the channel_value to configure
```
tnull, line 130: nit: No need to repeat the field name here. It's also probably generally advisable to reduce the use of variable notation in the public API documentation, if you do not mean to link something specific. So instead of, for example, `channel_value`, just write "value of the channel".
```suggestion
	/// The proportion of the channel value to configure as the
```

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// counterparty's channel_reserve i.e `their_channel_reserve_satoshis` for both outbound and
```
tnull, line 131:
```suggestion
	/// counterparty's channel reserve, i.e., `their_channel_reserve_satoshis` for both outbound and
```

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// inbound channels.
+	///
+	/// `their_channel_reserve_satoshis` is the minimum amount that the other node is to keep as a
```
tnull, line 134:
```suggestion
	/// `their_channel_reserve_satoshis` is the minimum balance that the other node
```

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// direct payment.
```
tnull, line 135:
```suggestion
	///  has to maintain at all times.
```

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// This ensures that if our counterparty broadcasts a revoked state, we can punish them by
+	/// claiming at least this value on chain.
+	///
+	/// channel_reserve values greater than 30% could be considered highly unreasonable, since that
+	/// amount can never be used for payments.
+	/// Also, if our selected channel_reserve for counterparty and counterparty's selected
```
tnull, line 141:
```suggestion
	/// Also, if our selected channel reserve for the counterparty and the counterparty's selected
```

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
...
 	///
 	/// Guaranteed to return a value no larger than channel_value_satoshis
 	///
-	/// This is used both for new channels and to figure out what reserve value we sent to peers
-	/// for channels serialized before we included our selected reserve value in the serialized
-	/// data explicitly.
-	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64) -> u64 {
+	/// This is used both for outbound and inbound channels and has lower bound of `MIN_THEIR_CHAN_RESERVE_SATOSHIS`
+	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64, config: &UserConfig) -> u64 {
+		let calculated_reserve = channel_value_satoshis.saturating_mul(config.channel_handshake_config.their_channel_reserve_proportion_millionth as u64) / 1000_000;
```
tnull, line 856:
```suggestion
		let their_channel_reserve = channel_value_satoshis.saturating_mul(config.channel_handshake_config.their_channel_reserve_proportion_millionth as u64) / 1_000_000;
```

tnull (commented):
Some doc nits and suggestions.

lightning/src/util/config.rs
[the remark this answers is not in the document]
G8XSU: i believe "sum up to target" is correct usage

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// channel_reserve for us, sum upto or greater than channel_value, channel negotiations will fail.
+	///
+	/// Note: Versions of LDK earlier than v0.0.104 will fail to read channels with any channel_reserve
+	/// different than default value.
+	///
+	/// Default value: is 1% of channel_value, configured as 10,000 millionths
+	/// Minimum value: If calculated proportional value is < 1000 sats, will be treated as 1000 sats
+	///    				instead, 1000 sats is just a safe implementation-specific lower bound.
+	/// Maximum value: 1,000,000, any values larger than 1Million will be treated as 1Million (or 100%)
```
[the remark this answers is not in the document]
G8XSU: i think there is information loss in this, will be keeping current version. 
As for other nits, i have addressed most of them which seemed legible.
And I don't think using grammar check on PR is legal :P

G8XSU:
> Can you rewrite the commit message to stay under 80 chars wide? Looks like its currently wrapped at 90?

Done

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// Also, if our selected channel reserve for counterparty and counterparty's selected
+	/// channel reserve for us sum up to equal or greater than channel value, channel negotiations
+	/// will fail.
+	///
+	/// Note: Versions of LDK earlier than v0.0.104 will fail to read channels with any channel reserve
+	/// other than the default value.
+	///
+	/// Default value: 1% of channel value, i.e., configured as 10,000 millionths.
+	/// Minimum value: If the calculated proportional value is less than 1000 sats, it will be treated
+	///					 as 1000 sats instead, which is a safe implementation-specific lower bound.
```
ViktorTigerstrom, line 149: This, as well as the indentation under "Maximum value" are incorrectly indented. For text alignment specifically (after the ///), you need to use `spaces` only and not `tabs` (which is contradictory as we indent the rest of the project with tabs).
This has a mix between indentations of tabs and spaces.

ViktorTigerstrom (commented):
Some final docs nits before I'm ACK, thanks for bearing with us :)!

lightning/src/ln/functional_tests.rs
```diff
@@ -1771,10 +1774,11 @@ fn test_inbound_outbound_capacity_is_not_zero() {
 	let _ = create_announced_chan_between_nodes_with_value(&nodes, 0, 1, 100000, 95000000, InitFeatures::known(), InitFeatures::known());
 	let channels0 = node_chanmgrs[0].list_channels();
 	let channels1 = node_chanmgrs[1].list_channels();
+	let default_config = UserConfig::default();
 	assert_eq!(channels0.len(), 1);
 	assert_eq!(channels1.len(), 1);
 
-	let reserve = Channel::<EnforcingSigner>::get_holder_selected_channel_reserve_satoshis(100000);
+	let reserve = Channel::<EnforcingSigner>::get_holder_selected_channel_reserve_satoshis(100000, &default_config);
```
tnull, line 1781:
```suggestion
	let reserve = Channel::<EnforcingSigner>::get_holder_selected_channel_reserve_satoshis(100_000, &default_config);
```

lightning/src/ln/channel.rs
```diff
@@ -6962,6 +6976,63 @@ mod tests {
...
+		let mut outbound_node_config = UserConfig::default();
+		outbound_node_config.channel_handshake_config.their_channel_reserve_proportion_millionth = (outbound_selected_channel_reserve_perc * 1000_000.0) as u32;
+		let chan = Channel::<EnforcingSigner>::new_outbound(&&fee_est, &&keys_provider, outbound_node_id, &InitFeatures::known(), channel_value_satoshis, 100000, 42, &outbound_node_config, 0, 42).unwrap();
+
+		let expected_outbound_selected_chan_reserve = cmp::max(MIN_THEIR_CHAN_RESERVE_SATOSHIS, (chan.channel_value_satoshis as f64 * outbound_selected_channel_reserve_perc) as u64);
+		assert_eq!(chan.holder_selected_channel_reserve_satoshis, expected_outbound_selected_chan_reserve);
+
+		let chan_open_channel_msg = chan.get_open_channel(genesis_block(network).header.block_hash());
+		let mut inbound_node_config = UserConfig::default();
+		inbound_node_config.channel_handshake_config.their_channel_reserve_proportion_millionth = (inbound_selected_channel_reserve_perc * 1000_000.0) as u32;
```
tnull, line 7020:
```suggestion
		inbound_node_config.channel_handshake_config.their_channel_reserve_proportion_millionth = (inbound_selected_channel_reserve_perc * 1_000_000.0) as u32;
```

lightning/src/ln/channel.rs
```diff
@@ -6962,6 +6976,63 @@ mod tests {
...
+		let seed = [42; 32];
+		let network = Network::Testnet;
+		let keys_provider = test_utils::TestKeysInterface::new(&seed, network);
+		let outbound_node_id = PublicKey::from_secret_key(&secp_ctx, &SecretKey::from_slice(&[42; 32]).unwrap());
+		let inbound_node_id = PublicKey::from_secret_key(&secp_ctx, &SecretKey::from_slice(&[7; 32]).unwrap());
+
+
+		let mut outbound_node_config = UserConfig::default();
+		outbound_node_config.channel_handshake_config.their_channel_reserve_proportion_millionth = (outbound_selected_channel_reserve_perc * 1000_000.0) as u32;
+		let chan = Channel::<EnforcingSigner>::new_outbound(&&fee_est, &&keys_provider, outbound_node_id, &InitFeatures::known(), channel_value_satoshis, 100000, 42, &outbound_node_config, 0, 42).unwrap();
```
tnull, line 7013:
```suggestion
		let chan = Channel::<EnforcingSigner>::new_outbound(&&fee_est, &&keys_provider, outbound_node_id, &InitFeatures::known(), channel_value_satoshis, 100_000, 42, &outbound_node_config, 0, 42).unwrap();
```

tnull (commented):
Generally looks good, just one more nit and a few minor suggestions.

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// Maximum value: 1,000,000, any values larger than 1 Million will be treated as 1 Million (or 100%)
+	///    			    instead, although channel negotiations will fail in that case.
+	pub their_channel_reserve_proportion_millionth: u32
```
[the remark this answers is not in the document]
G8XSU: Oh, i think matt mentioned it earlier as well. I might have misunderstood, i thought its for comments only. Will pluralize  field name as well.

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
...
+	/// required by us according to the configured or default
+	/// [`ChannelHandshakeConfig::their_channel_reserve_proportion_millionth`]
 	///
 	/// Guaranteed to return a value no larger than channel_value_satoshis
 	///
-	/// This is used both for new channels and to figure out what reserve value we sent to peers
-	/// for channels serialized before we included our selected reserve value in the serialized
-	/// data explicitly.
-	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64) -> u64 {
+	/// This is used both for outbound and inbound channels and has lower bound of `MIN_THEIR_CHAN_RESERVE_SATOSHIS`
```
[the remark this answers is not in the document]
G8XSU: * Missing article before countable noun is treated as optional and acceptable grammar in most IDE's for code documentation. (default:disabled in intellij)
* Same with period before paragraph end (default:disabled in intellij)
* Same for ',' after e.g or i.e  (default:disabled in intellij)

I am currently using default Intellij grammar for "English(USA)" and it relaxes for these conditions by default. 
I guess it could vary a little based on keyboard selection of user.

We are really discussing these IDE settings in PR's, which i don't think is worth our time in PR.
We can standardize them, if we want.
Either we stick to default acceptable grammar or we can decide to just enable those optional validation checks. 

At the same time, we don't want to discard all grammar and write illegible docs. We should work towards having acceptable grammar and legible docs.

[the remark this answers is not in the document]
tnull: > We are really discussing these IDE settings in PR's, which i don't think is worth our time in PR.
> We can standardize them, if we want.
> Either we stick to default acceptable grammar or we can decide to just enable those optional validation checks.

I tend to disagree. There are best practices to write documentation in full, grammatically correct sentences, and we should do our best to follow these. I think whether a particular IDE picks up on some of the details doesn't matter all too much. 

I also think discussing subtleties of wording etc. is time well spent in a PR, since it can make or break a good documentation **and** most importantly can reveal diverging understandings of the subject at hand.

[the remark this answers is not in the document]
ViktorTigerstrom: Given that the rest of the docs in the codebase is mixed with how strict the we are with period before paragraphs ends
and similar, I can agree on letting it slip :).
Though I also tend to agree with @tnull here. I don't see why we shouldn't make our best efforts with the grammar in new PRs if it's quick to fix and doesn't take much time, given that suggested grammar is _more correct_. IMO when it tends to just be a waste of time and be too nit picky is when time is spent on discussing specific words or sentences that are _equally correct_, and brings no extra clarity.

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,24 @@ impl<Signer: Sign> Channel<Signer> {
...
+	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64, config: &UserConfig) -> u64 {
+		let calculated_reserve = channel_value_satoshis.saturating_mul(config.channel_handshake_config.their_channel_reserve_proportion_millionth as u64) / 1_000_000;
+		cmp::min(channel_value_satoshis, cmp::max(calculated_reserve, MIN_THEIR_CHAN_RESERVE_SATOSHIS))
+	}
+
+	/// This is for legacy reasons, present for forward-compatibility. LDK versions older than " < 0.0.104"
+	/// don't know how read/handle values other than default from storage. Hence we use this function
```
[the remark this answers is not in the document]
ViktorTigerstrom: Yeah fair enough :)! I agree on that my suggestion isn't correct either.

What I specifically find confusing is the sentence:
"LDK versions older than " < 0.0.104" don't know how read/handle values other than default from storage.",
as those versions under no circumstances read the value from the storage at all, not even the default value. But, given the discussion about being too nit picky above, maybe that applies to me here as well 😁.

If @G8XSU thinks changing the wording here would clarify the docs, feel free to change this. Either way, I'm ok with the current wording here as well :)!

lightning/src/ln/channel.rs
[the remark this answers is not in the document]
G8XSU: Its not about whether IDE picks it up or not, it has the functionality but the chosen default:disabled means these rules are considered flexible and acceptable. 
I am up for discussing any subtleties, wordings or grammar. But for grammar, saying that only "British English" official rules are "correct" is far from reality. 

And I doubt adding a comma after "e.g." makes or breaks a doc. We are not writing a book here. We need to be practical and subjectively decide how strict we are. Specifically when it adds no additional value in understanding.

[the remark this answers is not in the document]
tnull: Mh, I think we're discussing two different things here. Should there be a considerable amount of time spent on splitting hairs regarding individual commas and/or missing articles? No, probably not.

But, I still think reviewers *should* point anything out that they think would improve the quality of the PR, even if it wouldn't warrant having a long discussion about it. Speaking for myself, most of the nits are really meant as suggestions and/or questions, which, at the end of the day, the author may or may not choose to consider.

lightning/src/ln/channelmanager.rs
```diff
@@ -1635,7 +1635,7 @@ impl<Signer: Sign, M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelMana
 		}
 	}
 
-	/// Gets the current configuration applied to all new channels,  as
+	/// Gets the current configuration applied to all new channels
```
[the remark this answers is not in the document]
G8XSU: 💯

lightning/src/ln/channel.rs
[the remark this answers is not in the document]
G8XSU: Main purpose: Its present for forward-compatibility, so that we don't persist default values for channels created by older versions. This is for the case when we have to roll-back to previous version, we have at-least those old channels operational.

If in future, we decide to remove forward compat with <104, we can get rid of this function/logic. 
I think this information is conveyed.
Didn't want to make it anymore complex than that.

lightning/src/ln/channel.rs
```diff
@@ -6962,6 +6979,64 @@ mod tests {
 		assert_eq!(chan_8.holder_max_htlc_value_in_flight_msat, chan_8_value_msat);
 	}
 
+	#[test]
+	fn test_configured_holder_selected_channel_reserve_satoshis() {
+
+		// Test that `new_outbound` and `new_from_req` creates a channel with the correct value set
+		// for `their_channel_reserve_proportion`, when configured with a valid percentage value.
```
ViktorTigerstrom, line 6986: I was trying to convey in my previous suggestion https://github.com/lightningdevkit/rust-lightning/pull/1619#discussion_r927005960 that I think this comment is a bit confusing, but I realize that I probably wasn't wasn't very clear :) :
The test asserts that the `channel_reserve`s are set to the correct value when the config value is configured with a valid value, and I think it would make more sense to word the comment to covey that instead of referring to that the config value has a correct value set. The test asserts the `channel_reserve`s values, and doesn't explicitly assert that: "`new_outbound` and `new_from_req` creates a channel with the correct value set for `their_channel_reserve_proportion`", although that is of course implicitly is correct of as well.

This isn't a big of a deal, but feel free to change this if you'd like. Either way I'm ok :).

ViktorTigerstrom (approved):
ACK eb620990ceef941fcc29a574543685a9fb9b8909

Although I would like to raise that @tnull's feedback https://github.com/lightningdevkit/rust-lightning/pull/1619#discussion_r927463658 of changing the name to `their_channel_reserve_proportional_millionths` was only partly addressed, as the name wasn't changed to "proportional" and kept as "proportion". I'm fine with both alternatives, although I think "proportional" makes more sense as we have another config flag with that wording (`forwarding_fee_proportional_millionths`), and "proportional" is therefore more coherent with the current field names.

Please update the commit message to use the one you choose to use however, as that's currently still using `their_channel_reserve_proportion_millionth` :).

lightning/src/ln/channel.rs
G8XSU: I am not sure if i understand this correctly, 
We shouldn't be testing the setters anyways(its trivial), we should test the functionality resulting from config.
I will clarify the comment.

tnull:
> Edit: Scratch my comment about the new name of the config field.
> 
> However, please update the commit message for the config field as that's currently still using `their_channel_reserve_proportion_millionth` and not the updated name.

Okay, now I'm confused by the discussion: as of 7c90ee3, it's still `their_channel_reserve_proportion_millionths`. Is this intentional or did it slip back in somehow? I still have a preference for `proportional`, but would also be fine if it was left as is to make some progress here. 

Other than that, LGTM.

G8XSU:
> > Edit: Scratch my comment about the new name of the config field.
> > However, please update the commit message for the config field as that's currently still using `their_channel_reserve_proportion_millionth` and not the updated name.
> 
> Okay, now I'm confused by the discussion: as of [7c90ee3](https://github.com/lightningdevkit/rust-lightning/commit/7c90ee3593afb87d1c3cd0be72c3986b45704fc5), it's still `their_channel_reserve_proportion_millionths`? Is this intentional or did it slip back in somehow? I still have a preference for `proportional`, but would also be fine if it was left as is to make some progress here.
> 
> Other than that, LGTM.

Fixed it

lightning/src/ln/channel.rs
ViktorTigerstrom: Thanks, your clarification addressed exactly what I found confusing :)!

ViktorTigerstrom (approved):
ACK 93ccd7f00746a0d66ab13e19d158aedad5a86f33

An absolute final nit would be to use \` chars around their_channel_reserve_proportional_millionths in the commit message header instead of ' chars, so that it's displayed as `their_channel_reserve_proportional_millionths` instead of 'their_channel_reserve_proportional_millionths'. 

Though that's absolutely not a requirement, so feel free to merge this without that change.

G8XSU:
> ACK [93ccd7f](https://github.com/lightningdevkit/rust-lightning/commit/93ccd7f00746a0d66ab13e19d158aedad5a86f33)
> 
> Absolute final nits: Use ` chars around their_channel_reserve_proportional_millionths in the commit message header instead of ' chars, so that it's displayed as `their_channel_reserve_proportional_millionths` instead of 'their_channel_reserve_proportional_millionths'.
> 
> The first sentence in the commit message body is also a bit hard to understand, and has been updated in the config docs since it was written. If you want to, update it something like the first sentence in the config flag docs i.e.: Configures the counterparty's channel reserve proportional to the channel value , i.e., `their_channel_reserve_satoshis` for both outbound and inbound channels.
> 
> Though these are absolutely not requirements, so feel free to merge this without changing this.

Git cli wasnt allowing for that char, was omitting the field altogether. 
Regarding commit description, its same as docs, i think its fine as it is.

ViktorTigerstrom:
> > ACK [93ccd7f](https://github.com/lightningdevkit/rust-lightning/commit/93ccd7f00746a0d66ab13e19d158aedad5a86f33)
> > Absolute final nits: Use `chars around their_channel_reserve_proportional_millionths in the commit message header instead of ' chars, so that it's displayed as`their_channel_reserve_proportional_millionths` instead of 'their_channel_reserve_proportional_millionths'.
> > The first sentence in the commit message body is also a bit hard to understand, and has been updated in the config docs since it was written. If you want to, update it something like the first sentence in the config flag docs i.e.: Configures the counterparty's channel reserve proportional to the channel value , i.e., `their_channel_reserve_satoshis` for both outbound and inbound channels.
> > Though these are absolutely not requirements, so feel free to merge this without changing this.
> 
> Git cli wasnt allowing for that char, was omitting the field altogether. Regarding commit description, its same as docs, i think its fine as it is.

Ok fair enough :)

tnull (dismissed):
LGTM.

lightning/src/ln/channel.rs
```diff
@@ -843,16 +846,25 @@ impl<Signer: Sign> Channel<Signer> {
...
-	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64) -> u64 {
+	/// This is used both for outbound and inbound channels and has lower bound
+	/// of `MIN_THEIR_CHAN_RESERVE_SATOSHIS`.
+	pub(crate) fn get_holder_selected_channel_reserve_satoshis(channel_value_satoshis: u64, config: &UserConfig) -> u64 {
+		let calculated_reserve = channel_value_satoshis.saturating_mul(config.channel_handshake_config.their_channel_reserve_proportional_millionths as u64) / 1_000_000;
+		cmp::min(channel_value_satoshis, cmp::max(calculated_reserve, MIN_THEIR_CHAN_RESERVE_SATOSHIS))
+	}
+
+	/// This is for legacy reasons, present for forward-compatibility.
+	/// LDK versions older than " < 0.0.104" don't know how read/handle values other than default
```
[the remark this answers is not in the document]
G8XSU: fixed it

lightning/src/ln/channel.rs
```diff
@@ -6962,6 +6979,64 @@ mod tests {
 		assert_eq!(chan_8.holder_max_htlc_value_in_flight_msat, chan_8_value_msat);
 	}
 
+    #[test]
+    fn test_configured_holder_selected_channel_reserve_satoshis() {
```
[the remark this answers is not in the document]
G8XSU: yea, probably because of playing with rustfmt :x

lightning/src/util/config.rs
```diff
@@ -126,6 +126,30 @@ pub struct ChannelHandshakeConfig {
...
+	/// Also, if our selected channel reserve for counterparty and counterparty's selected
+	/// channel reserve for us sum up to equal or greater than channel value, channel negotiations
+	/// will fail.
+	///
+	/// Note: Versions of LDK earlier than v0.0.104 will fail to read channels with any channel reserve
+	/// other than the default value.
+	///
+	/// Default value: 1% of channel value, i.e., configured as 10,000 millionths.
+	/// Minimum value: If the calculated proportional value is less than 1000 sats, it will be treated
+	///    			   as 1000 sats instead, which is a safe implementation-specific lower bound.
```
ViktorTigerstrom, line 149: I'm a bit confused, but I'm not sure when the mix between tabs and spaces after the /// got added again? I thought that they were addressed previously. Text alignment should be done with spaces only (and not tabs which is confusing as the rest of the code base uses tabs for indentation).

ViktorTigerstrom (dismissed):
ACK f509398723ad2eb326f1056cdf69994c5f7ef4f6

I've left a comment regarding incorrect text alignment, which I thought was fixed in previous commit. Though it's just a minor issue, so feel free to merge this without the fix if you want to land this.

lightning/src/util/config.rs
G8XSU: had fixed it, idk how it got added again, maybe due to my habit.

ViktorTigerstrom (approved):

tnull (approved):

PR #1643, opened 2022-08-01
Track in-flight HTLCs across payments when routing

jurvis:
Addresses #1267. 

## Purpose
As a sender, we want to keep track of all the HTLCs that are currently in flight. We currently do that between pathfinding attempts in `find_route`. However, that information is not currently shared across payments, leading to some inaccuracies.

## Proposed Implementation
Accounting for the HTLCs will consist of two efforts: keeping track of them, and using them. 

### Keep Track
In `pay_internal`, we want to respond to specific results of `send_payment` which will tell us when we'll need to store the path information for use later. 

Path information is then stored in `payment_cache` of our payer.

### Using
Before `find_route` is called, we retrieve all the paths we stored, and loop through them to construct a HashMap (`HashMap<(u64, bool), u64>`) mapping a channel's short channel id and direction with the amount being sent. 

This map is wrapped with our user-supplied score in a`AccountForInFlightHtlcs` struct, which will serve to delegate `Score` functions to its underlying scorer. It will also use our new liquidity map to construct a `ChannelUsage` accounting for the inflight HTLCs to be passed to `channel_penalty_msat`.

Last Edited: 2022-08-15T05:27:11.602Z

## Other Notes
Future opportunities:
- @ariard: track CLTV deltas to know when HTLCs are expiring/stuck
- @jkczyz: have router not need to depend on scorer (possibly in a future commit of this PR)

### DIFF
--- a/lightning-invoice/src/payment.rs
+++ b/lightning-invoice/src/payment.rs
@@ -77,8 +77,8 @@
 //! # }
 //! #
 //! # struct FakeRouter {}
-//! # impl<S: Score> Router<S> for FakeRouter {
-//! #     fn find_route(
+//! # impl Router for FakeRouter {
+//! #     fn find_route<S: Score>(
 //! #         &self, payer: &PublicKey, params: &RouteParameters, payment_hash: &PaymentHash,
 //! #         first_hops: Option<&[&ChannelDetails]>, scorer: &S
 //! #     ) -> Result<Route, LightningError> { unimplemented!() }
@@ -144,8 +144,10 @@ use crate::prelude::*;
 use lightning::ln::{PaymentHash, PaymentPreimage, PaymentSecret};
 use lightning::ln::channelmanager::{ChannelDetails, PaymentId, PaymentSendFailure};
 use lightning::ln::msgs::LightningError;
-use lightning::routing::scoring::{LockableScore, Score};
+use lightning::routing::gossip::NodeId;
+use lightning::routing::scoring::{ChannelUsage, LockableScore, Score};
 use lightning::routing::router::{PaymentParameters, Route, RouteHop, RouteParameters};
+use lightning::util::errors::APIError;
 use lightning::util::events::{Event, EventHandler};
 use lightning::util::logger::Logger;
 use time_utils::Time;
@@ -175,10 +177,9 @@ use time_utils;
 type ConfiguredTime = time_utils::Eternity;
 
 /// (C-not exported) generally all users should use the [`InvoicePayer`] type alias.
-pub struct InvoicePayerUsingTime<P: Deref, R, S: Deref, L: Deref, E: EventHandler, T: Time>
+pub struct InvoicePayerUsingTime<P: Deref, R: Router, S: Deref, L: Deref, E: EventHandler, T: Time>
 where
 	P::Target: Payer,
-	R: for <'a> Router<<<S as Deref>::Target as LockableScore<'a>>::Locked>,
 	S::Target: for <'a> LockableScore<'a>,
 	L::Target: Logger,
 {
@@ -207,6 +208,50 @@ impl<T: Time> PaymentInfo<T> {
 		}
 	}
 }
+
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+struct AccountForInFlightHtlcs<'a, S: Score> {
+	scorer: &'a mut S,
+	/// Maps a channel's short channel id and its direction to the liquidity used up.
+	inflight_htlcs: HashMap<(u64, bool), u64>,
+}
+
+#[cfg(c_bindings)]
+impl<'a, S:Score> lightning::util::ser::Writeable for AccountForInFlightHtlcs<'a, S> {
+	fn write<W: lightning::util::ser::Writer>(&self, writer: &mut W) -> Result<(), std::io::Error> { self.scorer.write(writer) }
+}
+
+impl<'a, S: Score> Score for AccountForInFlightHtlcs<'a, S> {
+	fn channel_penalty_msat(&self, short_channel_id: u64, source: &NodeId, target: &NodeId, usage: ChannelUsage) -> u64 {
+		if let Some(used_liqudity) = self.inflight_htlcs.get(&(short_channel_id, source < target)) {
+			let usage = ChannelUsage {
+				inflight_htlc_msat: usage.inflight_htlc_msat + used_liqudity,
+				..usage
+			};
+
+			self.scorer.channel_penalty_msat(short_channel_id, source, target, usage)
+		} else {
+			self.scorer.channel_penalty_msat(short_channel_id, source, target, usage)
+		}
+	}
+
+	fn payment_path_failed(&mut self, path: &[&RouteHop], short_channel_id: u64) {
+		self.scorer.payment_path_failed(path, short_channel_id)
+	}
+
+	fn payment_path_successful(&mut self, path: &[&RouteHop]) {
+		self.scorer.payment_path_successful(path)
+	}
+
+	fn probe_failed(&mut self, path: &[&RouteHop], short_channel_id: u64) {
+		self.scorer.probe_failed(path, short_channel_id)
+	}
+
+	fn probe_successful(&mut self, path: &[&RouteHop]) {
+		self.scorer.probe_successful(path)
+	}
+}
+
 /// Storing minimal payment attempts information required for determining if a outbound payment can
 /// be retried.
 #[derive(Clone, Copy)]
@@ -267,9 +312,9 @@ pub trait Payer {
 }
 
 /// A trait defining behavior for routing an [`Invoice`] payment.
-pub trait Router<S: Score> {
+pub trait Router {
 	/// Finds a [`Route`] between `payer` and `payee` for a payment with the given values.
-	fn find_route(
+	fn find_route<S: Score>(
 		&self, payer: &PublicKey, route_params: &RouteParameters, payment_hash: &PaymentHash,
 		first_hops: Option<&[&ChannelDetails]>, scorer: &S
 	) -> Result<Route, LightningError>;
@@ -314,10 +359,9 @@ pub enum PaymentError {
 	Sending(PaymentSendFailure),
 }
 
-impl<P: Deref, R, S: Deref, L: Deref, E: EventHandler, T: Time> InvoicePayerUsingTime<P, R, S, L, E, T>
+impl<P: Deref, R: Router, S: Deref, L: Deref, E: EventHandler, T: Time> InvoicePayerUsingTime<P, R, S, L, E, T>
 where
 	P::Target: Payer,
-	R: for <'a> Router<<<S as Deref>::Target as LockableScore<'a>>::Locked>,
 	S::Target: for <'a> LockableScore<'a>,
 	L::Target: Logger,
 {
@@ -440,13 +484,19 @@ where
 
 		let payer = self.payer.node_id();
 		let first_hops = self.payer.first_hops();
+		let inflight_htlcs = self.create_inflight_map();
 		let route = self.router.find_route(
-			&payer, params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
-			&self.scorer.lock()
+			&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
+			&AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs }
 		).map_err(|e| PaymentError::Routing(e))?;
 
 		match send_payment(&route) {
-			Ok(payment_id) => Ok(payment_id),
+			Ok(payment_id) => {
+				for path in route.paths {
+					self.process_path_inflight_htlcs(payment_hash, path);
+				}
+				Ok(payment_id)
+			},
 			Err(e) => match e {
 				PaymentSendFailure::ParameterError(_) => Err(e),
 				PaymentSendFailure::PathParameterError(_) => Err(e),
@@ -461,7 +511,19 @@ where
 						Err(e)
 					}
 				},
-				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, .. } => {
+				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, results } => {
+					// If a `PartialFailure` event returns a result that is an `Ok()`, it means that
+					// part of our payment is retried. When we receive `MonitorUpdateFailed`, it
+					// means that we are still waiting for our channel monitor update to be completed.
+					for (result, path) in results.iter().zip(route.paths.into_iter()) {
+						match result {
+							Ok(_) | Err(APIError::MonitorUpdateFailed) => {
+								self.process_path_inflight_htlcs(payment_hash, path);
+							},
+							_ => {},
+						}
+					}
+
 					if let Some(retry_data) = failed_paths_retry {
 						// Some paths were sent, even if we failed to send the full MPP value our
 						// recipient may misbehave and claim the funds, at which point we have to
@@ -479,6 +541,24 @@ where
 				},
 			},
 		}.map_err(|e| PaymentError::Sending(e))
+	}
+
+	// Takes in a path to have its information stored in `payment_cache`. This is done for paths
+	// that are pending retry.
+	fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, path: Vec<RouteHop>) {
+		self.payment_cache.lock().unwrap().entry(payment_hash)
+			.or_insert_with(|| PaymentInfo::new())
+			.paths.push(path);
+	}
+
+	// Find the path we want to remove in `payment_cache`. If it doesn't exist, do nothing.
+	fn remove_path_inflight_htlcs(&self, payment_hash: PaymentHash, path: &Vec<RouteHop>) {
+		self.payment_cache.lock().unwrap().entry(payment_hash)
+			.and_modify(|payment_info| {
+				if let Some(idx) = payment_info.paths.iter().position(|p| p == path) {
+					payment_info.paths.swap_remove(idx);
+				}
+			});
 	}
 
 	fn retry_payment(
@@ -508,17 +588,25 @@ where
 
 		let payer = self.payer.node_id();
 		let first_hops = self.payer.first_hops();
+		let inflight_htlcs = self.create_inflight_map();
+
 		let route = self.router.find_route(
 			&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
-			&self.scorer.lock()
+			&AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs }
 		);
+
 		if route.is_err() {
 			log_trace!(self.logger, "Failed to find a route for payment {}; not retrying ({:})", log_bytes!(payment_hash.0), attempts);
 			return Err(());
 		}
 
-		match self.payer.retry_payment(&route.unwrap(), payment_id) {
-			Ok(()) => Ok(()),
+		match self.payer.retry_payment(&route.as_ref().unwrap(), payment_id) {
+			Ok(()) => {
+				for path in route.unwrap().paths.into_iter() {
+					self.process_path_inflight_htlcs(payment_hash, path);
+				}
+				Ok(())
+			},
 			Err(PaymentSendFailure::ParameterError(_)) |
 			Err(PaymentSendFailure::PathParameterError(_)) => {
 				log_trace!(self.logger, "Failed to retry for payment {} due to bogus route/payment data, not retrying.", log_bytes!(payment_hash.0));
@@ -527,7 +615,19 @@ where
 			Err(PaymentSendFailure::AllFailedRetrySafe(_)) => {
 				self.retry_payment(payment_id, payment_hash, params)
 			},
-			Err(PaymentSendFailure::PartialFailure { failed_paths_retry, .. }) => {
+			Err(PaymentSendFailure::PartialFailure { failed_paths_retry, results, .. }) => {
+				// If a `PartialFailure` error contains a result that is an `Ok()`, it means that
+				// part of our payment is retried. When we receive `MonitorUpdateFailed`, it
+				// means that we are still waiting for our channel monitor update to complete.
+				for (result, path) in results.iter().zip(route.unwrap().paths.into_iter()) {
+					match result {
+						Ok(_) | Err(APIError::MonitorUpdateFailed) => {
+							self.process_path_inflight_htlcs(payment_hash, path);
+						},
+						_ => {},
+					}
+				}
+
 				if let Some(retry) = failed_paths_retry {
 					// Always return Ok for the same reason as noted in pay_internal.
 					let _ = self.retry_payment(payment_id, payment_hash, &retry);
@@ -544,6 +644,47 @@ where
 	pub fn remove_cached_payment(&self, payment_hash: &PaymentHash) {
 		self.payment_cache.lock().unwrap().remove(payment_hash);
 	}
+
+	/// Given a [`PaymentHash`], this function looks up inflight path attempts in the payment_cache.
+	/// Then, it uses the path information inside the cache to construct a HashMap mapping a channel's
+	/// short channel id and direction to the amount being sent through it.
+	///
+	/// This function should be called whenever we need information about currently used up liquidity
+	/// across payments.
+	fn create_inflight_map(&self) -> HashMap<(u64, bool), u64> {
+		let mut total_inflight_map: HashMap<(u64, bool), u64> = HashMap::new();
+		// Make an attempt at finding existing payment information from `payment_cache`. If it
+		// does not exist, it probably is a fresh payment and we can just return an empty
+		// HashMap.
+		for payment_info in self.payment_cache.lock().unwrap().values() {
+			for path in &payment_info.paths {
+				if path.is_empty() { break };
+				// total_inflight_map needs to be direction-sensitive when keeping track of the HTLC value
+				// that is held up. However, the `hops` array, which is a path returned by `find_route` in
+				// the router excludes the payer node. In the following lines, the payer's information is
+				// hardcoded with an inflight value of 0 so that we can correctly represent the first hop
+				// in our sliding window of two.
+				let our_node_id: PublicKey = self.payer.node_id();
+				let reversed_hops_with_payer = path.iter().rev().skip(1)
+					.map(|hop| hop.pubkey)
+					.chain(core::iter::once(our_node_id));
+				let mut cumulative_msat = 0;
+
+				// Taking the reversed vector from above, we zip it with just the reversed hops list to
+				// work "backwards" of the given path, since the last hop's `fee_msat` actually represents
+				// the total amount sent.
+				for (next_hop, prev_hop) in path.iter().rev().zip(reversed_hops_with_payer) {
+					cumulative_msat += next_hop.fee_msat;
+					total_inflight_map
+						.entry((next_hop.short_channel_id, NodeId::from_pubkey(&prev_hop) < NodeId::from_pubkey(&next_hop.pubkey)))
+						.and_modify(|used_liquidity_msat| *used_liquidity_msat += cumulative_msat)
+						.or_insert(cumulative_msat);
+				}
+			}
+		}
+
+		total_inflight_map
+	}
 }
 
 fn expiry_time_from_unix_epoch(invoice: &Invoice) -> Duration {
@@ -557,14 +698,23 @@ fn has_expired(route_params: &RouteParameters) -> bool {
 	} else { false }
 }
 
-impl<P: Deref, R, S: Deref, L: Deref, E: EventHandler, T: Time> EventHandler for InvoicePayerUsingTime<P, R, S, L, E, T>
+impl<P: Deref, R: Router, S: Deref, L: Deref, E: EventHandler, T: Time> EventHandler for InvoicePayerUsingTime<P, R, S, L, E, T>
 where
 	P::Target: Payer,
-	R: for <'a> Router<<<S as Deref>::Target as LockableScore<'a>>::Locked>,
 	S::Target: for <'a> LockableScore<'a>,
 	L::Target: Logger,
 {
 	fn handle_event(&self, event: &Event) {
+		match event {
+			Event::PaymentPathFailed { payment_hash, path, ..  }
+			| Event::PaymentPathSuccessful { path, payment_hash: Some(payment_hash), .. }
+			| Event::ProbeSuccessful { payment_hash, path, .. }
+			| Event::ProbeFailed { payment_hash, path, .. } => {
+				self.remove_path_inflight_htlcs(*payment_hash, path);
+			},
+			_ => {},
+		}
+
 		match event {
 			Event::PaymentPathFailed {
 				payment_id, payment_hash, rejected_by_dest, path, short_channel_id, retry, ..
@@ -633,7 +783,7 @@ mod tests {
 	use lightning::ln::features::{ChannelFeatures, NodeFeatures, InitFeatures};
 	use lightning::ln::functional_test_utils::*;
 	use lightning::ln::msgs::{ChannelMessageHandler, ErrorAction, LightningError};
-	use lightning::routing::gossip::NodeId;
+	use lightning::routing::gossip::{EffectiveCapacity, NodeId};
 	use lightning::routing::router::{PaymentParameters, Route, RouteHop};
 	use lightning::routing::scoring::ChannelUsage;
 	use lightning::util::test_utils::TestLogger;
@@ -645,6 +795,7 @@ mod tests {
 	use std::time::{SystemTime, Duration};
 	use time_utils::tests::SinceEpoch;
 	use DEFAULT_EXPIRY_TIME;
+	use lightning::util::errors::APIError::{ChannelUnavailable, MonitorUpdateFailed};
 
 	fn invoice(payment_preimage: PaymentPreimage) -> Invoice {
 		let payment_hash = Sha256::hash(&payment_preimage.0);
@@ -792,8 +943,8 @@ mod tests {
 		let final_value_msat = invoice.amount_milli_satoshis().unwrap();
 
 		let payer = TestPayer::new()
-			.fails_with_partial_failure(retry.clone(), OnAttempt(1))
-			.fails_with_partial_failure(retry, OnAttempt(2))
+			.fails_with_partial_failure(retry.clone(), OnAttempt(1), None)
+			.fails_with_partial_failure(retry, OnAttempt(2), None)
 			.expect_send(Amount::ForInvoice(final_value_msat))
 			.expect_send(Amount::OnRetry(final_value_msat / 2))
 			.expect_send(Amount::OnRetry(final_value_msat / 2));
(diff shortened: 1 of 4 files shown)


lightning-invoice/src/payment.rs
```diff
@@ -466,6 +494,46 @@ where
...
+		let reversed_hops_with_payer = core::iter::once((0u64, our_node_id)).chain(
+			hops.split_last().unwrap().1.iter().map(|hop| (hop.fee_msat, hop.pubkey))).rev();
+		let mut cumulative_msat = hops.last().unwrap().fee_msat;
+
+		// Taking the reversed vector from above, we zip it with just the reversed hops list to
+		// work "backwards" of the given path, since the last hop's `fee_msat` actually represents
+		// the total amount sent.
+		for (_, (next_hop, prev_hop)) in hops.iter().rev().zip(reversed_hops_with_payer).enumerate() {
+			cumulative_msat += prev_hop.0;
+			self.total_inflight_map.lock().unwrap()
```
jkczyz, line 513: We'll likely have to do something similar to scoring where we have a `LockableInflightHtlcs` trait and some `InflightHtlcs` struct (though no need for an associated type and trait). That way, when passing this to `find_route` we'll pass the locked version such that it sees a consistent view of the in-flight HTLCs.

lightning-invoice/src/payment.rs
```diff
@@ -466,6 +494,46 @@ where
...
+		// in our sliding window of two.
+		let our_node_id: PublicKey = self.payer.node_id();
+		let reversed_hops_with_payer = core::iter::once((0u64, our_node_id)).chain(
+			hops.split_last().unwrap().1.iter().map(|hop| (hop.fee_msat, hop.pubkey))).rev();
+		let mut cumulative_msat = hops.last().unwrap().fee_msat;
+
+		// Taking the reversed vector from above, we zip it with just the reversed hops list to
+		// work "backwards" of the given path, since the last hop's `fee_msat` actually represents
+		// the total amount sent.
+		for (_, (next_hop, prev_hop)) in hops.iter().rev().zip(reversed_hops_with_payer).enumerate() {
```
jkczyz, line 511: No need for `enumerate` here.

lightning-invoice/src/payment.rs
```diff
@@ -715,11 +787,31 @@ mod tests {
 		let payment_id = Some(invoice_payer.pay_invoice(&invoice).unwrap());
 		assert_eq!(*payer.attempts.borrow(), 1);
 
+		let inflight_map = invoice_payer.total_inflight_map.lock().unwrap();
+		// TestRouter returns two hops, with `fee_msat` divided by two. The invoice instantiated above
+		// is hard-coded to test sending an invoice with a value of 128 msats.
+		assert_eq!(inflight_map.get(&(0, true)).unwrap().clone(), 64);
+		assert_eq!(inflight_map.get(&(1, true)).unwrap().clone(), 64);
+		drop(inflight_map);
```
jkczyz, line 795: Surround this with a scope to avoid the explicit `drop`.

jkczyz:
@TheBlueMatt Would good to have some early feedback on this PR. Upcoming commits will use in-flight HTLC tracking during routing. Follow-up PR will deal with persistence.

dunxen:
Going to be using this for review club session 5. Nice to dig into payment logic a bit. :)

jurvis:
@TheBlueMatt 

> As for the followup to do the tracking, I don't think we need a new interface in the router at all, I think we can do it entirely by wrapping the scorer and changing the amounts we pass through to the user-provided scorer.

hm, I think we do need to pass in the inflight htlcs map into `get_route` because we do account for `used_channel_liquidities` when pathfinding [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/router.rs#L1561-L1575) as well.

lightning-invoice/src/payment.rs
```diff
@@ -562,22 +630,26 @@ where
...
 			},
 			Event::PaymentFailed { payment_hash, .. } => {
 				self.remove_cached_payment(&payment_hash);
 			},
 			Event::PaymentPathSuccessful { path, .. } => {
-				let path = path.iter().collect::<Vec<_>>();
-				self.scorer.lock().payment_path_successful(&path);
+				let path_with_refs = path.iter().collect::<Vec<_>>();
+				self.scorer.lock().payment_path_successful(&path_with_refs);
+				self.remove_path_inflight_htlcs(path);
```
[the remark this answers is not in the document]
jurvis: got it. if we track amount and path info in `payment_cache,` is there a way to key the path information with `PaymentHash` uniquely? or can we guarantee that it will not try the same path again?

I'm thinking specifically of what happens when we re-attempt to send an HTLC through a path that was attempted before and still outstanding. 

I think we decided to track inflight htlcs in a separate map so that they are easier to be looked up by scid across all payments, rather than local to them, which is similar to what we do [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/router.rs#L1563-L1566) in the router.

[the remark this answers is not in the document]
jkczyz: > > I think we decided to track inflight htlcs in a separate map so that they are easier to be looked up by scid across all payments, rather than local to them, which is similar to what we do [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/routing/router.rs#L1563-L1566) in the router.
> 
> Yea, both approaches are valid, I just worry about cases like the one you describe where we could accidentally have too many paths being tracked in the map and we end up forever thinking there's paths pending. I don't have specific examples, however, so I could just be wrong.

In worse case, I think we'd subtract too much and so would need a `saturating_sub`. i.e., The map always starts with zero, but there can be in-flight HTLCs from prior to a restart. And any HTLCs sent after a restart would have been added already.

I think the larger point, though, is when calling `channel_penalty_msat` we'd prefer to do one map lookup. Otherwise, we'd have to scan all the paths stored in every entry of `payment_cache` in order to sum up the in-flight HTLCs (i.e., a channel can have an in-flight HTLC for more than one payment, IIUC).

jkczyz:
> Yea, if we want to exactly match the behavior, but I'm not sure that is neccessary - it'll result in a few suboptimal decisions, but we can do it only in the scorer and we'll still get most of the way there, without more complexity in the router itself.

There may be another case when computing `available_value_contribution_msat`:

https://github.com/lightningdevkit/rust-lightning/blob/d02425121ec082689a18bb0b22bec226cc02ff7d/lightning/src/routing/router.rs#L1002-L1004

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jkczyz: Ah, that makes sense. So basically, store paths in `payment_cache`. Then when needing to call `find_route`, produce the map from `payment_cache`, lock the scorer, wrap it and the map in an internally defined scorer, and pass that to `find_route`.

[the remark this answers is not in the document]
jurvis: thanks for chiming in @jkczyz @TheBlueMatt -- this does make more sense now.

just to make sure I understand correctly: it sounds like we are okay trading off the efficiency of a single map lookup so we do not have to keep the data within `inflight_htlcs_map` in sync with `payment_cache`? When routing, we'll just simply compute the same contract (`HashMap<(u64, bool), u64)>`) from `payment_cache` to `find_route`?

jurvis:
> I do kinda now wonder if we shouldn't have made that a method on the scorer

which method? `channel_penalty_msat`? is your suggestion to pass our map of inflight htlcs to the scorer and just have it subtracted when calculating `capacity_msat`?

https://github.com/lightningdevkit/rust-lightning/blob/d02425121ec082689a18bb0b22bec226cc02ff7d/lightning/src/routing/scoring.rs#L807-L808

jkczyz:
> > I do kinda now wonder if we shouldn't have made that a method on the scorer
> 
> which method? `channel_penalty_msat`? is your suggestion to pass our map of inflight htlcs to the scorer and just have it subtracted when calculating `capacity_msat`?
> 
> https://github.com/lightningdevkit/rust-lightning/blob/d02425121ec082689a18bb0b22bec226cc02ff7d/lightning/src/routing/scoring.rs#L807-L808

@TheBlueMatt Did you mean *should* have not *shouldn't* have?

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jkczyz: Yeah, sorta. We still get the benefit of a single map look-up in `find_route` each time it calls `channel_penalty_msat`. It's just that we compute that map (i.e., `inflight_htlcs_map` once prior to each call to `find_route`. The map itself will actually be pretty small. (i.e., it will contain less channels than channels considered by `find_route`).

Will look something like this:

```rust
struct AccountForInFlightHtlcs<'a, S: Score> {
    scorer: &'a S,
    inflight_htlcs: HashMap<(u64, bool), u64)>,
}

impl<'_, S: Score> Score for AccountForInFlightHtlcs<S, '_> {
    // ...
}

let inflight_htlcs = ...;
let locked_scorer = self.scorer.lock();
let scorer = AccountForInFlightHtlcs { scorer: &*locked_scorer, inflight_htlcs };
let route = self.router.find_route(... , &scorer);
```

jkczyz:
Another thing to keep in mind is that we'd eventually like to move the `LockableScore` from `InvoicePayer` to `DefaultRouter` since a remote `Router` wouldn't need a scorer locally. So we may need to pass the in-flight HTLC map to `Router::find_route` even if it is not used by the standalone `find_route` function that `DefaultRoute::find_route` calls. But I think the change would be fairly trivial (i.e., moving the use of the `AccountForInFlightHtlcs` wrapper into `DefaultRoute::find_route`).

lightning-invoice/src/payment.rs
```diff
@@ -562,22 +630,26 @@ where
 				} else if *rejected_by_dest {
 					log_trace!(self.logger, "Payment {} rejected by destination; not retrying", log_bytes!(payment_hash.0));
 					self.payer.abandon_payment(payment_id.unwrap());
+					self.remove_path_inflight_htlcs(path);
```
ariard, line 633: What's the reasoning behind not adding `remove_path_inflight_htlcs` in the branch above (`payment_id.is_none()`) ?

lightning-invoice/src/payment.rs
```diff
@@ -466,6 +494,46 @@ where
 		}.map_err(|e| PaymentError::Sending(e))
 	}
 
+	fn process_path_inflight_htlcs(&self, hops: &Vec<RouteHop>) {
```
ariard, line 497: I think the PR would benefit from a better description of what exactly is aimed. AFAICT, it introduces tracking of in-flight HTLCs issued by the sender to update in consequence scoring's `DirectedChannelLiquidity` (to be introduced in future commit IIUC the direction) and generate more accurate scores ?

One interesting follow-up could be to track the CLTV delta as well, to know either when all the HTLC are going to expire and the link is going to be ready again or there is a HTLC stuck and the link should be disregarded (okay we might have few more rounds of iterations before we do that kind of scoring optimizations).

lightning-invoice/src/payment.rs
jurvis: no reason, I missed it. good catch 😅

lightning-invoice/src/payment.rs
```diff
@@ -763,6 +907,15 @@ mod tests {
 		});
 		assert_eq!(*event_handled.borrow(), true);
 		assert_eq!(*payer.attempts.borrow(), 2);
+
+		invoice_payer.handle_event(&Event::PaymentPathSuccessful {
```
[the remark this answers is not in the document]
jurvis: ah, I removed the assertions from the previous implementation I forgot to add the them back in

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +190,81 @@ where
...
+	fn new() -> Self {
+		PaymentInfo {
+			attempts: PaymentAttempts::new(),
+			paths: vec![vec![]],
+		}
+	}
+}
+
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+struct AccountForInFlightHtlcs<'a, S: Score> {
```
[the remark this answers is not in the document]
jurvis: yeah, I'm a little stuck trying to untangle myself with passing the wrapped scorer to `router.find_route`

```rust
let inflight_htlcs = {...};
let scorer = AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs };

let route = self.router.find_route(..., &scorer);
```

which causes an error:

```
note: expected reference `&<<S as Deref>::Target as LockableScore<'_>>::Locked`
               found reference `&AccountForInFlightHtlcs<'_, <<S as Deref>::Target as LockableScore<'_>>::Locked>`
```

The error is expected since our impl currently expects our scorer to be `Deref` and likewise, our router is expecting its scorer to be a `LockableScore`. And that's pretty much where I'm stuck because I'm rather unfamiliar with the type assertions. I tried removing the constraints for `R` and `S`, and just have `S: Score` instead of `S: Deref` but I ended up opening a barrage of errors in tests, so it looked like I was doing something wrong.

Thoughts?

lightning-invoice/src/payment.rs
jurvis: @ariard I updated the PR description to reflect the current discussion of what we are doing :)

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jkczyz: Hmm... the problem is that `Router` is expecting the same `Score` implementation as `InvoicePayer`, but we are wrapping it with `AccountForInFlightHtlcs`. Moving the `Score` type parameterization from `Router` to `Router::find_route` and changing the `Router` type parameterization on `InvoicePayer` to simplly `R: Router` will fix this. This is functionally equivalent to moving the `Score` parameterization to `DefaultRouter` as per https://github.com/lightningdevkit/rust-lightning/pull/1643#issuecomment-1209810558.

Will also need to change the above lines to:

```rust
let mut locked_scorer = self.scorer.lock();
let scorer = AccountForInFlightHtlcs { scorer: &mut locked_scorer, inflight_htlcs };
```

lightning-invoice/src/payment.rs
```diff
@@ -466,6 +494,46 @@ where
...
+	fn remove_path_inflight_htlcs(&self, hops: &Vec<RouteHop>) {
+		// Here, just like `process_path_inflight_htlcs`, we are given a vector of hops that excludes
+		// any information about the payer. However, in our `total_inflight_map`, we keep track
+		// of possible inbound/outbound liquidity from the payer as well. Therefore, it is accounted
+		// for by hard-coding the same tuple as in `process_path_inflight_htlcs`
+		let our_node_id: PublicKey = self.payer.node_id();
+		let hops_with_us = core::iter::once((0u64, our_node_id)).chain(hops.iter().map(|hop| (hop.fee_msat, hop.pubkey)));
+		
+		for (next_hop, prev_hop) in hops.iter().zip(hops_with_us) {
+			println!("Subtracting {} inflight msats from {}", next_hop.fee_msat, next_hop.short_channel_id);
```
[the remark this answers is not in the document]
ariard: Sure, though sometimes you might be willing to observe flow-wise, not subsystem.

lightning-invoice/src/payment.rs
ariard: Thanks for that!

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +189,81 @@ where
...
+	/// For the last hop, this should be the full value of the payment (might be more than
+	/// requested if we had to match htlc_minimum_msat).
+	pub fee_msat: u64,
+}
+
+impl<T: Time> PaymentInfo<T> {
+	fn new() -> Self {
+		PaymentInfo {
+			attempts: PaymentAttempts::new(),
+			paths: vec![vec![]],
```
jkczyz, line 223: This creates a vector with an empty path.

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +189,81 @@ where
...
+}
+
+impl<'a, S: Score> Score for AccountForInFlightHtlcs<'a, S> {
+	fn channel_penalty_msat(&self, short_channel_id: u64, source: &NodeId, target: &NodeId, usage: ChannelUsage) -> u64 {
+		if let Some(used_liqudity) = self.inflight_htlcs.get(&(short_channel_id, target < source)) {
+			let usage = ChannelUsage {
+				amount_msat: usage.amount_msat,
+				inflight_htlc_msat: usage.inflight_htlc_msat + used_liqudity,
+				effective_capacity: usage.effective_capacity,
+			};
```
jkczyz, line 242: You can use the following shorthand for updating one field:

```rust
let usage = ChannelUsage {
    inflight_htlc_msat: usage.inflight_htlc_msat + used_liqudity,
    ..usage
};
```

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +189,81 @@ where
...
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+struct AccountForInFlightHtlcs<'a, S: Score> {
+	scorer: &'a mut S,
+	/// Maps a channel's short channel id and its direction to the liquidity used up.
+	inflight_htlcs: HashMap<(u64, bool), u64>,
+}
+
+impl<'a, S: Score> Score for AccountForInFlightHtlcs<'a, S> {
+	fn channel_penalty_msat(&self, short_channel_id: u64, source: &NodeId, target: &NodeId, usage: ChannelUsage) -> u64 {
+		if let Some(used_liqudity) = self.inflight_htlcs.get(&(short_channel_id, target < source)) {
```
jkczyz, line 237: Could you flip these around to `source < target` for consistency with the router?

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +496,94 @@ where
...
-		).map_err(|e| PaymentError::Routing(e))?;
+		let inflight_htlcs = {
+			// Make an attempt at finding existing payment information from `payment_cache`. If it
+			// does not exist, it probably is a fresh payment and we can just return an empty
+			// HashMap.
+			if let Some(payment_info) = self.payment_cache.lock().unwrap().get(&payment_hash) {
+				let mut total_inflight_map: HashMap<(u64, bool), u64> = HashMap::new();
+
+				for hops in &payment_info.paths {
+					if hops.len() < 1 { break };
```
jkczyz, line 507:
```rust
if hops.is_empty() { break; }
```

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +496,94 @@ where
...
-			&self.scorer.lock()
-		).map_err(|e| PaymentError::Routing(e))?;
+		let inflight_htlcs = {
+			// Make an attempt at finding existing payment information from `payment_cache`. If it
+			// does not exist, it probably is a fresh payment and we can just return an empty
+			// HashMap.
+			if let Some(payment_info) = self.payment_cache.lock().unwrap().get(&payment_hash) {
+				let mut total_inflight_map: HashMap<(u64, bool), u64> = HashMap::new();
+
+				for hops in &payment_info.paths {
```
jkczyz, line 506: May be more intuitive if you used the singular `path` instead of the plural `hops` or an element in the list.

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +496,94 @@ where
...
+				HashMap::new()
+			}
+		};
+
+		let route = {
+			let scorer = AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs };
+
+			self.router.find_route(
+				&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
+				&scorer
```
jkczyz, line 541: FWIW, instead of having a `scorer` variable, you can inline the creation of `AccountForInFlightHtlcs` to avoid needing the scope.

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +496,94 @@ where
...
 						core::mem::drop(payment_cache);
 						Ok(self.pay_internal(params, payment_hash, send_payment)?)
 					} else {
 						Err(e)
 					}
 				},
-				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, .. } => {
+				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, results } => {
+					// The ones that are `Ok()`s + `MonitorUpdateFailed` in `results` need to be used to update the map
+					// get the index of them and run to `route` to get the values we need to care about.
```
jkczyz, line 568: Can't quite parse this. Could you re-phrase?

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +496,94 @@ where
...
+					for (idx, result) in results.iter().enumerate() {
+						match result {
+							Ok(_) => {
+								if let Some(failed_path) = route.paths.get(idx) {
+									self.process_path_inflight_htlcs(payment_hash, failed_path.to_vec());
```
jkczyz, line 573: Why copy into a new `Vec` here? Have `process_path_inflight_htlcs` take a refernece.

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +496,94 @@ where
...
 					}
 				},
-				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, .. } => {
+				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, results } => {
+					// The ones that are `Ok()`s + `MonitorUpdateFailed` in `results` need to be used to update the map
+					// get the index of them and run to `route` to get the values we need to care about.
+					for (idx, result) in results.iter().enumerate() {
+						match result {
+							Ok(_) => {
+								if let Some(failed_path) = route.paths.get(idx) {
```
jkczyz, line 572: I think we can assume the lengths of `route.paths` and `results` are the same and simply index into the former.

lightning-invoice/src/payment.rs
```diff
@@ -466,16 +603,46 @@ where
 		}.map_err(|e| PaymentError::Sending(e))
 	}
 
+	// Takes in a path to have its information stored in `payment_cache`. This is done for paths
+	// that are pending retry.
+	fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, hops: Vec<RouteHop>) {
+		let payment_info_route_hops: Vec<PaymentInfoRouteHop> = hops.iter().map(|h| PaymentInfoRouteHop { pubkey: h.pubkey, short_channel_id: h.short_channel_id, fee_msat: h.fee_msat }).collect();
```
jkczyz, line 609: Could you wrap at 100 characters?

lightning-invoice/src/payment.rs
```diff
@@ -466,16 +603,46 @@ where
...
+
+		self.payment_cache.lock().unwrap().entry(payment_hash)
+			.or_insert(PaymentInfo {
+				attempts: PaymentAttempts {
+					count: 1,
+					first_attempted_at: T::now()
+				},
+				paths: vec![]
+			})
+			.paths.push(payment_info_route_hops);
```
jkczyz, line 619: Why not use `PaymentInfo`? Won't the `count` field be inaccurate when this is called with `Ok` results? I think we want it to be zero in all cases actually and let `retry_payment` increment the count.

lightning-invoice/src/payment.rs
```diff
@@ -559,31 +769,38 @@ where
...
 					log_trace!(self.logger, "Payment {} rejected by destination; not retrying", log_bytes!(payment_hash.0));
 					self.payer.abandon_payment(payment_id.unwrap());
+					self.remove_path_inflight_htlcs(*payment_hash, path);
 				} else if retry.is_none() {
 					log_trace!(self.logger, "Payment {} missing retry params; not retrying", log_bytes!(payment_hash.0));
 					self.payer.abandon_payment(payment_id.unwrap());
+					self.remove_path_inflight_htlcs(*payment_hash, path);
 				} else if self.retry_payment(payment_id.unwrap(), *payment_hash, retry.as_ref().unwrap()).is_ok() {
 					// We retried at least somewhat, don't provide the PaymentPathFailed event to the user.
 					return;
```
jkczyz, line 783: Don't we need to remove it here? The retry could be on an entirely different path or multiple paths.

lightning-invoice/src/payment.rs
jurvis: ah, yes, that makes sense.

lightning-invoice/src/payment.rs
jurvis: hm, should we not, in this case, rely on the payment retry to send us another event (either `PaymentPathFailed` or `PaymentPathSuccessful`) which we'll then use instead to know when to remove?

jkczyz: Yes, we would indeed. But this *is* the event handler, so it needs to adjust for the path that has just failed (i.e., the one it just retried). And, actually, it should really adjust it before calling `retry_payment` otherwise the HTLC that just failed would still be accounted for as in-flight. I imagine we can just call `remove_path_inflight_htlcs` in one place prior to this long `if-else-if` expression.

lightning-invoice/src/payment.rs
jurvis: ah, I must have confused myself when I computed backwards to populate the map. Thanks :P

lightning-invoice/src/payment.rs
```diff
@@ -533,19 +532,16 @@ where
...
-			let scorer = AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs };
-
-			self.router.find_route(
+		let route = self.router.find_route(
 				&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
-				&scorer
-			).map_err(|e| PaymentError::Routing(e))?
-		};
+				&AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs }
+			).map_err(|e| PaymentError::Routing(e))?;
```
jkczyz, line 538: Reduce indent by one.

lightning-invoice/src/payment.rs
```diff
@@ -533,19 +532,16 @@ where
...
+
 
 		match send_payment(&route) {
 			Ok(payment_id) => {
 				for hops in route.paths {
```
jkczyz, line 543: s/hops/path

lightning-invoice/src/payment.rs
```diff
@@ -605,23 +602,21 @@ where
...
 
 		self.payment_cache.lock().unwrap().entry(payment_hash)
-			.or_insert(PaymentInfo {
-				attempts: PaymentAttempts {
-					count: 1,
-					first_attempted_at: T::now()
-				},
-				paths: vec![]
-			})
+			.or_insert(PaymentInfo::new())
```
jkczyz, line 611: Here and below, use `or_insert_with(PaymentInfo::new)` so the struct is only created when needed.

lightning-invoice/src/payment.rs
```diff
@@ -491,15 +647,59 @@ where
...
+		};
+
+		let route = {
+			let scorer = AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs };
+
+			self.router.find_route(
+				&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
+				&scorer
+			)
+		};
```
jkczyz, line 694: Remove the scope here, too.

lightning-invoice/src/payment.rs
```diff
@@ -466,16 +600,38 @@ where
 		}.map_err(|e| PaymentError::Sending(e))
 	}
 
+	// Takes in a path to have its information stored in `payment_cache`. This is done for paths
+	// that are pending retry.
+	fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, hops: &Vec<RouteHop>) {
```
jkczyz, line 605: s/hops/path

lightning-invoice/src/payment.rs
```diff
@@ -466,16 +600,38 @@ where
...
+			.map(|h| PaymentInfoRouteHop { pubkey: h.pubkey, short_channel_id: h.short_channel_id, fee_msat: h.fee_msat })
+			.collect();
+
+		self.payment_cache.lock().unwrap().entry(payment_hash)
+			.or_insert(PaymentInfo::new())
+			.paths.push(payment_info_route_hops);
+	}
+
+	// Find the path we want to remove in `payment_cache`. If it doesn't exist, do nothing.
+	fn remove_path_inflight_htlcs(&self, payment_hash: PaymentHash, hops: &Vec<RouteHop>) {
```
jkczyz, line 616: s/hops/path

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +495,92 @@ where
...
+							.and_modify(|used_liquidity_msat| *used_liquidity_msat += cumulative_msat)
+							.or_insert(cumulative_msat);
+					}
+				}
+
+				total_inflight_map
+			} else {
+				HashMap::new()
+			}
+		};
```
jkczyz, line 533: Looks like this is repeated below. Could you make a helper method?

lightning-invoice/src/payment.rs
```diff
@@ -491,15 +647,59 @@ where
...
+			)
+		};
+
+
 		if route.is_err() {
 			log_trace!(self.logger, "Failed to find a route for payment {}; not retrying ({:})", log_bytes!(payment_hash.0), attempts);
 			return Err(());
 		}
 
+
```
jkczyz, line 702: Remove extra blank line.

lightning-invoice/src/payment.rs
jurvis: gotcha, thanks!

lightning-invoice/src/payment.rs
jurvis: Without putting them inside the same scope, I run into an issue with a test (`pays_invoice_on_partial_failure`) which will print:

`thread 'payment::tests::pays_invoice_on_partial_failure' panicked at 'already borrowed: BorrowMutError'`

is there a better way to do this?

jkczyz: Should work if you inline the scorer.

lightning-invoice/src/payment.rs
```diff
@@ -466,16 +564,44 @@ where
 		}.map_err(|e| PaymentError::Sending(e))
 	}
 
+	// Takes in a path to have its information stored in `payment_cache`. This is done for paths
+	// that are pending retry.
+	fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, path: &Vec<RouteHop>) {
+		let payment_info_route_hops: Vec<PaymentInfoRouteHop> = path.iter()
+			.map(|h| PaymentInfoRouteHop { pubkey: h.pubkey, short_channel_id: h.short_channel_id, fee_msat: h.fee_msat })
```
jkczyz, line 571: Wrap these at 100 characters, too, by moving parameters down to their own line.

lightning-invoice/src/payment.rs
```diff
@@ -557,6 +732,7 @@ where
 					self.scorer.lock().payment_path_failed(&path, *short_channel_id);
 				}
 
+				self.remove_path_inflight_htlcs(*payment_hash, path);
```
jkczyz, line 735: We'll need to do this for the probing events, too, even though `InvoicePayer` hasn't exposed an interface for it yet. Let's pull these out into a separate `match` expression before this one to keep the code a bit more streamlined.

Here's a little rust pattern matching magic:

```rust
               match event {
                       Event::PaymentPathFailed { payment_hash, path, ..  }
                       | Event::PaymentPathSuccessful { path, payment_hash: Some(payment_hash), .. }
                       | Event::ProbeSuccessful { payment_hash, path, .. }
                       | Event::ProbeFailed { payment_hash, path, .. } => {
                               self.remove_path_inflight_htlcs(*payment_hash, path);
                       },
                       _ => {},
               }
```

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +189,80 @@ where
...
+	scorer: &'a mut S,
+	/// Maps a channel's short channel id and its direction to the liquidity used up.
+	inflight_htlcs: HashMap<(u64, bool), u64>,
+}
+
+impl<'a, S: Score> Score for AccountForInFlightHtlcs<'a, S> {
+	fn channel_penalty_msat(&self, short_channel_id: u64, source: &NodeId, target: &NodeId, usage: ChannelUsage) -> u64 {
+		if let Some(used_liqudity) = self.inflight_htlcs.get(&(short_channel_id, source > target)) {
+			let usage = ChannelUsage {
+				inflight_htlc_msat: usage.inflight_htlc_msat + used_liqudity,
```
jkczyz, line 239: Could you add test coverage for this? No test fails if I comment out this line. You'll likely need to do something fancy like make `TestRouter::find_route` call the the scorer and check some expectations that you set on `TestRouter` prior to using it.

lightning-invoice/src/payment.rs
```diff
@@ -715,11 +894,23 @@ mod tests {
...
+
 		invoice_payer.handle_event(&Event::PaymentSent {
 			payment_id, payment_preimage, payment_hash, fee_paid_msat: None
 		});
 		assert_eq!(*event_handled.borrow(), true);
 		assert_eq!(*payer.attempts.borrow(), 1);
+
+		// We should have cleared the payment_cache of information about this payment
+		let inflight_map = invoice_payer.create_inflight_map(&payment_hash);
+		assert_eq!(inflight_map.len(), 0)
```
jkczyz, line 913: You'll need to check this prior to handling the `PaymentSent` event because that will clear the entry from the cache entirely. But you want to check that information on a single path is cleared when `PaymentPathSuccessful` or `PaymentPathFailed` is handled. Incidentally, looks like these tests weren't updated when `PaymentPathSuccessful` was added which should be handled prior to `PaymentSent`. Let's make a separate test for this though instead of reusing this one.

lightning-invoice/src/payment.rs
```diff
@@ -1364,18 +1549,80 @@ mod tests {
...
+		invoice_payer.handle_event(&Event::PaymentPathSuccessful {
+			payment_id, payment_hash, path: route.paths[0].clone()
+		});
+
+		assert_eq!(inflight_map.get(&(0, true)), None);
+		assert_eq!(inflight_map.get(&(1, false)), None);
+		assert_eq!(inflight_map.get(&(2, true)), None);
+
+		// Second path should still be inflight
+		assert_eq!(inflight_map.get(&(1, true)).unwrap().clone(), 64);
```
jkczyz, line 1591: You'll need to create the in-flight map again before checking these expectations.

lightning-invoice/src/payment.rs
```diff
@@ -425,28 +495,56 @@ where
...
-				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, .. } => {
+				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, results } => {
+					// If a `PartialFailure` event returns a result that is either `Ok()` or an error
+					// with `MonitorUpdateFailed`, it means that the HTLCs for our payment is now
+					// either inflight, or being retried.
+					//
+					// Since the `results` field returned within `PartialFailure` are in the same
+					// order as our route hops, we can use the index for each `result` to look up its
+					// path in the `route` we found.
+					for (idx, result) in results.iter().enumerate() {
```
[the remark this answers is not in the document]
jurvis: Done in https://github.com/lightningdevkit/rust-lightning/pull/1643/commits/e8e05fdb287f8cd879f71068352be29967d94bad -- looking for suggestions on how to de-dupe the code, though.

jurvis:
I think this is ready for review :)

lightning-invoice/src/payment.rs
```diff
@@ -1440,11 +1805,24 @@ mod tests {
...
+			&self, payer: &PublicKey, route_params: &RouteParameters, _payment_hash: &PaymentHash,
+			_first_hops: Option<&[&ChannelDetails]>, scorer: &S
 		) -> Result<Route, LightningError> {
+			// Simulate calling the Scorer just as you would in find_route
+			let route = Self::route_for_value(route_params.final_value_msat);
+			for path in route.paths {
+				for hop in path {
+					let usage = ChannelUsage {
+						amount_msat: hop.fee_msat,
+						inflight_htlc_msat: hop.fee_msat,
```
jkczyz, line 1819: Should this be 0? The scorer here will be `AccountForInFlightHtlcs` wrapping `TestScorer`, IIUC. Or does this need to be a value for a particular test? Seems like `inflight_usage_trivial_test` fails when it is 0.

lightning-invoice/src/payment.rs
```diff
@@ -1464,6 +1842,7 @@ mod tests {
 
 	struct TestScorer {
 		expectations: Option<VecDeque<TestResult>>,
```
jkczyz, line 1844: Could you rename this `event_expectations` now that we have two sets?

lightning-invoice/src/payment.rs
```diff
@@ -477,7 +523,28 @@ where
...
+							Ok(_) => {
+								self.process_path_inflight_htlcs(payment_hash, &route.paths[idx]);
+							},
+							Err(err) => match err {
+								APIError::MonitorUpdateFailed => {
+									self.process_path_inflight_htlcs(payment_hash, &route.paths[idx]);
+								},
+								_ => {},
+							},
+						}
```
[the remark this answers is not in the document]
jurvis: ooo nice

lightning-invoice/src/payment.rs
```diff
@@ -1397,18 +1543,203 @@ mod tests {
...
+		let router = TestRouter {};
+		let scorer = RefCell::new(TestScorer::new());
+		let logger = TestLogger::new();
+		let invoice_payer =
+			InvoicePayer::new(&payer, router, &scorer, &logger, event_handler, Retry::Attempts(0));
+
+		invoice_payer.pay_invoice(&invoice_to_pay).unwrap();
+		let inflight_map = invoice_payer.create_inflight_map();
+
+		// Only the second path, that failed with `MonitorUpdateFailed` should be added to our
```
jkczyz, line 1752: s/that/which

lightning-invoice/src/payment.rs
```diff
@@ -1397,18 +1543,203 @@ mod tests {
...
+		let payment_hash = Some(PaymentHash(invoice_to_pay.payment_hash().clone().into_inner()));
+		let final_value_msat = invoice_to_pay.amount_milli_satoshis().unwrap();
+
+		let retry = TestRouter::retry_for_invoice(&invoice_to_pay);
+		let payer = TestPayer::new()
+			.fails_with_partial_failure(retry.clone(), OnAttempt(1),
+										Some(vec![
+											Ok(()),
+											Err(MonitorUpdateFailed)
+										]))
```
jkczyz, line 1772: Likewise.

lightning-invoice/src/payment.rs
```diff
@@ -1585,7 +1945,7 @@ mod tests {
 				return;
 			}
 
-			if let Some(expectations) = &self.expectations {
+			if let Some(expectations) = &self.event_expectations {
 				if !expectations.is_empty() {
 					panic!("Unsatisfied scorer expectations: {:?}", expectations);
 				}
```
jkczyz, line 2028: Could you do this for both sets of expectations? Otherwise, an unexercised expectation won't cause a test failure.

lightning-invoice/src/payment.rs
```diff
@@ -1397,18 +1543,203 @@ mod tests {
...
+		// have 64 msats inflight for paths considering the channel with scid of 1.
+		let payment_preimage_2 = PaymentPreimage([2; 32]);
+		let payment_invoice_2 = invoice(payment_preimage_2);
+		invoice_payer.pay_invoice(&payment_invoice_2).unwrap();
+	}
+
+	// This following partial failure tests are variants of `inflight_map_data_trivial_test` above,
+	// except with partial failure errors.
+	#[test]
+	fn partial_failure_with_monitor_update_failed_and_channel_unavailable() {
```
jkczyz, line 1726: Should be named in terms of inflight htlcs. Also, the specific type of error (e.g., channel unavailable) is unimportant so doesn't need to be in the name. Maybe something like `accounts_for_inflight_htlcs_sent_during_partial_failure`.

lightning-invoice/src/payment.rs
```diff
@@ -1397,18 +1543,203 @@ mod tests {
...
+		invoice_payer.pay_invoice(&invoice_to_pay).unwrap();
+		let inflight_map = invoice_payer.create_inflight_map();
+
+		// Only the second path, that failed with `MonitorUpdateFailed` should be added to our
+		// inflight map.
+		assert_eq!(inflight_map.len(), 1);
+	}
+
+	#[test]
+	fn partial_failure_with_one_monitor_update_failed() {
```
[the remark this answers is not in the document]
jurvis: `accounts_for_inflight_htlcs_while_waiting_for_monitor_update`?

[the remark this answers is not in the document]
jkczyz: Actually, given both are waiting for a monitor update, how about changing the other one to `accounts_for_some_inflight_htlcs_sent_during_partial_failure` and name this one `accounts_for_all_inflight_htlcs_sent_during_partial_failure`?

lightning-invoice/src/payment.rs
```diff
@@ -1637,6 +1654,72 @@ mod tests {
...
+		let final_value_msat = payment_invoice.amount_milli_satoshis().unwrap();
+		let router = TestRouter {};
+		let scorer = RefCell::new(TestScorer::new()
+			// 1st invoice, 1st path
+			.expect_usage(ChannelUsage { amount_msat: 10, inflight_htlc_msat: 0, effective_capacity: EffectiveCapacity::Unknown } )
+			.expect_usage(ChannelUsage { amount_msat: 20, inflight_htlc_msat: 0, effective_capacity: EffectiveCapacity::Unknown } )
+			.expect_usage(ChannelUsage { amount_msat: 64, inflight_htlc_msat: 0, effective_capacity: EffectiveCapacity::Unknown } )
+			// 1st invoice, 2nd path
+			.expect_usage(ChannelUsage { amount_msat: 64, inflight_htlc_msat: 0, effective_capacity: EffectiveCapacity::Unknown } )
+			// Retry 1 (called with `pay_internal`)
```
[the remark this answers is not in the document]
jurvis: ah, I meant to communicate here that we expect this retry to be called with `retry_payment` within `pay_internal`, whereas the latter is a second retry attempted by `retry_payment`.

the motivation for that is to make clear that "Retry 2" assertions help provide us with test coverage for the inflight HTLCs accounting in `retry_payment`. if that makes sense.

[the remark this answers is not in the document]
jurvis: I guess, it may be more precise to say, `handle_event`?

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jurvis: makes sense. I like that they're somewhat similar names so it shows that they are variants of each other

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jurvis: ignore my previous comments -- I know what you mean now. I think I just wanted to communicate that the first set of retry expectations are from our initial attempt and paying, whereas the second one is from a retry. 

I confused myself when I wrote those comments

lightning-invoice/src/payment.rs
```diff
@@ -223,6 +224,45 @@ impl<T: Time> PaymentInfo<T> {
 		}
 	}
 }
+
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+struct AccountForInFlightHtlcs<'a, S: Score> {
```
jkczyz, line 229: Forgot that this needs to implement `Writeable` for bindings release. See CI failures.

jkczyz: FWIW, it doesn't actually need to write anything though. May make sense to move that trait bound to `BackgroundProcessor` in a follow-up once scoring is moved into `DefaultRouter`.

lightning-invoice/src/payment.rs
```diff
@@ -1459,11 +1922,20 @@ mod tests {
...
-			&self, _short_channel_id: u64, _source: &NodeId, _target: &NodeId, _usage: ChannelUsage
-		) -> u64 { 0 }
+			&self, _short_channel_id: u64, _source: &NodeId, _target: &NodeId, usage: ChannelUsage
+		) -> u64 {
+			match self.scorer_expectations.borrow_mut().pop_front() {
+				Some(expectation) => {
+					assert_eq!(expectation.amount_msat, usage.amount_msat);
+					assert_eq!(expectation.inflight_htlc_msat, usage.inflight_htlc_msat);
+				},
+				None => {},
```
jkczyz, line 1932: I think we want an `Option` around `scorer_expectations` with `None` representing no expectations set (i.e., no checks). Otherwise, you can exhaust all expectations and not check these assertions if the scorer is still called.

lightning-invoice/src/payment.rs
jurvis: when you say it doesn't need to write anything, does that mean just returning `Ok(())` on `write`?

jkczyz: Yep

jurvis:
@jkczyz it looks like the remainder CI errors is related to stdlib's `io` not being available to us. do we have to implement something similar to what we do [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning-invoice/src/payment.rs#L1455-L1458) for TestScorer?

jkczyz:
> @jkczyz it looks like the remainder CI errors is related to stdlib's `io` not being available to us. do we have to implement something similar to what we do [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning-invoice/src/payment.rs#L1455-L1458) for TestScorer?

The problem is when compiling in `no-std` mode where `std` is completely unavailable. Typically, we use `core` instead and specifically the `core2::io` crate in this case, as `core` does not have `io`. See: https://github.com/lightningdevkit/rust-lightning/blob/414d8e1c4a9fcc8941c7af122a5da977e6c8720f/lightning/src/lib.rs#L91-L93

But I believe the code you linked to is what we want in `lightning-invoice`. Assuming we're not combining bindings and `no-std`, which I believe is the case.

You can also implement it as done there with `unreachable`. This is since `write` should never be called in the case of temporarily wrapping the user's scorer.

jurvis:
@jkczyz got it, thanks!

lightning-invoice/src/payment.rs
```diff
@@ -159,6 +159,8 @@ use core::ops::Deref;
 use core::time::Duration;
 #[cfg(feature = "std")]
 use std::time::SystemTime;
+use lightning::routing::gossip::NodeId;
+use lightning::util::errors::APIError;
```
jkczyz, line 163: Could you move these up with the other lightning imports, order accordingly?

lightning-invoice/src/payment.rs
```diff
@@ -477,7 +528,19 @@ where
 						Err(e)
 					}
 				},
-				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, .. } => {
+				PaymentSendFailure::PartialFailure { failed_paths_retry, payment_id, results } => {
+					// If a `PartialFailure` event returns a result that is an `Ok()`, it means that
+					// part of our payment is retried. When we receive `MonitorUpdateFailed`, it
+					// means that we are still waiting for our channel monitor update to be updated.
```
jkczyz, line 534: s/be updated/to complete

lightning-invoice/src/payment.rs
```diff
@@ -560,6 +677,46 @@ where
...
 		self.payment_cache.lock().unwrap().remove(payment_hash);
 	}
+
+	/// Given a [`PaymentHash`], this function looks up inflight path attempts in the payment_cache.
+	/// Then, it uses the path information inside the cache to construct a HashMap mapping a channel's
+	/// short channel id and direction to the amount being sent through it.
+	///
+	/// This function should be called whenever we need information about currently used up liquidity
+	/// across payments.
+	pub fn create_inflight_map(&self) -> HashMap<(u64, bool), u64> {
```
jkczyz, line 687: Does this need to be `pub`?

lightning-invoice/src/payment.rs
```diff
@@ -1397,18 +1564,263 @@ mod tests {
...
+			short_channel_id: None,
+			retry: Some(RouteParameters {
+				final_value_msat: final_value_msat / 2,
+				..TestRouter::retry_for_invoice(&payment_invoice)
+			}),
+		});
+	}
+
+	#[test]
+	fn accounts_for_all_inflight_htlcs_sent_during_partial_failure() {
```
jkczyz, line 1730: This test should be "some" and the other test should be "all" since `ChannelUnavailable` doesn't result in anything sent.

lightning-invoice/src/payment.rs
```diff
@@ -1397,18 +1564,263 @@ mod tests {
...
+							node_features: NodeFeatures::empty(),
+							short_channel_id: 0,
+							fee_msat: 10,
+							cltv_expiry_delta: 0
+						},
+						RouteHop {
+							pubkey: PublicKey::from_slice(&hex::decode("0324653eac434488002cc06bbfb7f10fe18991e35f9fe4302dbea6d2353dc0ab1c").unwrap()[..]).unwrap(),
+							channel_features: ChannelFeatures::empty(),
+							node_features: NodeFeatures::empty(),
+							short_channel_id: 1,
```
jkczyz, line 1811: So, technically, this isn't possible if the second path has a single hop from the payer to the recipient with the same `short_channel_id`, as it would mean the paths have different recipients. Then again, that was the case prior to your changes, so maybe shouldn't worry about making this realistic. It just confused me as to why the last test above has 4 hops when there should have been an overlap. But now it makes sense.

jkczyz (commented):
LGTM other than a few nits.

lightning-invoice/src/payment.rs
jurvis: oh, you're right. oops.

lightning-invoice/src/payment.rs
jurvis: hm, why will they have different recipients if the pubkey is the same?

jkczyz: The last hop in each path is the recipient, which are different here. Additionally, `TestPayer::node_id` returns pubkey `"032..."`, which is same as the node through hop with `short_channel_id` 1. This means that this hop connects different nodes on each path, which wouldn't be possible in practice.

So in actuality:
- the first path contains a cycle back to the sender on the second hop
- the second path has one hop going directly back to the sender (itself), but with a `short_channel_id` that conflicts with a hop in the first path 
- we see four hops in the map instead of three because on the first path the hop's key is `(1, true)` (i.e., `"02e..." < "032..."`) but on the second path it is `(1, false)` (i.e., `"032..." < "032..."`)

FWIW, this partly predates your change. If we want this to be both realistic and have appropriate test coverage, we should:
- Change `TestPayer::node_id` return a pubkey not used in any of the hops.
- Change the second path to have three hops, at least one of which doesn't overlap with the first path. Maybe two hops would suffice depending on how the tests are formulated.
- Re-number the hop's `short_channel_id` starting from 1 since your code uses 0 for a fake hop, which currently conflicts with the first path's first hop, though doesn't cause any problems.

But this change may affect preexisting tests.

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +189,85 @@ where
...
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+struct AccountForInFlightHtlcs<'a, S: Score> {
+	scorer: &'a mut S,
+	/// Maps a channel's short channel id and its direction to the liquidity used up.
+	inflight_htlcs: HashMap<(u64, bool), u64>,
+}
+
+#[cfg(c_bindings)]
+impl<'a, S:Score> lightning::util::ser::Writeable for AccountForInFlightHtlcs<'a, S> {
+	fn write<W: lightning::util::ser::Writer>(&self, writer: &mut W) -> Result<(), std::io::Error> { unreachable!() }
```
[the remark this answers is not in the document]
jurvis: Will that mean we'll need to change the `Score` trait bound to `WriteableScore`? I think [we were thinking](https://github.com/lightningdevkit/rust-lightning/pull/1643#discussion_r953833477) of alternatively just returning `Ok(())` and not doing anything here.

thoughts?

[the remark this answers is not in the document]
jurvis: ah, got it!

lightning-invoice/src/payment.rs
```diff
@@ -188,10 +189,85 @@ where
...
 }
 
+/// Used by [`InvoicePayerUsingTime::payment_cache`] to track the payments that are either
+/// currently being made, or have outstanding paths that need retrying.
+struct PaymentInfo<T: Time> {
+	attempts: PaymentAttempts<T>,
+	paths: Vec<Vec<PaymentInfoRouteHop>>,
+}
+
+/// Subset of properties that is represented in `RouteHop`, so we only store what we need in
```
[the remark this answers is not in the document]
jurvis: makes sense. there was a suggestion earlier than we could expand this to track CLTV expiry too, so it does make sense just to use `RouteHop.

lightning-invoice/src/payment.rs
jurvis: ahhh, gotcha. I didn't notice that about `TestPayer::node_id`. Makes sense now. I'll make the changes.

jurvis: hm, I just checked, it looks like `TestPayer::node_id` doesn't actually return `"32..."`?

payer pubkey:
image

1st path, 2nd hop pubkey
image


`TestPayer::node_id` call:

```
		fn node_id(&self) -> PublicKey {
			PublicKey::from_secret_key(&secp_ctx, &SecretKey::from_slice(&[42; 32]).unwrap())
		}
```

am i missing something?

jurvis: > since your code uses 0 for a fake hop

we actually never hardcode any of the channel ids. if you're referring to the `0` in `create_inflight_map`, that's meant to represent a dummy hop to our source, so that we can get the payer's node information when iterating backward of the path to construct the map

jurvis: think here's what I'm going to do:

```
       —>—————>— [ 029e... ] ————>—————
      /                                                                           \
   (3)                                                                         (4)
    /                                                                                \
[ Us ] --(0)--> [ 02ee... ] --(1)--> [ 032... ] --2--> [ 027f... ]
```

1. Adjust the last path to have 2 hops, one to `029e...` and then `027f...`
2. Make sure all the scids are unique
3. Update tests

lightning-invoice/src/payment.rs
```diff
@@ -527,6 +677,46 @@ where
...
+			for path in &payment_info.paths {
+				if path.is_empty() { break };
+				// total_inflight_map needs to be direction-sensitive when keeping track of the HTLC value
+				// that is held up. However, the `hops` array, which is a path returned by `find_route` in
+				// the router excludes the payer node. In the following lines, the payer's information is
+				// hardcoded with an inflight value of 0 so that we can correctly represent the first hop
+				// in our sliding window of two.
+				let our_node_id: PublicKey = self.payer.node_id();
+				let reversed_hops_with_payer: Vec<(u64, PublicKey)> = core::iter::once((0u64, our_node_id)).chain(
+					path.split_last().unwrap().1.iter().map(|hop| (hop.fee_msat, hop.pubkey))).rev().collect();
```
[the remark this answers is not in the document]
jurvis: > Additionally, could possibly use `iter().skip(1).map(...)` instead of `split_last().unwrap().1.iter().map(...)`

I don't think this is equivalent since the former skips the first item in the collection, whereas I'm trying to drop the last element out of my collection

lightning-invoice/src/payment.rs
jurvis: Resolved in https://github.com/lightningdevkit/rust-lightning/pull/1643/commits/1abd3df8ab460d4b7fffd32dae5ae7419a8739d5

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jkczyz: Ah, right. I think I misread the docs, which are worded backwards wrt the returned tuple.

Given you are using `rev()` with the `chain()`, you can do `iter().rev().skip(1).map(...).chain(core::iter::once(...))`

lightning-invoice/src/payment.rs
```diff
@@ -1407,11 +1841,29 @@ mod tests {
...
+		fn find_route<S: Score>(
+			&self, payer: &PublicKey, route_params: &RouteParameters, _payment_hash: &PaymentHash,
+			_first_hops: Option<&[&ChannelDetails]>, scorer: &S
 		) -> Result<Route, LightningError> {
+			// Simulate calling the Scorer just as you would in find_route
+			let route = Self::route_for_value(route_params.final_value_msat);
+			for path in route.paths {
+				for (idx, hop) in path.iter().enumerate() {
+					let usage = ChannelUsage {
+						amount_msat: hop.fee_msat,
```
jkczyz, line 1854: I believe the realistic value here would not just be the fees but the aggregate `fee_msat` from the last hop backwards to this one. You can walk the path backwards to compute this and reverse the expectations in the tests, as that is typically how the scorer would be called.

lightning-invoice/src/payment.rs
jkczyz: > hm, I just checked, it looks like `TestPayer::node_id` doesn't actually return `"32..."`?
> 

Whoops, I was using `42` hex instead of `42` decimal. 😅 But note that the first byte as `44` decimal corresponds to the last byte as `2c` hexadecimal. The full pubkey should be:

`035be5e9478209674a96e60f1f037f6176540fd001fa1d64694770c56a7709c42c`

jkczyz (commented):
Ok, I'm happy with this other than one test change to make it more realistic.

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jurvis: ooo yeah, that's clever.

jkczyz (commented):
LGTM. @TheBlueMatt could you take another look?

lightning-invoice/src/payment.rs
```diff
@@ -466,16 +544,36 @@ where
 		}.map_err(|e| PaymentError::Sending(e))
 	}
 
+	// Takes in a path to have its information stored in `payment_cache`. This is done for paths
+	// that are pending retry.
+	fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, path: &Vec<RouteHop>) {
+		self.payment_cache.lock().unwrap().entry(payment_hash)
+			.or_insert_with(|| PaymentInfo::new())
+			.paths.push(path.clone());
```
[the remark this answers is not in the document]
jurvis: hm, currently the `path` that gets passed in here in a reference, but `push` is expecting a value, that's why I  `clone`-d. is that not correct?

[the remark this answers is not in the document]
jkczyz: You can have `process_path_inflight_htlcs` take the path by value instead of by reference to avoid the clone. This is because at call sites you have ownership of the `Route`. See how @TheBlueMatt uses `drain` to take ownership during iteration and `zip` to iterate over both paths and results in parallel here: https://github.com/lightningdevkit/rust-lightning/pull/1643#discussion_r956200113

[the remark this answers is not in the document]
jkczyz: Though maybe a little cleaner with `into_iter` instead of `drain`:

```diff
diff --git a/lightning-invoice/src/payment.rs b/lightning-invoice/src/payment.rs
index 8497ca8c..4f76c709 100644
--- a/lightning-invoice/src/payment.rs
+++ b/lightning-invoice/src/payment.rs
@@ -493,7 +493,7 @@ where
                match send_payment(&route) {
                        Ok(payment_id) => {
                                for path in route.paths {
-                                       self.process_path_inflight_htlcs(payment_hash, &path);
+                                       self.process_path_inflight_htlcs(payment_hash, path);
                                }
                                Ok(payment_id)
                        },
@@ -515,10 +515,10 @@ where
                                        // If a `PartialFailure` event returns a result that is an `Ok()`, it means that
                                        // part of our payment is retried. When we receive `MonitorUpdateFailed`, it
                                        // means that we are still waiting for our channel monitor update to be completed.
-                                       for (idx, result) in results.iter().enumerate() {
+                                       for (path, result) in route.paths.into_iter().zip(results.iter()) {
                                                match result {
                                                        Ok(_) | Err(APIError::MonitorUpdateFailed) => {
-                                                               self.process_path_inflight_htlcs(payment_hash, &route.paths[idx]);
+                                                               self.process_path_inflight_htlcs(payment_hash, path);
                                                        },
                                                        _ => {},
                                                }
@@ -545,7 +545,7 @@ where
 
        // Takes in a path to have its information stored in `payment_cache`. This is done for paths
        // that are pending retry.
-       fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, path: &Vec<RouteHop>) {
+       fn process_path_inflight_htlcs(&self, payment_hash: PaymentHash, path: Vec<RouteHop>) {
                self.payment_cache.lock().unwrap().entry(payment_hash)
                        .or_insert_with(|| PaymentInfo::new())
                        .paths.push(path.clone());
@@ -603,7 +603,7 @@ where
                match self.payer.retry_payment(&route.as_ref().unwrap(), payment_id) {
                        Ok(()) => {
                                for path in route.unwrap().paths {
-                                       self.process_path_inflight_htlcs(payment_hash, &path);
+                                       self.process_path_inflight_htlcs(payment_hash, path);
                                }
                                Ok(())
                        },
@@ -619,10 +619,10 @@ where
                                // If a `PartialFailure` error contains a result that is an `Ok()`, it means that
                                // part of our payment is retried. When we receive `MonitorUpdateFailed`, it
                                // means that we are still waiting for our channel monitor update to complete.
-                               for (idx, result) in results.iter().enumerate() {
+                               for (path, result) in route.unwrap().paths.into_iter().zip(results.iter()) {
                                        match result {
                                                Ok(_) | Err(APIError::MonitorUpdateFailed) => {
-                                                       self.process_path_inflight_htlcs(payment_hash, &route.as_ref().unwrap().paths[idx]);
+                                                       self.process_path_inflight_htlcs(payment_hash, path);
                                                },
                                                _ => {},
                                        }

```

For reference on the different type of iterators and how rust may infer which one is used:

https://doc.rust-lang.org/std/iter/index.html#the-three-forms-of-iteration
https://doc.rust-lang.org/std/iter/index.html#for-loops-and-intoiterator
https://doc.rust-lang.org/std/iter/index.html#iterating-by-reference

lightning-invoice/src/payment.rs
```diff
@@ -1364,24 +1532,290 @@ mod tests {
...
+		let payment_preimage = PaymentPreimage([1; 32]);
+		let invoice = invoice(payment_preimage);
+		let payment_hash = Some(PaymentHash(invoice.payment_hash().clone().into_inner()));
+		let final_value_msat = invoice.amount_milli_satoshis().unwrap();
+
+		let payer = TestPayer::new().expect_send(Amount::ForInvoice(final_value_msat));
+		let final_value_msat = invoice.amount_milli_satoshis().unwrap();
+		let route = TestRouter::route_for_value(final_value_msat);
+		let router = TestRouter {};
+		let scorer = RefCell::new(TestScorer::new());
```
adamritter, line 1549: Are these RefCells really needed? Wouldn't just creating TestScorers work generally?

jurvis: I believe this is because `InvoicePayer` expects the scorer to implement `LockableScore`. 

`TestScorer` here is really just a mocked version of a scorer that we'd expect users to implement on their own

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jurvis: ah, I see. thanks for taking the time to put together the diffs and explaining how ownership works. I definitely need to read up more on it, lol

lightning-invoice/src/payment.rs
jkczyz: Yeah, a scorer needs to be locked to (a) use its mutable interface in `InvoicePayer`'s event handling and (b) use its immutable interface when routing such that its state doesn't change between calls. Additionally, another reference to it may be needed for serialization.

adamritter: I see, thanks! MinCostFlowRouter pull request is at a point where it needs to be tested together with InvoicePayer and tracking in-flight HTLCs, but in a conversion with Matt Corallo, he said that as MinCostFlowRouter needs to access liquidity tracking, but not a custom scorer provided by LDK users, LiquidityTracker should be a separate trait implemented by the scorers. Another option in my opinion would be keeping liquidity tracking from scoring totally separate, and DefaultRouter should request the estimated liquidity information and provide it to the immutable Scorer (that can be customized by LDK user, while liquidity tracking implementation probably shouldn't be surfaced in the API).

What do you guys think?

Here's a link to the comments:

https://github.com/lightningdevkit/rust-lightning/pull/1668#issuecomment-1230544540

jkczyz: Yeah, I think we want at least two follow-ups to this PR:

(1) Move the `LockableScore` type parameter to `DefaultRouter` since some routers (e.g., remote server routing) don't need a scorer.
(2) Move liquidity tracking to either the `Payer` trait or a dedicated one, having `InvoicePayer` query it and pass the liquidity information to `Router` akin to how it currently passes a scorer.

The latter comes from an observation that I made in review club. `ChannelManager` actually knows all the  liquidity information given it is the one producing `PaymentPathFailed` and `PaymentPathSuccessful` events. It also would make liquidity tracking persist across restarts.

See also #1157 for exposing pending payments more generally, which should be considered when making this change.

lightning-invoice/src/payment.rs
```diff
@@ -527,6 +644,47 @@ where
...
+					.chain(core::iter::once((0u64, our_node_id)));
+				let mut cumulative_msat = 0;
+
+				// Taking the reversed vector from above, we zip it with just the reversed hops list to
+				// work "backwards" of the given path, since the last hop's `fee_msat` actually represents
+				// the total amount sent.
+				for (next_hop, prev_hop) in path.iter().rev().zip(reversed_hops_with_payer) {
+					cumulative_msat += next_hop.fee_msat;
+					total_inflight_map
+						.entry((next_hop.short_channel_id, prev_hop.1 < next_hop.pubkey))
```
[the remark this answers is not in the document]
jurvis: ah, good to know. thanks!

jkczyz (approved):

PR #1652, opened 2022-08-05
Onion messages: support reply paths 

valentinewallace:
Adds support for sending reply paths in our onion messages, and for decoding them on onion message read.


lightning/src/onion_message/blinded_route.rs
```diff
@@ -113,6 +114,56 @@ fn encrypt_payload<P: Writeable>(payload: P, encrypted_tlvs_ss: [u8; 32]) -> Vec
...
+	}
+}
+
+impl LengthReadable for BlindedRoute {
+	fn read<R: LengthRead>(r: &mut R) -> Result<Self, DecodeError> {
+		let introduction_node_id = Readable::read(r)?;
+		let blinding_point = Readable::read(r)?;
+		let mut blinded_hops: Vec<BlindedHop> = Vec::new();
+		let mut bytes_read = 66;
+		while bytes_read < r.total_bytes() {
```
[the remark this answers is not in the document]
valentinewallace: It should be possible to get a legitimate `ShortRead` error in `BlindedHop::readable` too. Not sure how we can tell which it is without complicating things?

lightning/src/onion_message/messenger.rs
```diff
@@ -142,7 +142,7 @@ impl<Signer: Sign, K: Deref, L: Deref> OnionMessenger<Signer, K, L>
 
 	/// Send an empty onion message to `destination`, routing it through `intermediate_nodes`.
 	/// See [`OnionMessenger`] for example usage.
-	pub fn send_onion_message(&self, intermediate_nodes: &[PublicKey], destination: Destination) -> Result<(), SendError> {
+	pub fn send_onion_message(&self, intermediate_nodes: &[PublicKey], destination: Destination, reply_path: Option<BlindedRoute>) -> Result<(), SendError> {
```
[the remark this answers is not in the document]
valentinewallace: Hmm I think that makes sense. This API may remain pretty low-level and keep a version of the parameter, but we'll need some way of getting a `reply_path` for outbound `invoice_request`s, likely via the `router` you mention.

lightning/src/onion_message/blinded_route.rs
```diff
@@ -113,6 +114,52 @@ fn encrypt_payload<P: Writeable>(payload: P, encrypted_tlvs_ss: [u8; 32]) -> Vec
 	writer.0
 }
 
+impl Writeable for BlindedRoute {
+	fn write<W: Writer>(&self, w: &mut W) -> Result<(), io::Error> {
+		self.introduction_node_id.write(w)?;
+		self.blinding_point.write(w)?;
+		for hop in &self.blinded_hops {
```
jkczyz, line 121: Looks like we'll want to write the number of hops as a `u8` in order to mach the encoding used in the `offer` message. Not in the latest spec draft, but see discord discussion.

lightning/src/onion_message/messenger.rs
```diff
@@ -209,9 +209,17 @@ impl<Signer: Sign, K: Deref, L: Deref> OnionMessenger<Signer, K, L>
...
-				log_info!(self.logger, "Received an onion message with path_id: {:02x?}", path_id);
+				log_info!(self.logger,
+					"Received an onion message with path_id: {:02x?} and {}reply_path",
+						path_id, if reply_path.is_some() { "" } else { "no " });
+				#[cfg(test)]
+				{
+					if let Some(reply_path) = reply_path {
+						self.send_onion_message(&[], Destination::BlindedRoute(reply_path), None).unwrap();
+					}
+				}
```
jkczyz, line 222: Hmmm... could we just put this in the relevant tests?

lightning/src/onion_message/blinded_route.rs
```diff
@@ -113,6 +114,52 @@ fn encrypt_payload<P: Writeable>(payload: P, encrypted_tlvs_ss: [u8; 32]) -> Vec
...
+					if !tracking_reader.have_read {
+						break
+					}
+					return Err(DecodeError::ShortRead)
+				},
+				Err(e) => return Err(e)
+			};
+			blinded_hops.push(blinded_hop);
+		}
+		if blinded_hops.len() == 0 {
```
jkczyz, line 147: nit: `is_empty`

lightning/src/onion_message/blinded_route.rs
```diff
@@ -113,6 +114,52 @@ fn encrypt_payload<P: Writeable>(payload: P, encrypted_tlvs_ss: [u8; 32]) -> Vec
...
+				Ok(h) => h,
+				Err(DecodeError::ShortRead) => {
+					if !tracking_reader.have_read {
+						break
+					}
+					return Err(DecodeError::ShortRead)
+				},
+				Err(e) => return Err(e)
+			};
+			blinded_hops.push(blinded_hop);
```
jkczyz, line 145: This can be simplified to:

```rust
match Readable::read(&mut tracking_reader) {
    Ok(h) => blinded_hops.push(h),
    Err(DecodeError::ShortRead) if !tracking_reader.have_read => break,
    Err(e) => return Err(e),
}
```

lightning/src/onion_message/messenger.rs
valentinewallace: I'm not sure how, unless we wanna add some test-only struct field handler for reply paths instead?

lightning/src/onion_message/blinded_route.rs
valentinewallace: Updated to this

lightning/src/onion_message/messenger.rs
jkczyz: To take a step back, I guess it was unclear to me what new behavior we're trying to test. Looks like it's primarily that `send_onion_message` will *encode a reply path* for the destination. Would it be sufficient to simply update the logging assertion in `pass_along_path` to check if a reply path was given?

Presumably, after handling is added, both the test-only code and logging assertions would no longer be necessary as handling could be tested properly. Otherwise, seems like we're adding test-only behavior to test a behavior that doesn't yet exist (i.e., *using a reply path* when handling).

valentinewallace: >  Looks like it's primarily that send_onion_message will encode a reply path for the destination.

FWIW, my thinking was that we're not just testing this^, but also that we can actually use a reply path that we decode, and make sure that reply reaches its destination (so just feature correctness basically). 

I don't feel too strongly though, since it's fair to say that it'll be tested later anyway. Thoughts?

jkczyz: Isn't that testing the test rather than the system under test? (i.e., testing that the correct reply path was passed to `send_onion_message`)

valentinewallace: I mean (a version of) this code will be not-`#[cfg(test)]` eventually. But it is fair that the behavior doesn't exist now, I'll remove this + the relevant tests

jkczyz (approved):

valentinewallace:
Had to rebase to repro the test failure locally. Also removed some warnings that I didn't realize appeared on the last push.

lightning/src/onion_message/messenger.rs
```diff
@@ -344,6 +347,7 @@ fn packet_payloads_and_keys<T: secp256k1::Signing + secp256k1::Verification>(
 		} else if let Some(encrypted_payload) = enc_payload_opt {
 			payloads.push((Payload::Receive {
 				control_tlvs: ReceiveControlTlvs::Blinded(encrypted_payload),
+				reply_path: reply_path.take(),
```
jkczyz, line 350: IIUC, `reply_path` is captured by reference (I think?) in the closure, so it will be `None` if this case is hit more than once when called by `construct_keys_callback`. Is this the desired behavior? I guess so because there would only be one `Payload::Receive`?

Related, the same may occur below where `reply_path.take()` is called again outside of the closure. Partly trying to check my understanding if that case will be hit below but not in the closure.

valentinewallace: Yep, that's intentional. There should only be one `Payload::Receive` pushed, so `reply_path` should only be `take`n once.

jkczyz (approved):

PR #1659, opened 2022-08-09
Fix payment onion payload decode 

valentinewallace:
Prior to this change, we could have failed to decode a valid payload of size \>253. This is because we were decoding the length (a BigSize, big-endian) as a VarInt (little-endian).

Plus some minor cleanups. Pair-debugged with @wpaulino.

### DIFF
--- a/lightning/src/ln/msgs.rs
+++ b/lightning/src/ln/msgs.rs
@@ -1417,14 +1417,6 @@ impl Writeable for OnionHopData {
 	}
 }
 
-// ReadableArgs because we need onion_utils::decode_next_hop to accommodate payment packets and
-// onion message packets.
-impl ReadableArgs<()> for OnionHopData {
-	fn read<R: Read>(r: &mut R, _arg: ()) -> Result<Self, DecodeError> {
-		<Self as Readable>::read(r)
-	}
-}
-
 impl Readable for OnionHopData {
 	fn read<R: Read>(mut r: &mut R) -> Result<Self, DecodeError> {
 		use bitcoin::consensus::encode::{Decodable, Error, VarInt};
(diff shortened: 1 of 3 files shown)


dunxen (dismissed):
ACK 15074bd

tnull:
LGTM!

Out of curiosity: should [this `VarInt`](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/onion_message/packet.rs#L165) also be a `BigSize`?

lightning/src/ln/msgs.rs
```diff
@@ -2824,4 +2819,40 @@ mod tests {
...
+		use util::ser::HighZeroBytesDroppedBigSize;
+		let payload = msgs::OnionHopData {
+			format: OnionHopDataFormat::NonFinalNode {
+				short_channel_id: 0xdeadbeef1bad1dea,
+			},
+			amt_to_forward: 1000,
+			outgoing_cltv_value: 0xffffffff,
+		};
+		let mut encoded_payload = Vec::new();
+		let test_bytes = vec![42 as u8; 1000];
```
tnull, line 2847:
```suggestion
		let test_bytes = vec![42u8; 1000];
```

valentinewallace:
> LGTM!
> 
> Out of curiosity: should [this `VarInt`](https://github.com/lightningdevkit/rust-lightning/commit/bf007ea7632b7b95c280440bc50713e55784e315#diff-e7b166fcbf13dedcac0a1f9fbb653da97a61df2f60a825177385f59d513ff35dR165) also be a `BigSize`? Or am I looking at the [wrong place](https://github.com/lightning/bolts/commit/d5cbac3f4cce0a4306ab2005f192aee2bf2432da#diff-ad65f0beaac5cef88f5fd7a8b9ca36cbc5a790f4815b4b50b9ea794a55aaf012R624) of the spec?

Yup, exactly! That fix was needed for https://github.com/lightningdevkit/rust-lightning/pull/1652 (and covered by the tests added) so I stuck it in there.

dunxen (approved):
reACK dfbebbf

tnull:
> Yup, exactly! That fix was needed for #1652 (and covered by the tests added) so I stuck it in there.

Ah, thought so!

tnull (approved):

Issue #1607, opened 2022-07-11
Onion message follow-ups 


valentinewallace:
- [ ] [Force users to use `path_id`](https://github.com/lightningdevkit/rust-lightning/pull/1503#discussion_r917414102)
- [ ] Enable the addition of dummy blinded hops for increased privacy
- [ ] On creation, pad blinded route hops to be equal length 
- [x] Support 1-hop blinded paths
- [ ] Rate limit based on channel vs non-channel peers
- [x] Rate limit based on outbound peer buffer limits
- [x] Refuse to send onion messages to offline peers and drop forwards to offline peers
- [ ] Config option for forwarding over channels-only
- [x] Custom TLVs, sending-side + handler on receiving side
- [ ] Implement BOLT spec test vectors
- [ ] Consider setting max number of hops before refusing to send
- [ ] Custom `get_route` for finding an OM path (current `get_route` has a lot of custom stuff for payments)
- [x] Flip on OM feature bit (may want to conditionally flip based on whether a user has an `OnionMessenger`)
- [ ] Refactor `onion_utils::construct_onion_packet_with_init_noise` to take one Vec instead of two
- [ ] Minor: make `onion_utils::decode_next_hop` private, see https://github.com/lightningdevkit/rust-lightning/pull/1503#discussion_r935941292 
- [x] Add fuzz testing 
- [ ] Don't disconnect peers on garbage onion messages (cc https://github.com/lightningdevkit/rust-lightning/pull/1503#discussion_r903911989)

valentinewallace:
> Support receiving onion messages to phantom nodes.

cc https://github.com/lightningdevkit/rust-lightning/pull/1503#discussion_r920330843

valentinewallace:
Due to recent spec updates forcing the use of blinded paths, we'll want to support 1-hop blinded paths now too.

valentinewallace:
> We should presumably include padding (someone taking over #1771) no?

Yeah, that would be good.

PR #1669, opened 2022-08-15
Support onion message pathfinding 

valentinewallace:
We *could* reuse router::get_route, but it's better to start from scratch
because get_route includes a lot of payment-specific computations that impact
performance.

Partially addresses #1607


lightning/src/onion_message/router.rs
```diff
@@ -0,0 +1,248 @@
...
+	FinalHopConversion,
+}
+
+//  Heavily adapted from https://github.com/samueltardieu/pathfinding/blob/master/src/directed/dijkstra.rs
+//  TODO: how2credit the repo (is that necessary?)?
+/// Run Dijkstra's from `start` until `found_target` indicates that we've found the destination.
+/// `successor_callback` must invoke the callback that it is provided on each of a given node's
+/// next-hop peers. `convert_final_hop` may be used to convert an intermediate processing hop type
+/// (`N`) to a final path hop type (`H`).
+fn dijkstra<N, H, FN, FS, FC>(start: N, successor_callback: &mut FN, found_target: FS,
```
tnull, line 89: This interface seems to be kept very generic to accommodate different usage scenarios by the original author. I think when we're adapting this we can fill in a lot of these types, which should improve readability?

lightning/src/onion_message/router.rs
```diff
@@ -0,0 +1,248 @@
...
+	}
+}
+
+// Errors if `convert_path_hop` fails.
+fn reverse_path<N, H, FC>(parents: HashMap<N, (N, u64)>, start: N, convert_path_hop: &mut FC) -> Result<Vec<H>, ()>
+	where N: Eq + Hash + Copy + Ord,
+				FC: FnMut(&N) -> Result<H, ()>,
+{
+	let mut path = vec![convert_path_hop(&start)?];
+	let mut curr = start;
```
tnull, line 142: IIUC, this `start` is actually the target hop. So renaming the variable accordingly could spare future confusion.

valentinewallace:
Ok it's clearly less readable than I thought with a generic Dijkstra's. Staying generic would allow us to pretty much reuse the test vectors in the original repo, but it doesn't seem worth. Gonna convert to draft for now and rewrite the generic portions

tnull:
> Ok it's clearly less readable than I thought with a generic Dijkstra's. Staying generic would allow us to pretty much reuse the test vectors in the original repo, but it doesn't seem worth. Gonna convert to draft for now and rewrite the generic portions

Ah, didn't realize keeping it generic had that intention. Well, it's not *unreadable* and reusing test vectors may be worth it. 
If we however also see a way of refactoring the payment pathfinding to reuse the same generic implementation, we'd should really consider it, IMO.

valentinewallace:
> I believe the upstream dijkstras you pulled from is actually buggy, anyway, so the generic-ness isn't worth much :p

Lol I might've introduced bugs in my quest to eliminate all vecs from the original, to be fair

valentinewallace:
> > Ok it's clearly less readable than I thought with a generic Dijkstra's. Staying generic would allow us to pretty much reuse the test vectors in the original repo, but it doesn't seem worth. Gonna convert to draft for now and rewrite the generic portions
> 
> Ah, didn't realize keeping it generic had that intention. Well, it's not _unreadable_ and reusing test vectors may be worth it. If we however also see a way of refactoring the payment pathfinding to reuse the same generic implementation, we'd should definitely consider it, IMO.

I did wonder if there were other cases where we'd want a generic implementation. I have no idea whether that'd be realistic with our payment pathfinding though lol, we'd need a way to "decrement available liquidity" for hops in paths that were already found, at least, but maybe that's easy? 🤔

lightning/src/onion_message/router.rs
```diff
@@ -0,0 +1,248 @@
...
+
+use alloc::collections::BinaryHeap;
+use core::hash::Hash;
+use core::ops::Deref;
+use prelude::*;
+
+/// Find a path for sending an onion message.
+pub fn find_path<L: Deref, GL: Deref>(
+	our_node_pubkey: &PublicKey, receiver_pubkey: &PublicKey, network_graph: &NetworkGraph<GL>, first_hops: Option<&[&PublicKey]>, logger: L
+) -> Result<Vec<PublicKey>, LightningError> where L::Target: Logger, GL::Target: Logger
```
ariard, line 15: So we wouldn't reuse all the information we're learning and storing in our `ProbabilisticScorer`. I can see how we're limited with the current penalty being based on the link-level and here we might be interested by node-level reliability in the path construction. That said, we can also assume that a reliable channel == a reliable onion message communication channel and go with it. I don't know if the spec says anything here or what are the thinking of other implementations ?

valentinewallace:
Thanks for all the feedback! This needs a lot of work and I'm switching it up to work on custom OMs first -- closing for now.

lightning/src/onion_message/router.rs
```diff
@@ -0,0 +1,248 @@
...
+enum Error {
+	/// No path exists to the destination.
+	PathNotFound,
+	/// Converting the processing hop type to the final hop type failed, see `dijkstra`'s
+	/// `convert_final_hop` parameter.
+	FinalHopConversion,
+}
+
+//  Heavily adapted from https://github.com/samueltardieu/pathfinding/blob/master/src/directed/dijkstra.rs
+//  TODO: how2credit the repo (is that necessary?)?
```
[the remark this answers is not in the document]
valentinewallace: #1724 ended up not adapting this crate's implementation anymore (shoutout to Wikpedia)

PR #1681, opened 2022-08-23
Workaround GH Actions issue for 1.45

dunxen:
See https://github.com/rust-lang/cargo/issues/10303

It seems we were only having that particular issue when pinning tokio to 1.14 for Rust 1.45. I'm not sure if it happens at later steps and didn't want to enable these env vars globally across the other jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,8 @@ jobs:
       - name: Pin tokio to 1.14 for Rust 1.45
         if: "matrix.build-net-old-tokio"
         run: cargo update -p tokio --precise "1.14.0" --verbose
+        env:
+          CARGO_NET_GIT_FETCH_WITH_CLI: "true"
       - name: Build on Rust ${{ matrix.toolchain }} with net-tokio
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -62,6 +62,9 @@ jobs:
       - name: Pin tokio to 1.14 for Rust 1.45
         if: "matrix.build-net-old-tokio"
         run: cargo update -p tokio --precise "1.14.0" --verbose
+        env:
+          CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+          CARGO_HTTP_MULTIPLEXING: "false"
```
tnull, line 67: IIUC, this second line would only be necessary if we used a proxy? Sure this is required for Github Actions?

tnull (approved):
You, sir, are a hero!

.github/workflows/build.yml
dunxen: Ah yes, looks like you're right. Let's remove it in that case since it's redundant.

tnull (approved):

jkczyz (approved):

PR #1685, opened 2022-08-25
Miscellaneous anchor outputs preparatory changes

wpaulino:
This PR includes a few miscellaneous changes that will be required for the integration of anchors outputs. They are standalone and should be fine to merge without the remaining changes to come to help split up the upcoming work into multiple PRs.


lightning/src/ln/chan_utils.rs
```diff
@@ -616,12 +629,17 @@ pub fn build_htlc_transaction(commitment_txid: &Txid, feerate_per_kw: u32, conte
...
+		htlc.amount_msat / 1000
+	} else {
+		let total_fee = feerate_per_kw as u64 * weight / 1000;
+		htlc.amount_msat / 1000 - total_fee
+	};
 
 	let mut txouts: Vec<TxOut> = Vec::new();
 	txouts.push(TxOut {
 		script_pubkey: get_revokeable_redeemscript(revocation_key, contest_delay, broadcaster_delayed_payment_key).to_v0_p2wsh(),
-		value: htlc.amount_msat / 1000 - total_fee //TODO: BOLT 3 does not specify if we should add amount_msat before dividing or if we should divide by 1000 before subtracting (as we do here)
```
dunxen, line 624: Do we need to keep this TODO for the above change?

As far as I can infer from the spec it seems that what we do here isn't wrong but probably should be made clearer there. At least "`Multiply feerate_per_kw by xxx (xxx if option_anchor_outputs applies) and divide by 1000 (rounding down)`" seems to match what we do.

arik-so (dismissed):

lightning/src/ln/chan_utils.rs
dunxen: Ok, it's here: https://github.com/lightning/bolts/pull/1020

dunxen (approved):

lightning/src/ln/chan_utils.rs
```diff
@@ -66,14 +66,27 @@ pub(crate) enum HTLCType {
...
 
 impl HTLCType {
-	/// Check if a given tx witnessScript len matchs one of a pre-signed HTLC
-	pub(crate) fn scriptlen_to_htlctype(witness_script_len: usize) ->  Option<HTLCType> {
-		if witness_script_len == 133 {
-			Some(HTLCType::OfferedHTLC)
-		} else if witness_script_len >= 136 && witness_script_len <= 139 {
-			Some(HTLCType::AcceptedHTLC)
+	/// Check if a given tx witnessScript len matchs one of a pre-signed HTLC. The returned
+	/// `HTLCType` can match incorrectly based on the value of `opt_anchors`.
```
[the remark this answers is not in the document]
dunxen: Looks like consensus to have it be zero: https://github.com/bitcoin/bips/blob/master/bip-0147.mediawiki

lightning/src/ln/chan_utils.rs
```diff
@@ -60,18 +60,56 @@ pub fn htlc_timeout_tx_weight(opt_anchors: bool) -> u64 {
...
+			} else if witness.len() == 5 && second_to_last.len() == 0 {
+				Some(Self::OfferedTimeout)
+			} else if witness.len() == 3 && second_to_last.len() == 0 {
+				Some(Self::AcceptedTimeout)
+			} else if witness.len() == 5 && second_to_last.len() == 32 {
+				Some(Self::AcceptedPreimage)
+			} else {
+				None
+			}
+		} else if witness_script.len() >= 137 && witness_script.len() <= 143 {
```
ariard, line 103: I think here it could be valuable to a) constify the value, actually we already do it in `functional_test_utils.rs` for those exact values and b) for each branch document the type of outputs and witness stack to ensure detection is actually correct. #1585 issue was affecting this exact code path.

lightning/src/ln/channel.rs
```diff
@@ -1466,7 +1466,12 @@ impl<Signer: Sign> Channel<Signer> {
 			($htlc: expr, $outbound: expr, $source: expr, $state_name: expr) => {
 				if $outbound == local { // "offered HTLC output"
 					let htlc_in_tx = get_htlc_in_commitment!($htlc, true);
-					if $htlc.amount_msat / 1000 >= broadcaster_dust_limit_satoshis + (feerate_per_kw as u64 * htlc_timeout_tx_weight(self.opt_anchors()) / 1000) {
+					let htlc_tx_fee = if self.opt_anchors() {
+						0
+					} else {
+						feerate_per_kw as u64 * htlc_timeout_tx_weight(false) / 1000
+					};
+					if $htlc.amount_msat / 1000 >= broadcaster_dust_limit_satoshis + htlc_tx_fee {
```
ariard, line 1474: I think `max_dust_htlc_exposure_msat` in config should be documented in consequence to notice the reduction in safety funds exposure, as the mempools fees are not to be taken account anymore.

lightning/src/ln/chan_utils.rs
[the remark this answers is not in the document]
ariard: > Wait, is the empty-checkmultisig-extra-stack-value thing consensus or just standardness? I can never remember (if its just standardness then the second-to-last-element could be malleable).

Was implemented in Core as one of the BIP62: https://github.com/bitcoin/bitcoin/blob/83434208035fda9c2b50df7b9f3ede896797cf04/src/script/interpreter.h#L62  though this one was never activated. However this rule was part of the Segwit deployment: https://github.com/bitcoin/bitcoin/blob/83434208035fda9c2b50df7b9f3ede896797cf04/src/validation.cpp#L1959

arik-so (approved):
Still LGTM :)

lightning/src/ln/chan_utils.rs
```diff
@@ -42,6 +42,11 @@ use chain;
 use util::crypto::sign;
 
 pub(crate) const MAX_HTLCS: u16 = 483;
+pub(crate) const OFFERED_HTLC_SCRIPT_WEIGHT: usize = 133;
+pub(crate) const OFFERED_HTLC_SCRIPT_WEIGHT_ANCHORS: usize = 136;
+// The weight range for `accepted_htlc_script` encompasses both the non-anchors and anchors variants.
+pub(crate) const MIN_ACCEPTED_HTLC_SCRIPT_WEIGHT: usize = 136;
```
ariard, line 48: nit: "The `accepted_htlc_script` size can vary in function of CLTV argument value"

ariard (commented):
SGTM 168e76b

Thanks for documenting `f43a7e8`, make it easier to detect wrong script detection.

dunxen (approved):

arik-so (approved):

PR #1650, opened 2022-08-04
Expose onion messages in public API 

valentinewallace:
This completes the final anti-DoS steps needed for us to be able to make the onion messages API public and integrate the `OnionMessenger` with the `PeerManager`. See commit messages for more info.

TODO: add tests for the last commit after it's conceptually ACK'd 

Based on #1604

PR #1688, opened 2022-08-31
Onion messages: flip feature bit 🎉

valentinewallace:
Based on #1650


lightning/src/ln/features.rs
```diff
@@ -952,15 +958,15 @@ mod tests {
...
+			// - onion_messages
 			// - option_channel_type | option_scid_alias
 			// - option_zeroconf
 			assert_eq!(node_features.flags.len(), 7);
 			assert_eq!(node_features.flags[0], 0b00000010);
 			assert_eq!(node_features.flags[1], 0b01010001);
 			assert_eq!(node_features.flags[2], 0b00001010);
 			assert_eq!(node_features.flags[3], 0b00001000);
-			assert_eq!(node_features.flags[4], 0b00000000);
+			assert_eq!(node_features.flags[4], 0b10000000);
```
dunxen, line 979: I think I gotta spend some time figuring out how we handle feature bits internally lol

lightning/src/ln/features.rs
```diff
@@ -548,6 +548,12 @@ impl InitFeatures {
 	pub(crate) fn to_context<C: sealed::Context>(&self) -> Features<C> {
 		self.to_context_internal()
 	}
+
+	pub(crate) fn known_without_onion_messages() -> Self {
+		let mut features = Self::known();
+		features.flags[4] &= 0b00111111;
+		features
+	}
```
jkczyz, line 556: Similarly won't need this.

lightning/src/ln/peer_handler.rs
```diff
@@ -1028,7 +1033,8 @@ impl<Descriptor: SocketDescriptor, CM: Deref, RM: Deref, OM: Deref, L: Deref, CM
 
 								peer.their_node_id = Some(their_node_id);
 								insert_node_id!();
-								let features = InitFeatures::known();
+								let features = if self.message_handler.onion_message_handler.supports_onion_message_forwarding()
+									{ InitFeatures::known() } else { InitFeatures::known_without_onion_messages() };
```
jkczyz, line 1037: Could we set a member variable for this in the constructor? Would avoid repeating oursleves.

lightning/src/ln/msgs.rs
```diff
@@ -896,6 +896,9 @@ pub trait ChannelMessageHandler : MessageSendEventsProvider {
 	// Error:
 	/// Handle an incoming error message from the given peer.
 	fn handle_error(&self, their_node_id: &PublicKey, msg: &ErrorMessage);
+
+	/// Advertise onion message forwarding support in broadcasted node announcements.
+	fn advertise_onion_message_support(&self);
```
jkczyz, line 901: I get that we're trying to only advertise the feature if we have a handler that will support it, but this seems like it should be more of a configuration parameter. Otherwise, we're shoehorning it into a trait meant for handling messages even though this only affects message broadcasting.

Trying to think if there is a better way of accomplishing this...

lightning/src/ln/peer_handler.rs
```diff
@@ -506,6 +508,7 @@ pub struct PeerManager<Descriptor: SocketDescriptor, CM: Deref, RM: Deref, OM: D
 	/// when an event process call is waiting.
 	blocked_event_processors: AtomicBool,
 	our_node_secret: SecretKey,
+	our_features: InitFeatures,
```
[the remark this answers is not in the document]
valentinewallace: Oh lol. @jkczyz do you have a preference?

lightning/src/ln/msgs.rs
jkczyz: Some questions / ideas to bounce of you both:

Do we expect the user to ever need to write their own `OnionMessageHandler`? Or will they always configure `OnionMessenger` with individual handlers for the different types of onion message payloads? If the latter, instead of using `IgnoringMessageHandler `, maybe we could have it be `Option<OnionMessenger>` and use its presence to signal if onion message forwarding is supported? Might be worth doing this regardless instead of having two levels of configuration.

For `broadcast_node_announcement`, could this also take `InitFeatures` as a parameter? I guess we'd want only `PeerManager` to determine which features to pass to it. So could have a `broadcast_node_announcement` method on `PeerManager` (without the parameter), which delegates to `ChannelManager` (with the parameter). Though that means having a somewhat unrelated `broadcast_node_announcement` on `ChannelMessageHandler` or `MessageSendEventsProvider`. Alternatively, have `BackgroundProcessor` be responsible for calling `broadcast_node_announcement` on `ChannelManager` with features obtained from `PeerManager`.

lightning/src/ln/peer_handler.rs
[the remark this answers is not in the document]
jkczyz: Possibly would depend on the resolution of https://github.com/lightningdevkit/rust-lightning/pull/1688#discussion_r963961407.

lightning/src/ln/msgs.rs
jkczyz: > I don't think so, there's plenty of things the ChannelManager is responsible for and needs to select the features of :(.

Not sure I understand here. Looks like `broadcast_node_announcement` is the only place it needs our features.

jkczyz: Hmmm... I see . Though that would also mean the features advertised by `Init` may not be accurate if a custom channel message handler were used? Seems each handler may want to declare which features they support (i.e., implement), and the `PeerManager` would want to take the intersection of these for what is actually used in either `Init` or `ChannelAnnouncement`?

jkczyz: er... s/intersection/union

valentinewallace:
Oops, mis-push. Updating

jkczyz (approved):

lightning/src/ln/features.rs
```diff
@@ -39,8 +39,13 @@
 //!     (see [BOLT-4](https://github.com/lightning/bolts/blob/master/04-onion-routing.md) for more information).
 //! - `BasicMPP` - requires/supports that a node can receive basic multi-part payments
 //!     (see [BOLT-4](https://github.com/lightning/bolts/blob/master/04-onion-routing.md#basic-multi-part-payments) for more information).
+//! - `Wumbo` - requires/supports that a node create large channels. Called `option_support_large_channel` in the spec.
```
ariard, line 42: Could say "large channels (superior to `MAX_FUNDING_SATOSHIS_NO_WUMBO`)"

lightning/src/ln/msgs.rs
```diff
@@ -980,6 +980,19 @@ pub trait OnionMessageHandler : OnionMessageProvider {
 	/// Indicates a connection to the peer failed/an existing connection was lost. Allows handlers to
 	/// drop and refuse to forward onion messages to this peer.
 	fn peer_disconnected(&self, their_node_id: &PublicKey, no_connection_possible: bool);
+
+	// Handler information:
+	/// Gets the node feature flags which this handler itself supports. All available handlers are
+	/// queried similarly and their feature flags are OR'd together to form the [`NodeFeatures`]
+	/// which are broadcasted in our node_announcement message.
```
ariard, line 987: could be [`NodeAnnouncement`]

ariard (commented):
Light Code Review ACK 43ae31b9

ariard (dismissed):

lightning/src/ln/features.rs
valentinewallace: Ah, that const isn't public

jkczyz (approved):

ariard (approved):
ACK 5d9dddd1

PR #1689, opened 2022-08-31
Expose new BumpChannelClose event for channels with anchor outputs

wpaulino:
This PR introduces a new `BumpChannelClose` event to be consumed by users indicating that they should craft a proper child anchor transaction with a sufficient feerate to increase the commitment transaction's (its parent) feerate and help get it confirmed.

Future work will also feature a similar event to bump the fees of HTLC success and timeout transactions.

### DIFF
--- a/lightning/src/ln/chan_utils.rs
+++ b/lightning/src/ln/chan_utils.rs
@@ -736,6 +736,14 @@ pub(crate) fn get_anchor_output<'a>(commitment_tx: &'a Transaction, funding_pubk
 		.map(|(idx, txout)| (idx as u32, txout))
 }
 
+/// Returns the witness required to satisfy and spend an anchor input.
+pub fn build_anchor_input_witness(funding_key: &PublicKey, funding_sig: &Signature) -> Witness {
+	let anchor_redeem_script = chan_utils::get_anchor_redeemscript(funding_key);
+	let mut funding_sig = funding_sig.serialize_der().to_vec();
+	funding_sig.push(EcdsaSighashType::All as u8);
+	Witness::from_vec(vec![funding_sig, anchor_redeem_script.to_bytes()])
+}
+
 /// Per-channel data used to build transactions in conjunction with the per-commitment data (CommitmentTransaction).
 /// The fields are organized by holder/counterparty.
 ///


lightning/src/util/events.rs
```diff
@@ -602,6 +620,41 @@ pub enum Event {
 		/// Destination of the HTLC that failed to be processed.
 		failed_next_destination: HTLCDestination,
 	},
+	/// Indicates that a channel featuring anchors outputs is to be closed by broadcasting the local
+	/// commitment transaction. Since commitment transactions have a static feerate pre-agreed upon,
+	/// they may need additional fees to be attached through a child transaction using the popular
+	/// Child-Pays-For-Parent (CPFP) fee bumping technique. This child transaction must include the
```
ariard, line 626: can be linked directly: https://bitcoinops.org/en/topics/cpfp/

lightning/src/util/events.rs
```diff
@@ -602,6 +620,41 @@ pub enum Event {
...
+	/// anchor input found within `anchor_descriptor` along with additional inputs to meet the
+	/// target feerate. Once the transaction is constructed, it must be fully signed for and
```
ariard, line 628: "Failure to meet the target feerate decreases the confirmation odds of this transaction detected as time-sensitive, this could lead to loss of funds" Like to make clearer meeting the target is a MUST?

lightning/src/util/events.rs
```diff
@@ -602,6 +620,41 @@ pub enum Event {
...
+	///
+	/// The consumer should be able to sign for any of the additional inputs included within the
+	/// child anchor transaction. To sign its anchor input, a [`InMemorySigner`] should be
+	/// re-derived through [`KeysManager::derive_channel_keys`] with the help of
+	/// [`AnchorDescriptor::channel_keys_id`] and [`AnchorDescriptor::channel_value_satoshis`].
+	///
+	/// It is possible to receive more than an instance of this event if a valid child anchor
+	/// transaction is never broadcast or is but not with a sufficient fee to be mined. Care should
+	/// be taken by the consumer of the event to ensure any future iterations of the child anchor
+	/// transaction adhere to the Replace-By-Fee (RBF) rules for fee bumps to be accepted into the
```
ariard, line 641: can be linked: https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md

lightning/src/chain/channelmonitor.rs
```diff
@@ -2216,6 +2240,27 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
 					self.lockdown_from_offchain = true;
 					if *should_broadcast {
 						self.broadcast_latest_holder_commitment_txn(broadcaster, logger);
+						// If the channel supports anchor outputs, we'll need to emit an external
+						// event to be consumed such that a child transaction is broadcast with a
+						// high enough feerate for the parent commitment transaction to confirm.
+						if self.onchain_tx_handler.opt_anchors() {
```
ariard, line 2246: Note, iirc `should_broadcast` doesn't mean there are effectively pending HTLCs making the confirmation of this transaction an emergency. In fact, even if we force-close, we could still be satisfied by a slow confirmation and save on fees ? Maybe `ChannelForceClosed` should be enriched with a `bool pending_htlcs` argument

lightning/src/chain/onchaintx.rs
```diff
@@ -162,6 +164,19 @@ impl Writeable for Option<Vec<Option<(usize, Signature)>>> {
 	}
 }
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) enum ClaimEvent {
+	BumpCommitment {
+		target_feerate_sat_per_1000_weight: u32,
+		commitment_tx: Transaction,
+		anchor_output_idx: u32,
```
ariard, line 172: The event could return the expiration timelocks and the value of all HTLCs if the user would like to be more aggressive in term of fee-bumping than our default suggestion? Or maybe we could expose some internal setting like a fee-bumping frequency `low,medium,high`. Ultimately it's a choice between fee-savings and risk of non-confirmation.

lightning/src/chain/channelmonitor.rs
ariard: In fact, ideally we would return back the amount of the HTLCs at stake, and the user would commit the fee-bumping values they can afford in function of their appreciation of the risk loss. IIRC, LND does something similar, or at least consider the values of the HTLC before to decide on an unilateral closing. I think this is a type of advanced APIs that could be interesting to offer to our users in the future.

In the meanwhile I think it could be good to track follow-ups in an issue.

ariard: Ah I think we have a confusion here between the `ChannelForceClosed` flag `should_broadcast` and the function `should_broadcast_holder_commitment_txn` called in `block_confirmed`? For the latter, I agree if there is no pending HTLC we''ll never end up broadcasting through this flow. For the former I think the slow confirmation semantic can be captured by the new `pending_htlcs`. I think comment can be updated to indicate that if `pending_htlcs.is_empty()`, confirmation is not time-sensitive.

lightning/src/util/events.rs
```diff
@@ -602,6 +620,41 @@ pub enum Event {
...
+	/// [`InMemorySigner`]: crate::chain::keysinterface::InMemorySigner
+	/// [`KeysManager::derive_channel_keys`]: crate::chain::keysinterface::KeysManager::derive_channel_keys
+	BumpChannelClose {
+		/// The target feerate to use for the child transaction spending the anchor output of the
+		/// channel's commitment transaction to broadcast and bump the fee of.
+		target_feerate_sat_per_1000_weight: u32,
+		/// The channel's commitment transaction to bump the fee of. This transaction should be
+		/// broadcast along with the anchor transaction constructed as a result of consuming this
+		/// event.
+		commitment_tx: Transaction,
```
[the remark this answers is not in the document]
ariard: Yeah definitively a good call to have a generic `BumpTransaction` to avoid introducing in the future new events for splicing/dual-funding fee-bumping!

ariard (commented):
I'm Approach ACK on both the `BumpTransaction` enum, the data scope returned back to the user and the event itself. To be looked more, the callsites where we're triggering the event, to see if there is way to simplify or clean paths.

lightning/src/chain/package.rs
```diff
@@ -583,7 +616,7 @@ impl PackageTemplate {
 	}
 	/// Gets the amount of all outptus being spent by this package, only valid for malleable
```
arik-so, line 617: outputs*

lightning/src/chain/package.rs
```diff
@@ -594,61 +627,71 @@ impl PackageTemplate {
...
+			});
+		}
+		for (i, (outpoint, out)) in self.inputs.iter().enumerate() {
+			log_debug!(logger, "Adding claiming input for outpoint {}:{}", outpoint.txid, outpoint.vout);
+			if !out.finalize_input(&mut bumped_tx, i, onchain_handler) { return None; }
 		}
+		log_debug!(logger, "Finalized transaction {} ready to broadcast", bumped_tx.txid());
+		Some(bumped_tx)
+	}
+	pub(crate) fn finalize_untractable_package<L: Deref, Signer: Sign>(
```
arik-so, line 683: I think "intractable" is the more common form

lightning/src/chain/package.rs
```diff
@@ -594,61 +627,71 @@ impl PackageTemplate {
...
-				} else { panic!("API Error: Package must not be inputs empty"); }
+				debug_assert!(destination_scripts.len() > 1);
+				// value: 8 bytes ; var_int: 1 byte ; pk_script: `pk_script.len()`
+				destination_scripts.iter()
+					.fold(0, |acc, script| acc + (8 + 1 + script.len()) * WITNESS_SCALE_FACTOR)
 			},
+		};
+		transaction_weight + inputs_weight + witnesses_weight + output_weight
+	}
+	pub(crate) fn finalize_malleable_package<L: Deref, Signer: Sign>(
```
arik-so, line 655: A comment specifying when malleable vs. not should be used might be helpful

lightning/src/util/events.rs
```diff
@@ -753,6 +844,15 @@ impl Writeable for Event {
 					(2, failed_next_destination, required),
 				})
 			},
+			&Event::BumpTransaction(ref event)=> {
+				27u8.write(writer)?;
+				match event {
+					// We never write the ChannelClose events as they'll be replayed upon restarting
+					// anyway if the commitment transaction remains
+					// unconfirmed.
```
dunxen, line 852: nit: Can keep this on same line

lightning/src/util/events.rs
```diff
@@ -196,6 +196,94 @@ impl_writeable_tlv_based_enum_upgradable!(HTLCDestination,
...
+	/// by its originator.
+	pub expiry: u32,
+	/// The index of the pending HTLC's output in the commitment transaction.
+	pub output_index: u32,
+}
+
+/// Represents the different types of transactions, originating from LDK, to be bumped.
+#[derive(Clone, Debug)]
+pub enum BumpTransaction {
+	/// Indicates that a channel featuring anchors outputs is to be closed by broadcasting the local
```
dunxen, line 235:
```suggestion
	/// Indicates that a channel featuring anchor outputs is to be closed by broadcasting the local
```

lightning/src/util/events.rs
```diff
@@ -196,6 +196,94 @@ impl_writeable_tlv_based_enum_upgradable!(HTLCDestination,
...
+	/// is constructed, it must be fully signed for and broadcasted by the consumer of the event
+	/// along with the `commitment_tx` enclosed. Note that the `commitment_tx` must always be
+	/// broadcast first, as the child anchor transaction depends on it.
+	///
+	/// The consumer should be able to sign for any of the additional inputs included within the
+	/// child anchor transaction. To sign its anchor input, an [`InMemorySigner`] should be
+	/// re-derived through [`KeysManager::derive_channel_keys`] with the help of
+	/// [`AnchorDescriptor::channel_keys_id`] and [`AnchorDescriptor::channel_value_satoshis`].
+	///
+	/// It is possible to receive more than an instance of this event if a valid child anchor
```
dunxen, line 252: nit: maybe this is a bit clearer

```suggestion
	/// It is possible to receive more than one instance of this event if a valid child anchor
```

lightning/src/util/events.rs
```diff
@@ -196,6 +196,94 @@ impl_writeable_tlv_based_enum_upgradable!(HTLCDestination,
...
+	/// The transaction input's outpoint corresponding to the commitment transaction's anchor
+	/// output.
+	pub outpoint: OutPoint,
+}
+
+/// Describes a pending HTLC in a commitment transaction.
+#[derive(Clone, Debug)]
+pub struct CommitmentPendingHTLC {
+	/// Whether the pending HTLC originated from LDK.
+	pub outbound: bool,
```
dunxen, line 222: Does `outbound` here provide more clarity than just using `offered`? I'm not sure of the context of use.

lightning/src/chain/package.rs
```diff
@@ -686,14 +717,42 @@ impl PackageTemplate {
...
+	pub(crate) fn compute_package_feerate<F: Deref>(
+		&self, fee_estimator: &LowerBoundedFeeEstimator<F>, conf_target: ConfirmationTarget,
+	) -> u32 where F::Target: FeeEstimator {
+		let feerate_estimate = fee_estimator.bounded_sat_per_1000_weight(conf_target);
+		if self.feerate_previous != 0 {
+			// If old feerate inferior to actual one given back by Fee Estimator, use it to compute new fee...
+			if feerate_estimate as u64 > self.feerate_previous {
+				feerate_estimate
+			} else {
+				// ...else just increase the previous feerate by 25% (because that's a nice number)
```
dunxen, line 733: > (because that's a nice number)

Just something that works okay or does another implementation do this? :sweat_smile:

ariard: There is no BOLT recommendation on a compelling fee-bumping strategy for time-sensitive closes. Ideally, we would rely on a smart fee-estimator telling us the needed feerate increase to stay in top block of network mempools, though that type of fee-estimation doesn't exist yet.

lightning/src/chain/channelmonitor.rs
ariard: Yes, I think this is okay that `OnchainTxHandler` returns a reasonable default fee bumping strategy, based on `ConfirmationTarget::HighPriority`. Indeed, the event consumer could choose to ignore the feerate and employ their own fee bumping strategy based on historical network mempool data and loss aversion of pending HTLCs. 

> when confirmation is not even an urgency (in the case of an empty commitment).

Sounds fee bumping could be turned off in case of empty commitment. Or at least an option, if the user would like to get back commitment balance liquidity quickly.

lightning/src/chain/onchaintx.rs
```diff
@@ -388,23 +417,62 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
-			// Note: Currently, amounts of holder outputs spending witnesses aren't used
-			// as we can't malleate spending package to increase their feerate. This
-			// should change with the remaining anchor output patchset.
-			if let Some(transaction) = cached_request.finalize_package(self, 0, self.destination_script.clone(), logger) {
-				return Some((None, 0, transaction));
+			// Untractable packages cannot have their fees bumped through Replace-By-Fee. Some
+			// packages may support fee bumping through Child-Pays-For-Parent, indicated by those
+			// which require external funding.
+			let inputs = cached_request.inputs();
+			debug_assert_eq!(inputs.len(), 1);
```
ariard, line 435: I know this is a debug assert, but I do have a doubt in the future. Some untractable packages today (`PackageMalleability::Untractable`) like `HolderHTLCOutput` become malleable post-anchor thanks to sighash_single. I think that's okay as we don't touch the underlying package API right now, though we might hit ourselves in the future if we enable aggregation of second-stage HTLC transactions.

lightning/src/chain/package.rs
```diff
@@ -686,14 +717,42 @@ impl PackageTemplate {
...
+				(self.feerate_previous + (self.feerate_previous / 4)).try_into().unwrap()
+			}
+		} else {
+			feerate_estimate
+		}
+	}
+
+	pub(crate) fn requires_external_funding(&self) -> bool {
```
ariard, line 741: nice if documented even if minor

lightning/src/chain/onchaintx.rs
```diff
@@ -388,23 +418,68 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
+								new_timer,
+								target_feerate_sat_per_1000_weight as u64,
+								OnchainClaim::Event(ClaimEvent::BumpCommitment {
+									target_feerate_sat_per_1000_weight,
+									commitment_tx: tx.clone(),
+									anchor_output_idx: idx,
+								}),
+							))
+						},
+						// An anchor output was not found. There's nothing we can do other than
```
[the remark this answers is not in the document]
ariard: Essentially it means our local commitment transaction doesn't include HTLCs outputs or a meaningful balance output and we can shrug about confirmation.

lightning/src/chain/onchaintx.rs
```diff
@@ -603,9 +683,16 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
-				log_info!(logger, "Broadcasting RBF-bumped onchain {}", log_tx!(bump_tx));
-				broadcaster.broadcast_transaction(&bump_tx);
+			if let Some((new_timer, new_feerate, bump_claim)) = self.generate_claim(cur_height, &request, &*fee_estimator, &*logger) {
+				match bump_claim {
+					OnchainClaim::Tx(bump_tx) => {
+						log_info!(logger, "Broadcasting RBF-bumped onchain {}", log_tx!(bump_tx));
+						broadcaster.broadcast_transaction(&bump_tx);
+					},
+					OnchainClaim::Event(claim_event) => {
+						self.pending_claim_events.insert(*first_claim_txid, claim_event);
```
ariard, line 693: Could add a `log_info` as there is one for `OnchainClaim::Tx`

ariard (commented):
Still have to review some of the changes in `package.rs`, looks good overall.

Thanks for dealing with the rather murky package API, my apologies for the lack of documentation there.

lightning/src/chain/onchaintx.rs
```diff
@@ -388,23 +427,71 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
+									commitment_tx: tx.clone(),
+									anchor_output_idx: idx,
+								}),
+							))
+						},
+						// An anchor output was not found. There's nothing we can do other than
+						// attempt to broadcast the transaction with its current fee rate and hope
+						// it confirms. This is essentially the same behavior as a commitment
+						// transaction without anchor outputs.
+						None => Some((None, 0, OnchainClaim::Tx(tx.clone()))),
```
ariard, line 487: IIRC, the case where we don't find a second anchor output for the channel is when we have no fund at stakes in the channel (no-htlc, no-balance) output. We could consider the channel early closed and return a `OnchainEvent::ChannelClosed`, to allow consumer to react in consequence.

lightning/src/chain/package.rs
```diff
@@ -594,61 +626,60 @@ impl PackageTemplate {
...
-					if let Some(final_tx) = outp.get_finalized_tx(outpoint, onchain_handler) {
-						log_debug!(logger, "Adding claiming input for outpoint {}:{}", outpoint.txid, outpoint.vout);
-						log_debug!(logger, "Finalized transaction {} ready to broadcast", final_tx.txid());
-						return Some(final_tx);
-					}
-					return None;
-				} else { panic!("API Error: Package must not be inputs empty"); }
-			},
+	pub(crate) fn finalize_malleable_package<L: Deref, Signer: Sign>(
+		&self, onchain_handler: &mut OnchainTxHandler<Signer>, value: u64, destination_script: Script, logger: &L,
```
ariard, line 644: useless comma at the end?

lightning/src/chain/channelmonitor.rs
```diff
@@ -2299,6 +2321,39 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
 	pub fn get_and_clear_pending_events(&mut self) -> Vec<Event> {
 		let mut ret = Vec::new();
 		mem::swap(&mut ret, &mut self.pending_events);
+		#[cfg(test)] // TODO: remove-on-anchors-release
```
ariard, line 2324: I wonder if this best practice to abuse Rust compilation flags to feature gate internal LDK feature into completion. I would rather add a new config option `option_anchor_output_v0` and turns it to false by default. It should allow you to exercise new code with our current test framework without us having to wonder if we have not forget a random `cfg(test)` nullifying some critical code paths when the feature is complete. At the same time it provides a modular way for beta users to play with feature. 

What do we think ?

lightning/src/chain/channelmonitor.rs
```diff
@@ -2299,6 +2321,39 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
 	pub fn get_and_clear_pending_events(&mut self) -> Vec<Event> {
```
ariard, line 2321: I think we should update `ChainMonitor::process_pending_events` in consequence, noticing the user that this call is from now on important for the funds safety and call frequence should at least match every block ?

ariard (commented):
I think we're good though I'm blocking a bit as using `#[cfg(test)] TODO: remove-on-anchors-release` to feature gate anchor output support. I can't remember we have gradually release sophisticated feature in the past, so sounds a first time. While I think it's worthy to do so in the present case I don't know if it's the best approach, see comment.

lightning/src/chain/channelmonitor.rs
```diff
@@ -2299,6 +2321,39 @@ impl<Signer: Sign> ChannelMonitorImpl<Signer> {
...
 		mem::swap(&mut ret, &mut self.pending_events);
+		#[cfg(test)] // TODO: remove-on-anchors-release
+		for claim_event in self.onchain_tx_handler.get_and_clear_pending_claim_events().drain(..) {
+			match claim_event {
+				ClaimEvent::BumpCommitment {
+					package_target_feerate_sat_per_1000_weight, commitment_tx, anchor_output_idx,
+				} => {
+					let commitment_txid = commitment_tx.txid();
+					debug_assert_eq!(self.current_holder_commitment_tx.txid, commitment_txid);
+					let pending_htlcs = self.current_holder_commitment_tx.htlc_outputs.iter()
```
arik-so, line 2332: this seems like something quite useful that perhaps ought to be defined in an easily callable function, either here or on a commitment_tx

lightning/src/chain/keysinterface.rs
```diff
@@ -645,6 +646,25 @@ impl InMemorySigner {
 		witness.push(witness_script.clone().into_bytes());
 		Ok(witness)
 	}
+
+	/// Computes the fully-signed witness for a commitment transaction's anchor output used as an
+	/// input within `anchor_tx`, which spends the commitment transaction, at index `input`.
+	///
+	/// Panics if `anchor_tx`'s sighash cannot be computed or index `input` is not valid within
+	/// `anchor_tx`.
+	pub fn sign_holder_anchor_output(
```
arik-so, line 655: isn't it technically the anchor input?

lightning/src/chain/onchaintx.rs
```diff
@@ -388,23 +427,71 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
+					let funding_pubkey = &self.channel_transaction_parameters.holder_pubkeys.funding_pubkey;
+					match chan_utils::get_anchor_output(&tx, funding_pubkey) {
+						// An anchor output was found, so we should yield a funding event externally.
+						Some((idx, _)) => {
+							// Our target feerate will depend on whether we have any HTLCs present
+							// within our commitment.
+							let conf_target = if self.holder_commitment.trust().htlcs().is_empty() {
+								ConfirmationTarget::Background
+							} else {
+								ConfirmationTarget::HighPriority
```
arik-so, line 469: is this strategy effective? Should we entertain the option of checking the HTLC and if, say, it's a preimage spend that's expiring soon, use a high priority, whereas if there's still a bunch of time left to expiration, perhaps use medium priority?

arik-so (approved):

lightning/src/chain/onchaintx.rs
```diff
@@ -667,12 +788,20 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
-				log_info!(logger, "Broadcasting onchain {}", log_tx!(bump_tx));
-				broadcaster.broadcast_transaction(&bump_tx);
+				match bump_claim {
+					OnchainClaim::Tx(bump_tx) => {
+						log_info!(logger, "Broadcasting onchain {}", log_tx!(bump_tx));
+						broadcaster.broadcast_transaction(&bump_tx);
+					},
+					#[cfg(feature = "anchors")]
+					OnchainClaim::Event(claim_event) => {
+						self.pending_claim_events.insert(_first_claim_txid_height.0, claim_event);
```
ariard, line 802: Could add a `log_info`.

ariard (commented):
ACK e159214

ariard (commented):
ACK c963d3a

arik-so (dismissed):

arik-so (approved):

PR #1690, opened 2022-09-02
Bump bitcoin from 0.29.0 (now yanked) to 0.29.1

arik-so:


Kixunil:
This shouldn't be needed, I think. We only removed accidentally published macros.

tnull:
> This shouldn't be needed, I think. We only removed accidentally published macros.

Then again, is there a reason not to do it?

Kixunil:
There is: less noise in git if you don't. But I'm not a maintainer here so it's up to you to decide what's better.

dunxen:
> There is: less noise in git if you don't. But I'm not a maintainer here so it's up to you to decide what's better.

I can understand the noise perspective 😄, but the change seems pretty isolated to a few lines and shouldn't affect git blame-ing as much as tons of formatting or changes like that.

I also find not depending on a yanked version more sanitary (even if it was just removed because of macro visibility). And don't think we've ever had pushback about just keeping the crate up to date.

But yeah, maintainers' call! 🙂

Kixunil:
> there's no changes

If you mean no changes that affect the `lightning` crate your whole comment is entirely correct. :)

PR #1694, opened 2022-09-05
Move `LockableScore` requirement away from `Router` trait

jurvis:
## Purpose
A scorer is not necessary for users who may be using a different server from their payer node for routing. This PR will allow them to implement a `Router` without it. 

## Implementation
* Move `Score` type parameterization from `InvoicePayer` and `Router` to
`DefaultRouter`
* Adds a new field, `scorer`, to `DefaultRouter`
* Move `AccountsForInflightHtlcs` to `DefaultRouter`, which we
will use to wrap the new `scorer` field, so scoring only happens in
`DefaultRouter` explicitly.
* Add scoring related functions to `Router` trait that we used to call
directly from `InvoicePayer`.
* Instead of parameterizing `scorer` in `find_route`, we replace it with
inflight_map so `InvoicePayer` can pass on information about inflight
HTLCs to the router.

### DIFF
--- a/lightning-invoice/src/payment.rs
+++ b/lightning-invoice/src/payment.rs
@@ -15,9 +15,10 @@
 //! and payee using information provided by the payer and from the payee's [`Invoice`], when
 //! applicable.
 //!
-//! [`InvoicePayer`] is parameterized by a [`LockableScore`], which it uses for scoring failed and
-//! successful payment paths upon receiving [`Event::PaymentPathFailed`] and
-//! [`Event::PaymentPathSuccessful`] events, respectively.
+//! [`InvoicePayer`] uses its [`Router`] parameterization for optionally notifying scorers upon
+//! receiving the [`Event::PaymentPathFailed`] and [`Event::PaymentPathSuccessful`] events.
+//! It also does the same for payment probe failure and success events using [`Event::ProbeFailed`]
+//! and [`Event::ProbeSuccessful`].
 //!
 //! [`InvoicePayer`] is capable of retrying failed payments. It accomplishes this by implementing
 //! [`EventHandler`] which decorates a user-provided handler. It will intercept any
@@ -32,9 +33,7 @@
 //! # extern crate lightning_invoice;
 //! # extern crate secp256k1;
 //! #
-//! # #[cfg(feature = "no-std")]
-//! # extern crate core2;
-//! #
+//! # use lightning::io;
 //! # use lightning::ln::{PaymentHash, PaymentPreimage, PaymentSecret};
 //! # use lightning::ln::channelmanager::{ChannelDetails, PaymentId, PaymentSendFailure};
 //! # use lightning::ln::msgs::LightningError;
@@ -45,15 +44,10 @@
 //! # use lightning::util::logger::{Logger, Record};
 //! # use lightning::util::ser::{Writeable, Writer};
 //! # use lightning_invoice::Invoice;
-//! # use lightning_invoice::payment::{InvoicePayer, Payer, Retry, Router};
+//! # use lightning_invoice::payment::{InFlightHtlcs, InvoicePayer, Payer, Retry, Router};
 //! # use secp256k1::PublicKey;
 //! # use std::cell::RefCell;
 //! # use std::ops::Deref;
-//! #
-//! # #[cfg(not(feature = "std"))]
-//! # use core2::io;
-//! # #[cfg(feature = "std")]
-//! # use std::io;
 //! #
 //! # struct FakeEventProvider {}
 //! # impl EventsProvider for FakeEventProvider {
@@ -78,10 +72,15 @@
 //! #
 //! # struct FakeRouter {}
 //! # impl Router for FakeRouter {
-//! #     fn find_route<S: Score>(
+//! #     fn find_route(
 //! #         &self, payer: &PublicKey, params: &RouteParameters, payment_hash: &PaymentHash,
-//! #         first_hops: Option<&[&ChannelDetails]>, scorer: &S
+//! #         first_hops: Option<&[&ChannelDetails]>, _inflight_htlcs: InFlightHtlcs
 //! #     ) -> Result<Route, LightningError> { unimplemented!() }
+//! #
+//! #     fn notify_payment_path_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64) {  unimplemented!() }
+//! #     fn notify_payment_path_successful(&self, path: Vec<&RouteHop>) {  unimplemented!() }
+//! #     fn notify_payment_probe_successful(&self, path: Vec<&RouteHop>) {  unimplemented!() }
+//! #     fn notify_payment_probe_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64) { unimplemented!() }
 //! # }
 //! #
 //! # struct FakeScorer {}
@@ -115,7 +114,7 @@
 //! # let router = FakeRouter {};
 //! # let scorer = RefCell::new(FakeScorer {});
 //! # let logger = FakeLogger {};
-//! let invoice_payer = InvoicePayer::new(&payer, router, &scorer, &logger, event_handler, Retry::Attempts(2));
+//! let invoice_payer = InvoicePayer::new(&payer, router, &logger, event_handler, Retry::Attempts(2));
 //!
 //! let invoice = "...";
 //! if let Ok(invoice) = invoice.parse::<Invoice>() {
@@ -141,11 +140,11 @@ use bitcoin_hashes::Hash;
 use bitcoin_hashes::sha256::Hash as Sha256;
 
 use crate::prelude::*;
+use lightning::io;
 use lightning::ln::{PaymentHash, PaymentPreimage, PaymentSecret};
 use lightning::ln::channelmanager::{ChannelDetails, PaymentId, PaymentSendFailure};
 use lightning::ln::msgs::LightningError;
 use lightning::routing::gossip::NodeId;
-use lightning::routing::scoring::{ChannelUsage, LockableScore, Score};
 use lightning::routing::router::{PaymentParameters, Route, RouteHop, RouteParameters};
 use lightning::util::errors::APIError;
 use lightning::util::events::{Event, EventHandler};
@@ -167,7 +166,7 @@ use std::time::SystemTime;
 /// See [module-level documentation] for details.
 ///
 /// [module-level documentation]: crate::payment
-pub type InvoicePayer<P, R, S, L, E> = InvoicePayerUsingTime::<P, R, S, L, E, ConfiguredTime>;
+pub type InvoicePayer<P, R, L, E> = InvoicePayerUsingTime::<P, R, L, E, ConfiguredTime>;
 
 #[cfg(not(feature = "no-std"))]
 type ConfiguredTime = std::time::Instant;
@@ -177,15 +176,13 @@ use time_utils;
 type ConfiguredTime = time_utils::Eternity;
 
 /// (C-not exported) generally all users should use the [`InvoicePayer`] type alias.
-pub struct InvoicePayerUsingTime<P: Deref, R: Router, S: Deref, L: Deref, E: EventHandler, T: Time>
+pub struct InvoicePayerUsingTime<P: Deref, R: Router, L: Deref, E: EventHandler, T: Time>
 where
 	P::Target: Payer,
-	S::Target: for <'a> LockableScore<'a>,
 	L::Target: Logger,
 {
 	payer: P,
 	router: R,
-	scorer: S,
 	logger: L,
 	event_handler: E,
 	/// Caches the overall attempts at making a payment, which is updated prior to retrying.
@@ -206,49 +203,6 @@ impl<T: Time> PaymentInfo<T> {
 			attempts: PaymentAttempts::new(),
 			paths: vec![],
 		}
-	}
-}
-
-/// Used to store information about all the HTLCs that are inflight across all payment attempts
-struct AccountForInFlightHtlcs<'a, S: Score> {
-	scorer: &'a mut S,
-	/// Maps a channel's short channel id and its direction to the liquidity used up.
-	inflight_htlcs: HashMap<(u64, bool), u64>,
-}
-
-#[cfg(c_bindings)]
-impl<'a, S:Score> lightning::util::ser::Writeable for AccountForInFlightHtlcs<'a, S> {
-	fn write<W: lightning::util::ser::Writer>(&self, writer: &mut W) -> Result<(), std::io::Error> { self.scorer.write(writer) }
-}
-
-impl<'a, S: Score> Score for AccountForInFlightHtlcs<'a, S> {
-	fn channel_penalty_msat(&self, short_channel_id: u64, source: &NodeId, target: &NodeId, usage: ChannelUsage) -> u64 {
-		if let Some(used_liqudity) = self.inflight_htlcs.get(&(short_channel_id, source < target)) {
-			let usage = ChannelUsage {
-				inflight_htlc_msat: usage.inflight_htlc_msat + used_liqudity,
-				..usage
-			};
-
-			self.scorer.channel_penalty_msat(short_channel_id, source, target, usage)
-		} else {
-			self.scorer.channel_penalty_msat(short_channel_id, source, target, usage)
-		}
-	}
-
-	fn payment_path_failed(&mut self, path: &[&RouteHop], short_channel_id: u64) {
-		self.scorer.payment_path_failed(path, short_channel_id)
-	}
-
-	fn payment_path_successful(&mut self, path: &[&RouteHop]) {
-		self.scorer.payment_path_successful(path)
-	}
-
-	fn probe_failed(&mut self, path: &[&RouteHop], short_channel_id: u64) {
-		self.scorer.probe_failed(path, short_channel_id)
-	}
-
-	fn probe_successful(&mut self, path: &[&RouteHop]) {
-		self.scorer.probe_successful(path)
 	}
 }
 
@@ -314,10 +268,18 @@ pub trait Payer {
 /// A trait defining behavior for routing an [`Invoice`] payment.
 pub trait Router {
 	/// Finds a [`Route`] between `payer` and `payee` for a payment with the given values.
-	fn find_route<S: Score>(
+	fn find_route(
 		&self, payer: &PublicKey, route_params: &RouteParameters, payment_hash: &PaymentHash,
-		first_hops: Option<&[&ChannelDetails]>, scorer: &S
+		first_hops: Option<&[&ChannelDetails]>, inflight_htlcs: InFlightHtlcs
 	) -> Result<Route, LightningError>;
+	/// Lets the router know that payment through a specific path has failed.
+	fn notify_payment_path_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64);
+	/// Lets the router know that payment through a specific path was successful.
+	fn notify_payment_path_successful(&self, path: Vec<&RouteHop>);
+	/// Lets the router know that a payment probe was successful.
+	fn notify_payment_probe_successful(&self, path: Vec<&RouteHop>);
+	/// Lets the router know that a payment probe failed.
+	fn notify_payment_probe_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64);
 }
 
 /// Strategies available to retry payment path failures for an [`Invoice`].
@@ -359,10 +321,9 @@ pub enum PaymentError {
 	Sending(PaymentSendFailure),
 }
 
-impl<P: Deref, R: Router, S: Deref, L: Deref, E: EventHandler, T: Time> InvoicePayerUsingTime<P, R, S, L, E, T>
+impl<P: Deref, R: Router, L: Deref, E: EventHandler, T: Time> InvoicePayerUsingTime<P, R, L, E, T>
 where
 	P::Target: Payer,
-	S::Target: for <'a> LockableScore<'a>,
 	L::Target: Logger,
 {
 	/// Creates an invoice payer that retries failed payment paths.
@@ -370,12 +331,11 @@ where
 	/// Will forward any [`Event::PaymentPathFailed`] events to the decorated `event_handler` once
 	/// `retry` has been exceeded for a given [`Invoice`].
 	pub fn new(
-		payer: P, router: R, scorer: S, logger: L, event_handler: E, retry: Retry
+		payer: P, router: R, logger: L, event_handler: E, retry: Retry
 	) -> Self {
 		Self {
 			payer,
 			router,
-			scorer,
 			logger,
 			event_handler,
 			payment_cache: Mutex::new(HashMap::new()),
@@ -487,7 +447,7 @@ where
 		let inflight_htlcs = self.create_inflight_map();
 		let route = self.router.find_route(
 			&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
-			&AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs }
+			inflight_htlcs
 		).map_err(|e| PaymentError::Routing(e))?;
 
 		match send_payment(&route) {
@@ -592,7 +552,7 @@ where
 
 		let route = self.router.find_route(
 			&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
-			&AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs }
+			inflight_htlcs
 		);
 
 		if route.is_err() {
@@ -651,7 +611,7 @@ where
 	///
 	/// This function should be called whenever we need information about currently used up liquidity
 	/// across payments.
-	fn create_inflight_map(&self) -> HashMap<(u64, bool), u64> {
+	fn create_inflight_map(&self) -> InFlightHtlcs {
 		let mut total_inflight_map: HashMap<(u64, bool), u64> = HashMap::new();
 		// Make an attempt at finding existing payment information from `payment_cache`. If it
 		// does not exist, it probably is a fresh payment and we can just return an empty
@@ -683,7 +643,7 @@ where
 			}
 		}
 
-		total_inflight_map
+		InFlightHtlcs(total_inflight_map)
 	}
 }
 
@@ -698,10 +658,9 @@ fn has_expired(route_params: &RouteParameters) -> bool {
 	} else { false }
 }
 
-impl<P: Deref, R: Router, S: Deref, L: Deref, E: EventHandler, T: Time> EventHandler for InvoicePayerUsingTime<P, R, S, L, E, T>
+impl<P: Deref, R: Router, L: Deref, E: EventHandler, T: Time> EventHandler for InvoicePayerUsingTime<P, R, L, E, T>
 where
 	P::Target: Payer,
-	S::Target: for <'a> LockableScore<'a>,
 	L::Target: Logger,
 {
 	fn handle_event(&self, event: &Event) {
@@ -721,7 +680,7 @@ where
 			} => {
 				if let Some(short_channel_id) = short_channel_id {
 					let path = path.iter().collect::<Vec<_>>();
-					self.scorer.lock().payment_path_failed(&path, *short_channel_id);
+					self.router.notify_payment_path_failed(path, *short_channel_id)
 				}
 
 				if payment_id.is_none() {
@@ -744,7 +703,7 @@ where
 			},
 			Event::PaymentPathSuccessful { path, .. } => {
 				let path = path.iter().collect::<Vec<_>>();
-				self.scorer.lock().payment_path_successful(&path);
+				self.router.notify_payment_path_successful(path);
 			},
 			Event::PaymentSent { payment_hash, .. } => {
 				let mut payment_cache = self.payment_cache.lock().unwrap();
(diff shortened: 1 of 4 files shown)


jurvis:
I haven't updated the test suite yet, but I wanted to seek concept ACKs on my approach. I'm not 100% sure we want to parameterize `inflight_htlcs_map` in `find_route`, in particular.

jkczyz:
> Generally concept ack. Its kinda annoying to make `DefaultRouter` also an `EventHandler` but only optionally, though - it means we have to pass both the router and the InvoicePayer to the BackgroundProcessor to pipe events through. As an alternative, we could make the `Router` trait also require `EventHandler`, which means we could then call through to the Router's event handling directly in `InvoicePayer`. Its a bit awkward but at least less code for the user, dunno how @jkczyz would feel about that, though.

My thinking had been we would expand the `Router` trait with methods for handling success / failure rather than having `DefaultRouter` implement `EventHandler`. Someone may have a `RemoteRouter` implementation that should still be notified and handle accordingly (e.g., by sending an RPC to the remote server to update the remote scorer).

jurvis:
> `Router` trait with methods for handling success / failure rather than having `DefaultRouter` implement `EventHandler`
 
 ah, got it. makes sense!

jurvis:
Used a different approach in 2032f46 -- instead of implementing `EventHandler` in `DefaultRouter`, have payer call `payment_path_successful`, `payment_path_failed`, `probe_success`, and `probe_failure` as newly defined function in the `Router` trait.

jurvis:
Docs need updating -- I'll look into that in a future commit

lightning-invoice/src/payment.rs
```diff
@@ -15,9 +15,9 @@
 //! and payee using information provided by the payer and from the payee's [`Invoice`], when
 //! applicable.
 //!
-//! [`InvoicePayer`] is parameterized by a [`LockableScore`], which it uses for scoring failed and
-//! successful payment paths upon receiving [`Event::PaymentPathFailed`] and
-//! [`Event::PaymentPathSuccessful`] events, respectively.
+//! [`InvoicePayer`] uses its [`Router`] parameterization for scoring failed and successful payment
+//! paths upon receiving [`Event::PaymentPathFailed`] and [`Event::PaymentPathSuccessful`] events,
```
tnull, line 19: Maybe update to mention the probing events also while you're here?

lightning-invoice/src/payment.rs
```diff
@@ -314,10 +275,18 @@ pub trait Payer {
...
 pub trait Router {
 	/// Finds a [`Route`] between `payer` and `payee` for a payment with the given values.
-	fn find_route<S: Score>(
+	fn find_route(
 		&self, payer: &PublicKey, route_params: &RouteParameters, payment_hash: &PaymentHash,
-		first_hops: Option<&[&ChannelDetails]>, scorer: &S
+		first_hops: Option<&[&ChannelDetails]>, inflight_htlc_map: HashMap<(u64, bool), u64>
 	) -> Result<Route, LightningError>;
+	/// Lets the router know that payment through a specific path has failed.
+	fn payment_path_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64);
```
tnull, line 283: Since the context for these methods might not entirely be clear at this point and a user might not be sure when to use them, could you mention in the comments that they will update the internal scorer? Also, maybe we might want to prefix them with `notify_` at this point?

lightning-invoice/src/payment.rs
```diff
@@ -314,10 +275,18 @@ pub trait Payer {
 /// A trait defining behavior for routing an [`Invoice`] payment.
 pub trait Router {
 	/// Finds a [`Route`] between `payer` and `payee` for a payment with the given values.
-	fn find_route<S: Score>(
+	fn find_route(
 		&self, payer: &PublicKey, route_params: &RouteParameters, payment_hash: &PaymentHash,
-		first_hops: Option<&[&ChannelDetails]>, scorer: &S
+		first_hops: Option<&[&ChannelDetails]>, inflight_htlc_map: HashMap<(u64, bool), u64>
```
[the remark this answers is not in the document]
jurvis: good idea. I'll add that :)

lightning-invoice/src/payment.rs
jurvis: `Router`s will not necessarily have an internal scorer unless they use `DefaultRouter.` 

Perhaps we could mention that this method is designed for delegating path failure events to an (optional) scorer?

lightning-invoice/src/payment.rs
```diff
@@ -1766,9 +1742,15 @@ mod tests {
 		assert_eq!(inflight_map.len(), 5);
 	}
 
-	struct TestRouter;
+	struct TestRouter {
+		scorer: RefCell<TestScorer>,
+	}
 
 	impl TestRouter {
+		fn new(scorer: RefCell<TestScorer>) -> Self {
```
jkczyz, line 1740: Let's hide the `RefCell` part of the interface by having this method wrap the scorer.

lightning-invoice/src/payment.rs
```diff
@@ -1860,6 +1842,8 @@ mod tests {
 
 					// Since the path is reversed, the last element in our iteration is the first
 					// hop.
+					let mut locked_scorer = self.scorer.lock();
+					let scorer = AccountForInFlightHtlcs::new(locked_scorer.deref_mut(), inflight_htlc_map.clone());
```
jkczyz, line 1836: It's a little strange that we are testing `AccountForInFlightHtlcs` this way rather through `DefaultRouter`. Then again, seems doing the latter would require using the actual `find_route` code, which wouldn't lend to a simple test setup. Fine with keeping this as is, but wanted to raise this in case others have any better ideas.

lightning-invoice/src/utils.rs
```diff
@@ -440,33 +440,63 @@ fn filter_channels(channels: Vec<ChannelDetails>, min_inbound_capacity_msat: Opt
...
 			let mut locked_random_seed_bytes = self.random_seed_bytes.lock().unwrap();
 			*locked_random_seed_bytes = sha256::Hash::hash(&*locked_random_seed_bytes).into_inner();
 			*locked_random_seed_bytes
 		};
-		find_route(payer, params, &self.network_graph, first_hops, &*self.logger, scorer, &random_seed_bytes)
+
+		find_route(
+			payer, params, &self.network_graph, first_hops, &*self.logger,
+				&AccountForInFlightHtlcs::new(&mut self.scorer.lock(), inflight_htlc_map),
+				   &random_seed_bytes
```
jkczyz, line 482: These should be indented the same amount as the previous line since they are all parameters to the same function.

lightning-invoice/src/utils.rs
```diff
@@ -510,6 +540,60 @@ where
...
+		self.scorer.payment_path_successful(path)
+	}
+
+	fn probe_failed(&mut self, path: &[&RouteHop], short_channel_id: u64) {
+		self.scorer.probe_failed(path, short_channel_id)
+	}
+
+	fn probe_successful(&mut self, path: &[&RouteHop]) {
+		self.scorer.probe_successful(path)
+	}
```
jkczyz, line 593: These are never actually called since `AccountForInFlightHtlcs` is only used for `channel_penalty_msat`, so they can be implemented as `unreachable!()`. I wonder if we should split `Score` into two different traits or if we are fine with duplicating part of the interface in `Router`. I guess either way, any `Score` implementation used by `DefaultRouter` would also need to implement this half of the trait. Not sure what would be prefered. @tnull @TheBlueMatt Any thoughts?

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jurvis: Added in 00766359

lightning-invoice/src/payment.rs
jurvis: Fixed in ea84acfe

lightning-invoice/src/payment.rs
jurvis: Fixed in ea84acfe

lightning-invoice/src/payment.rs
jurvis: Fixed in 9f34cbfe

lightning-invoice/src/utils.rs
jurvis: fixed in f7667f5e

lightning-invoice/src/utils.rs
jurvis: Addressed in 7d95196a

lightning-invoice/src/utils.rs
```diff
@@ -510,6 +540,52 @@ where
 	}
 }
 
+
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+pub(crate) struct AccountForInFlightHtlcs<'a, S: Score> {
+	scorer: &'a mut S,
+	/// Maps a channel's short channel id and its direction to the liquidity used up.
+	inflight_htlcs: HashMap<(u64, bool), u64>,
```
tnull, line 548: Do we want to reuse the `InFlightHtlcs` structure here?

lightning-invoice/src/utils.rs
```diff
@@ -510,6 +540,52 @@ where
 	}
 }
 
+
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
+pub(crate) struct AccountForInFlightHtlcs<'a, S: Score> {
```
tnull, line 545: Mh, do we come up with a name that mainly conveys what this object *is* in addition to what it *does*/its purpose is? Absolutely no blocker, but I find the naming a bit less descriptive than in could be. Unfortunately I'm not coming up with a good name either, just things like `InFlightHtlcScoreWrapper` or similar.

tnull (commented):
Generally LGTM.

lightning-invoice/src/utils.rs
jkczyz: I believe so, otherwise `InflightHtlcs::used_liquidity_msat` isn't actually called, IIUC. We'll need to define this in the `lightning` crate, however, given the dependency graph. May need to do so anyhow in the next follow-up, which will move tracking from `InvoicePayer` to `ChannelManager`.

lightning-invoice/src/utils.rs
jkczyz: Maybe S`corerAccountingForInFlightHtlcs`?

lightning-invoice/src/payment.rs
```diff
@@ -740,6 +740,28 @@ where
 	}
 }
 
+/// Data structure used to track information about inflight HTLCs.
+pub struct InflightHtlcs {
```
jkczyz, line 744: Let's capitalize the "f" for consistency with with `AccountsForInFlightHtlcs`. It "looks right" given in-flight is hyphenated. Though, I do prefer to keep variable names as `inflight` for no other reason than that it looks better than `in_flight`. 😛

jurvis: good point 😅

lightning-invoice/src/utils.rs
jurvis: resolved in [9606a8d](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/9606a8d52fbe6dfc2735c724128f56feb45ae76a)

lightning-invoice/src/payment.rs
jurvis: resolved in [d953a05](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/d953a059d5ea25296a9e0e7f7abe65480e0d14bd)

lightning-invoice/src/utils.rs
jurvis: resolved in [24f9dc6](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/24f9dc638d408567f0197d7df1e1c6e8bd447d3c)

lightning-invoice/src/utils.rs
```diff
@@ -510,6 +540,52 @@ where
 	}
 }
 
+
+/// Used to store information about all the HTLCs that are inflight across all payment attempts
```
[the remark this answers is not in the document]
jurvis: resolved in [d953a05](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/d953a059d5ea25296a9e0e7f7abe65480e0d14bd)

lightning-invoice/src/utils.rs
jkczyz: Ah, I wasn't thinking straight on where this needed to be defined since it is used internally in `AccountForInFlightHtlcs` it can remain where you have it.

lightning-invoice/src/payment.rs
```diff
@@ -108,14 +116,14 @@
 //!     match event {
 //!         Event::PaymentPathFailed { .. } => println!("payment failed after retries"),
 //!         Event::PaymentSent { .. } => println!("payment successful"),
-//!         _ => {},
+//!         _ => {}
```
jkczyz, line 119: Leave comma for consistency with the other arms.

lightning-invoice/src/payment.rs
```diff
@@ -15,9 +15,10 @@
 //! and payee using information provided by the payer and from the payee's [`Invoice`], when
 //! applicable.
 //!
-//! [`InvoicePayer`] is parameterized by a [`LockableScore`], which it uses for scoring failed and
-//! successful payment paths upon receiving [`Event::PaymentPathFailed`] and
-//! [`Event::PaymentPathSuccessful`] events, respectively.
+//! [`InvoicePayer`] uses its [`Router`] parameterization for optionally notifying scorers on
+//! the failure and success of payment paths upon receiving the  [`Event::PaymentPathFailed`] and
```
jkczyz, line 19: Drop "on the failure and success of payment paths" and "respectively" later. Better to be a less verbose given the redundancy.

lightning-invoice/src/payment.rs
```diff
@@ -15,9 +15,10 @@
...
+//! [`Event::PaymentPathSuccessful`] events, respectively. It also does the same for payment probe
+//! failure and success events using [`Event::ProbeFailed`] and [`Event::ProbeSuccessful`].
```
jkczyz, line 21: Similarly here.

lightning-invoice/src/payment.rs
jkczyz, line 19: Extra space after "receiving the".

lightning-invoice/src/payment.rs
```diff
@@ -45,7 +48,7 @@
 //! # use lightning::util::logger::{Logger, Record};
 //! # use lightning::util::ser::{Writeable, Writer};
 //! # use lightning_invoice::Invoice;
-//! # use lightning_invoice::payment::{InvoicePayer, Payer, Retry, Router};
+//! # use lightning_invoice::payment::{InvoicePayer, Payer, Retry, Router, InFlightHtlcs};
```
jkczyz, line 51: nit: order lexicographically

lightning-invoice/src/payment.rs
```diff
@@ -78,10 +81,15 @@
...
 //! # struct FakeRouter {}
 //! # impl Router for FakeRouter {
-//! #     fn find_route<S: Score>(
-//! #         &self, payer: &PublicKey, params: &RouteParameters, payment_hash: &PaymentHash,
-//! #         first_hops: Option<&[&ChannelDetails]>, scorer: &S
-//! #     ) -> Result<Route, LightningError> { unimplemented!() }
+//! # 	fn find_route(
+//! # 		&self, payer: &PublicKey, params: &RouteParameters, payment_hash: &PaymentHash,
+//! # 		first_hops: Option<&[&ChannelDetails]>, _inflight_htlcs: InFlightHtlcs
+//! # 	) -> Result<Route, LightningError> { unimplemented!() }
```
jkczyz, line 87: Let's use four spaces instead of tabs in the docs. Forget why this was the case. May just be simpler for editors.

lightning-invoice/src/payment.rs
```diff
@@ -78,10 +81,15 @@
...
+//! #
+//! #   fn notify_payment_path_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64) {  unimplemented!() }
+//! # 	fn notify_payment_path_successful(&self, path: Vec<&RouteHop>) {  unimplemented!() }
+//! # 	fn notify_payment_probe_successful(&self, path: Vec<&RouteHop>) {  unimplemented!() }
+//! # 	fn notify_payment_probe_failed(&self, path: Vec<&RouteHop>, short_channel_id: u64) { unimplemented!() }
```
jkczyz, line 92: Similarly here. One space after `#` and four spaces to indent.

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +740,48 @@ where
...
+
+impl InFlightHtlcs {
+	/// Creates a new instance with an optional map of liquidity information.
+	pub fn new(map: Option<HashMap<(u64, bool), u64>>) -> Self {
+		InFlightHtlcs { map: map.unwrap_or(HashMap::new()) }
+	}
+
+	/// Returns liquidity in msat given the public key of the HTLC source, destination, and short
+	/// channel id.
+	pub fn used_liquidity_msat(&self, source: &NodeId, destination: &NodeId, channel_scid: u64) -> Option<&u64> {
```
jkczyz, line 760: Use `target` instead of `destination` for consistent terminology. Likewise throughout the docs.

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +740,48 @@ where
 	}
 }
 
+/// Data structure used to track information about inflight HTLCs.
+pub struct InFlightHtlcs {
+	/// A map with liquidity value (in msat) keyed by a short channel id and the direction the HTLC
+	/// is traveling in. The direction boolean is determined by checking if the HTLC source's public
+	/// key is less than its destination. See [`InflightHtlcs::used_liquidity_msat`] for more
+	/// details.
+	pub map: HashMap<(u64, bool), u64>
```
jkczyz, line 749: Does this need to be `pub`?

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +740,48 @@ where
 	}
 }
 
+/// Data structure used to track information about inflight HTLCs.
+pub struct InFlightHtlcs {
+	/// A map with liquidity value (in msat) keyed by a short channel id and the direction the HTLC
+	/// is traveling in. The direction boolean is determined by checking if the HTLC source's public
+	/// key is less than its destination. See [`InflightHtlcs::used_liquidity_msat`] for more
```
jkczyz, line 747: `InFlightHtlcs` (or use `Self`)

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +740,48 @@ where
...
+	use utils::{ScorerAccountingForInFlightHtlcs, create_invoice_from_channelmanager_and_duration_since_epoch};
 	use bitcoin_hashes::sha256::Hash as Sha256;
 	use lightning::ln::PaymentPreimage;
 	use lightning::ln::features::{ChannelFeatures, NodeFeatures, InitFeatures};
 	use lightning::ln::functional_test_utils::*;
 	use lightning::ln::msgs::{ChannelMessageHandler, ErrorAction, LightningError};
 	use lightning::routing::gossip::{EffectiveCapacity, NodeId};
 	use lightning::routing::router::{PaymentParameters, Route, RouteHop};
-	use lightning::routing::scoring::ChannelUsage;
+	use lightning::routing::scoring::{ChannelUsage, Score, LockableScore};
```
jkczyz, line 777: Lexicographical order.

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +740,48 @@ where
...
+	/// A map with liquidity value (in msat) keyed by a short channel id and the direction the HTLC
+	/// is traveling in. The direction boolean is determined by checking if the HTLC source's public
+	/// key is less than its destination. See [`InflightHtlcs::used_liquidity_msat`] for more
+	/// details.
+	pub map: HashMap<(u64, bool), u64>
+}
+
+impl InFlightHtlcs {
+	/// Creates a new instance with an optional map of liquidity information.
+	pub fn new(map: Option<HashMap<(u64, bool), u64>>) -> Self {
```
jkczyz, line 754: Does this need to be an `Option`?

lightning-invoice/src/payment.rs
jkczyz, line 749: Let's name this `used_liquidity_msat` here and in the parameter to `new`.

lightning-invoice/src/payment.rs
```diff
@@ -487,7 +455,7 @@ where
 		let inflight_htlcs = self.create_inflight_map();
 		let route = self.router.find_route(
 			&payer, &params, &payment_hash, Some(&first_hops.iter().collect::<Vec<_>>()),
-			&AccountForInFlightHtlcs { scorer: &mut self.scorer.lock(), inflight_htlcs }
+			InFlightHtlcs::new(Some(inflight_htlcs))
```
jkczyz, line 458: Let's have `create_inflight_map` return this instead of wrapping at each call site. Also, rename to `create_inflight_htlcs`

lightning-invoice/src/payment.rs
jkczyz: Also, note that another option is to make the struct a tuple struct:

```rust
pub struct InFlightHtlcs(HashMap<(u64, bool), u64>);
```

lightning-invoice/src/payment.rs
jurvis: Fixed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addresed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in c5206b3bf646b61db957ef7dd80b79538114181d

lightning-invoice/src/payment.rs
jurvis: Addressed in 2f811f39f6aec80d8c9531853c91d646d31df98f

lightning-invoice/src/payment.rs
jurvis: Addressed in 2f811f39f6aec80d8c9531853c91d646d31df98f

lightning-invoice/src/payment.rs
jurvis: Addressed in 2f811f39f6aec80d8c9531853c91d646d31df98f

lightning-invoice/src/payment.rs
jurvis: Addressed in 441f28ee177c101bd5ee68ff3273472e37f14690

jkczyz (dismissed):
LGTM. Will need a squash once other reviewers are ready.

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +740,40 @@ where
 	}
 }
 
+/// A map with liquidity value (in msat) keyed by a short channel id and the direction the HTLC
+/// is traveling in. The direction boolean is determined by checking if the HTLC source's public
+/// key is less than its destination. See [`InFlightHtlcs::used_liquidity_msat`] for more
+/// details.
+pub struct InFlightHtlcs(HashMap<(u64, bool), u64>);
```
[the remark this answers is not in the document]
jkczyz: If we keep it private and implement `Readable` and `Writeable`, would that suffice?

lightning-invoice/src/payment.rs
```diff
@@ -34,6 +35,8 @@
 //! #
 //! # #[cfg(feature = "no-std")]
 //! # extern crate core2;
+//! # #[cfg(feature = "no-std")]
+//! # extern crate hashbrown;
```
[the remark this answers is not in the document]
jurvis: fixed in [b886529](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/b88652964b84352790bbc8385374e59f58f828e5)

lightning-invoice/src/payment.rs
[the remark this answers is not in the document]
jurvis: implemented `Readable` and `Writeable` for `InFlightHtlcs` in [b194c1d](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/b194c1d28d816d7cbc95c5b7dfcc037a62022618)

lightning-invoice/src/payment.rs
```diff
@@ -773,25 +738,53 @@ where
...
+
+impl InFlightHtlcs {
+	/// Returns liquidity in msat given the public key of the HTLC source, target, and short channel
+	/// id.
+	pub fn used_liquidity_msat(&self, source: &NodeId, target: &NodeId, channel_scid: u64) -> Option<&u64> {
+		self.0.get(&(channel_scid, source < target))
+	}
+}
+
+#[cfg(not(feature = "no-std"))]
```
tnull, line 755: Why does this have to be `std`?

tnull (commented):
Generally LGTM

lightning-invoice/src/payment.rs
jkczyz: Ah, I guess we have to do something like this:

https://github.com/lightningdevkit/rust-lightning/blob/48d21bad7b0b8dc567ed0e4e3f55c62e48fcb422/lightning/src/lib.rs#L84-L89

I didn't realize at the time that we already have a `core2` dependency in this crate.

tnull: Mh, couldn't we just use the re-export `lightning::io`? That's what I recently did in #1708 .

jkczyz: Yeah, that would work. I just wasn't sure what our conventions were, tbh. We have a dependency in `lightning-invoice`'s `Cargo.toml` file  from https://github.com/lightningdevkit/rust-lightning/commit/01915810d42506addbaef244b0cbb31e8b3856b5 but don't appear to use it anywhere outside of doc tests.

jkczyz: SGTM

jurvis: so it sounds like we want something like this instead of only implementing it for std?

```
#[cfg(feature = "std")]
/// Re-export of either `lightning::io` or `std::io`, depending on the `std` feature flag.
pub use std::io;
#[cfg(not(feature = "std"))]
/// Re-export of either `lightning::io` or `std::io`, depending on the `std` feature flag.
pub use lightning::io;

...

impl lightning::util::ser::Writeable for InFlightHtlcs {
	fn write<W: lightning::util::ser::Writer>(&self, writer: &mut W) -> Result<(), io::Error> { self.0.write(writer) }
}

impl lightning::util::ser::Readable for InFlightHtlcs {
	fn read<R: lightning::io::Read>(reader: &mut R) -> Result<Self, lightning::ln::msgs::DecodeError> {
		let infight_map: HashMap<(u64, bool), u64> = lightning::util::ser::Readable::read(reader)?;
		Ok(Self(infight_map))
	}
}

```

jkczyz: You should only need `use lightning::io;` (no `pub`, no conditional import). Basically, just inherit which `io` is used from the `lightning` crate rather than doing any configuration in `lightning-invoice`. Also, update doc tests accordingly.

jkczyz: `read`'s type parameter should use `io::Read` in its trait bound, too.

jurvis: 👍

jkczyz:
Feel free to squash.

jurvis:
squashed to [82b46ed](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/82b46edb2d4607b9627efa3cd5fa3062cb51db5a) without changes

jkczyz (approved):

lightning-invoice/src/payment.rs
```diff
@@ -45,16 +47,11 @@
...
 //! # use lightning::util::ser::{Writeable, Writer};
 //! # use lightning_invoice::Invoice;
-//! # use lightning_invoice::payment::{InvoicePayer, Payer, Retry, Router};
+//! # use lightning_invoice::payment::{InFlightHtlcs, InvoicePayer, Payer, Retry, Router};
 //! # use secp256k1::PublicKey;
 //! # use std::cell::RefCell;
 //! # use std::ops::Deref;
 //! #
-//! # #[cfg(not(feature = "std"))]
-//! # use core2::io;
```
tnull, line 54: Ah, you could then also remove the dependency, since it's not used anymore:
```
diff --git lightning-invoice/Cargo.toml lightning-invoice/Cargo.toml
index cd67bcd3..852f62f4 100644
--- lightning-invoice/Cargo.toml
+++ lightning-invoice/Cargo.toml
@@ -15,7 +15,7 @@ rustdoc-args = ["--cfg", "docsrs"]

 [features]
 default = ["std"]
-no-std = ["hashbrown", "lightning/no-std", "core2/alloc"]
+no-std = ["hashbrown", "lightning/no-std"]
 std = ["bitcoin_hashes/std", "num-traits/std", "lightning/std", "bech32/std"]

 [dependencies]
@@ -25,7 +25,6 @@ secp256k1 = { version = "0.24.0", default-features = false, features = ["recover
 num-traits = { version = "0.2.8", default-features = false }
 bitcoin_hashes = { version = "0.11", default-features = false }
 hashbrown = { version = "0.11", optional = true }
-core2 = { version = "0.3.0", default-features = false, optional = true }
 serde = { version = "1.0.118", optional = true }

 [dev-dependencies]
```

tnull (commented):
LGTM, one nit.

lightning-invoice/src/payment.rs
jurvis: 👍 addressed and squashed into [c353c3e](https://github.com/lightningdevkit/rust-lightning/pull/1694/commits/c353c3ed7c40e689a3b9fb6730c6dabbd3c92cc5)

tnull (approved):

jkczyz (approved):

PR #1705, opened 2022-09-08
Allow keys to be unpersisted

tnull:
Previously the `KVStorePersister` only allowed to `persist` a `Writeables` under a given key. However, we may want to be able to remove this set key after the fact. Here we therefore extend the trait with a `unpersist` method and implement it for `FilesystemPersister`.

I also added a commit rewording the `FilesystemPersister` docs to reflect the broader use case it fulfils by now.

### DIFF
--- a/lightning/src/util/persist.rs
+++ b/lightning/src/util/persist.rs
@@ -21,8 +21,12 @@ use super::{logger::Logger, ser::Writeable};
 /// and [`Persist`] traits.  It uses "manager", "network_graph",
 /// and "monitors/{funding_txo_id}_{funding_txo_index}" for keys.
 pub trait KVStorePersister {
-	/// Persist the given writeable using the provided key
+	/// Persist the given writeable using the provided key.
 	fn persist<W: Writeable>(&self, key: &str, object: &W) -> io::Result<()>;
+
+	/// Unpersist (i.e., remove) the writeable previously persisted under the provided key.
+	/// Returns `true` if the key was present, and `false` otherwise.
+	fn unpersist(&self, key: &str) -> io::Result<bool>;
 }
 
 /// Trait that handles persisting a [`ChannelManager`], [`NetworkGraph`], and [`WriteableScore`] to disk.
--- a/lightning-background-processor/src/lib.rs
+++ b/lightning-background-processor/src/lib.rs
@@ -717,6 +717,14 @@ mod tests {
 
 			self.filesystem_persister.persist(key, object)
 		}
+
+		fn unpersist(&self, key: &str) -> std::io::Result<bool> {
+			if key == "manager" || key == "network_graph" || key == "scorer" {
+				return Err(std::io::Error::new(std::io::ErrorKind::Other, "Refusing to unpersist this key"));
+			}
+
+			self.filesystem_persister.unpersist(key)
+		}
 	}
 
 	fn get_full_filepath(filepath: String, filename: String) -> String {
--- a/lightning-persister/src/lib.rs
+++ b/lightning-persister/src/lib.rs
@@ -128,6 +128,12 @@ impl KVStorePersister for FilesystemPersister {
 		dest_file.push(key);
 		util::write_to_file(dest_file, object)
 	}
+
+	fn unpersist(&self, key: &str) -> std::io::Result<bool> {
+		let mut dest_file = PathBuf::from(self.path_to_channel_data.clone());
+		dest_file.push(key);
+		util::delete_file(dest_file)
+	}
 }
 
 #[cfg(test)]
--- a/lightning-persister/src/lib.rs
+++ b/lightning-persister/src/lib.rs
@@ -28,8 +28,8 @@ use std::io::Cursor;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 
-/// FilesystemPersister persists channel data on disk, where each channel's
-/// data is stored in a file named after its funding outpoint.
+/// `FilesystemPersister` persists a given `Writeable` object on disk. In the case of channel data,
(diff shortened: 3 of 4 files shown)


tnull:
Let me know if I should incorporate anything discussed in #1417/#1426/#1427 in this PR, e.g., a `list()` method.

lightning/src/util/persist.rs
```diff
@@ -21,8 +21,12 @@ use super::{logger::Logger, ser::Writeable};
 /// and [`Persist`] traits.  It uses "manager", "network_graph",
 /// and "monitors/{funding_txo_id}_{funding_txo_index}" for keys.
 pub trait KVStorePersister {
-	/// Persist the given writeable using the provided key
+	/// Persist the given writeable using the provided key.
 	fn persist<W: Writeable>(&self, key: &str, object: &W) -> io::Result<()>;
+
+	/// Unpersist (i.e., remove) the writeable previously persisted under the provided key.
+	/// Returns `true` if the key was present, and `false` otherwise.
+	fn unpersist(&self, key: &str) -> io::Result<bool>;
```
G8XSU, line 29: Could you add motivation for adding this and where it will or can be used? 
* Its one thing if its for space optimization (delete items no longer being used)
* And another if we are going to be relying on this function for immediate delete of items. 
Depending on use-case, Ideally i would like it to be just former, to support some of the systems which support async delete. (And in this case we might not care much if it doesn't succeed immediately or is not strongly consistent.)

tnull: Do you mean add to motivation in comments, in the commit message, or here on GitHub?

I generally think that it would be nice to have `list` and `unpersist` functionalities, if the persister is going to be used for a wider array of objects. My particular motivation was the initial draft of a `PaymentInfoStore` for `LdkLite` (https://github.com/lightningdevkit/ldk-lite/pull/13) which uses the persister, where I however didn't want to store one huge blob of data under a single key. I therefore now store each payment info separately using the payment hash as a key. It would however be nice to be able to prune that store eventually, hence the need for an `unpersist` method.

G8XSU: I meant in PR description.(for future reference as well)

Got it, from your description our intention is space optimization. (i.e. pruning)
It would be best if we add to documentation of this api that this will be used for pruning. and hence we can utilize async deletes or eventually consistent deletes. 
This also means: 
1. Logic correctness shouldn't depend on immediate delete.
2. Even if we miss on deleting an item, worst that will happen is some wasted storage.

tnull: > 1. Logic correctness shouldn't depend on immediate delete.
> 2. Even if we miss on deleting an item, worst that will happen is some wasted storage.

I tend to disagree. Yes, the current motivation for this is pruning some data which is fine to be delayed, which however might not be the case for other future use cases. It would also pretty annoying to just silently fail to unpersist for some reason. But, more importantly, I think it would be very confusing to have an interface in which different but very similarly named methods behave differently or provide very different consistency guarantees. I'd therefore argue that while data removal is almost always not as critical as persistence, we should try to align the guarantees two methods as far as possible?

G8XSU: Another way is to have a flag, which allows for both. (if you think there will be future requirement for it and feel strongly about it) (imo we should keep it for pruning purpose to start with or by Default)
Not all systems provide same guarantees while deleting and we should think beyond filesystem-persistence.

And in LN and LDK usecase, persist and delete can live with different consistency requirements. In all case we need persist to be strongly consistent, whereas its not the same for delete.

The concern for "methods behave differently" is valid but that doesn't mean we should make api or consistency decisions based on that. 
In which case do we expect immediate delete?
We already have big overhaul ongoing for async persist, and if we can be conservative while adding delete functionality then why not ?

tnull: Alright. I think we currently don't know the exact requirements emerging from future use-cases or backends. I'd argue that don't want to have a flag there, which would make the interface really confusing. Other than that we should probably just figure it out once the broader requirements get a bit more clear? 

Do you have any concrete changes in mind I should include in this PR already?

tnull: > I don't think a separate key per payment is the right approach here, either? That is a lot of keys, and with very small values. Generally many filesystems are really bad at folders with lots of keys, as many operations expect to load the full set of keys in a folder at a time, so I don't think we can do that in general.

Yes, but I think the current `PaymentInfoStore` draft should really be a starting point, since we want a database backend for these kinds of data anyways. So regarding that concrete use case I'd argue we still want add `unpersist` to the trait, but implement a `KVStorePersister` with a different backend in LDKLite (probably using the `sled` database at first, since it's already in use for BDK persistence).

tnull: Hm, but I don't see why we then have an upstream trait at all, if it doesn't capture functionalities that ~all implementations would need? Even more so than for `unpersist` this is true for `list` and `get` functionalities, which literally *every* user/implementation will need. Take for example `ldk-sample` which of course also needs these functionalities for initialization, but solves the problem by simply bypassing the trait and even `FilesystemPersister` by just going through `std::fs`. However, this of course doesn't work for all backend implementations, so while the trait is supposed to enable different backends, it currently just works if we assume storing data in files.

In particular if we want to make the persister configurable for the user, e.g., in LDKLite, the `KVStorePersister` trait needs to offer a way to list and retrieve the data stored for a specific key. Of course this could be solved by yet another augmenting downstream trait specific to LDKLite, but this is not the idea of a general `KVStorePersister` trait, is it?

tnull:
Talked about this offline, closing this for now, will open a follow-up for `get`/`list` and implement a `Unpersist` trait downstream.

PR #1721, opened 2022-09-14
Fix typo in comment in Cargo.toml

acid-bit:
Hey! I was reading the code and spotted a typo, so here is the fix (starting small :))


tnull (approved):
Welcome! :)

valentinewallace (approved):

PR #1724, opened 2022-09-15
Support onion message pathfinding

valentinewallace:
Take 2, Github wouldn't let me reopen #1669 🤔 

We could reuse router::get_route, but it's better to start from scratch
because get_route includes a lot of payment-specific computations that impact
performance.

- [x] check for OM feature bit
- [x] finish tests

Partially addresses https://github.com/lightningdevkit/rust-lightning/issues/1607

~Based on #1731~ 

Known followups:
* scoring
* benchmarking (will need to update the netgraph file to support the OM feature bit)

### DIFF
--- a/lightning/src/routing/onion_message.rs
+++ b/lightning/src/routing/onion_message.rs
@@ -48,6 +48,7 @@ pub fn find_path<L: Deref, GL: Deref>(
 	let network_nodes = graph_lock.nodes();
 	let our_node_id = NodeId::from_pubkey(our_node_pubkey);
 	let dest_node_id = NodeId::from_pubkey(destination);
+	if our_node_id == dest_node_id { return Err(Error::InvalidDestination) }
 
 	// Add our start and first-hops to `frontier`, which is the set of hops that we'll next explore.
 	let start = NodeId::from_pubkey(&our_node_pubkey);
@@ -56,6 +57,7 @@ pub fn find_path<L: Deref, GL: Deref>(
 	frontier.push(PathBuildingHop { cost: 0, node_id: start, parent_node_id: start });
 	if let Some(first_hops) = first_hops {
 		for hop in first_hops {
+			if hop.counterparty.node_id == *our_node_pubkey { return Err(Error::InvalidFirstHop) }
 			if !hop.counterparty.features.supports_onion_messages() { continue; }
 			let node_id = NodeId::from_pubkey(&hop.counterparty.node_id);
 			frontier.push(PathBuildingHop { cost: 1, node_id, parent_node_id: start });
@@ -110,6 +112,10 @@ pub enum Error {
 	PathNotFound,
 	/// We failed to convert this node id into a [`PublicKey`].
 	InvalidNodeId(secp256k1::Error),
+	/// We attempted to generate a path to ourselves, which is not allowed.
+	InvalidDestination,
+	/// First hops cannot have our node id as a counterparty node id.
+	InvalidFirstHop,
 }
 
 impl fmt::Display for Error {
@@ -118,6 +124,8 @@ impl fmt::Display for Error {
 			Error::PathNotFound => write!(f, "Failed to find a path to the destination"),
 			Error::InvalidNodeId(e) =>
 				write!(f, "Failed to convert a node id into a PublicKey with error: {}", e),
+			Error::InvalidDestination => write!(f, "Cannot generate a route to ourselves"),
+			Error::InvalidFirstHop => write!(f, "First hops cannot have our node id as a counterparty node id"),
 		}
 	}
 }
@@ -249,5 +257,22 @@ mod tests {
 		let path = super::find_path(&our_id, &node_pks[0], &network_graph, Some(&our_chans.iter().collect::<Vec<_>>()), Arc::clone(&logger)).unwrap();
 		assert_eq!(path.len(), 1);
 		assert_eq!(path[0], node_pks[0]);
+	}
+
+	#[test]
+	fn invalid_first_hop() {
+		// Check that we can't generate a path if first_hops contains a counterparty node id that
+		// is equal to our node id.
+		let mut features = InitFeatures::empty();
+		features.set_onion_messages_optional();
+		let (secp_ctx, network_graph, gossip_sync, _, logger) = build_graph_with_features(features.to_context());
+		let (_, our_id, privkeys, node_pks) = get_nodes(&secp_ctx);
+
+		let bad_first_hop = vec![get_channel_details(Some(2), our_id, features, 100000)];
+		let err = super::find_path(&our_id, &node_pks[2], &network_graph, Some(&bad_first_hop.iter().collect::<Vec<_>>()), Arc::clone(&logger)).unwrap_err();
+		assert_eq!(err, super::Error::InvalidFirstHop);
+
+		let path = super::find_path(&our_id, &node_pks[2], &network_graph, None, Arc::clone(&logger)).unwrap();
+		assert_eq!(path.len(), 2);
 	}
 }
--- a/lightning/src/routing/onion_message.rs
+++ b/lightning/src/routing/onion_message.rs
@@ -1,0 +1,188 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! [Onion message] pathfinding lives here.
+//!
+//! Finding paths for onion messages is necessary for retrieving invoices and fulfilling invoice
+//! requests from [offers]. It differs from payment pathfinding in that channel liquidity, fees, and
+//! knobs such as `htlc_maximum_msat` do not factor into path selection -- onion messages require a
+//! peer connection and nothing more. However, we still use the network graph because onion messages
+//! between channel peers are likely to be prioritized over those between non-channel peers.
+//!
+//! [Onion message]: crate::onion_message
+//! [offers]: <https://github.com/lightning/bolts/pull/798>
+
+use bitcoin::secp256k1::{self, PublicKey};
+
+use crate::ln::channelmanager::ChannelDetails;
+use crate::routing::gossip::{NetworkGraph, NodeId};
+use crate::util::logger::Logger;
+
+use alloc::collections::BinaryHeap;
+use core::cmp;
+use core::fmt;
+use core::ops::Deref;
+use crate::prelude::*;
+
+/// Finds a route from us to the given `destination` node.
+/// If we have private channels, it may be useful to fill in `first_hops` with the results from
+/// [`ChannelManager::list_usable_channels`]. If `first_hops` is not filled in, connected peers
+/// without public channels will not be forwarded over.
+///
+/// [`ChannelManager::list_usable_channels`]: crate::ln::channelmanager::ChannelManager::list_usable_channels
+pub fn find_path<L: Deref, GL: Deref>(
+	our_node_pubkey: &PublicKey, destination: &PublicKey, network_graph: &NetworkGraph<GL>, first_hops: Option<&[&ChannelDetails]>, logger: L
+) -> Result<Vec<PublicKey>, Error> where L::Target: Logger, GL::Target: Logger
+{
+	log_trace!(logger, "Searching for an onion message path from origin {} to destination {} and {} first hops {}overriding the network graph",
+		our_node_pubkey, destination, first_hops.map(|hops| hops.len()).unwrap_or(0),
+		if first_hops.is_some() { "" } else { "not " });
+	let graph_lock = network_graph.read_only();
+	let network_channels = graph_lock.channels();
+	let network_nodes = graph_lock.nodes();
+	let our_node_id = NodeId::from_pubkey(our_node_pubkey);
+	let dest_node_id = NodeId::from_pubkey(destination);
+
+	// Add our start and first-hops to `frontier`, which is the set of hops that we'll next explore.
+	let start = NodeId::from_pubkey(&our_node_pubkey);
+	let mut frontier = BinaryHeap::new();
+	frontier.push(PathBuildingHop { cost: 0, node_id: start, parent_node_id: start });
+	if let Some(first_hops) = first_hops {
+		for hop in first_hops {
+			if !hop.counterparty.features.supports_onion_messages() { continue; }
+			let node_id = NodeId::from_pubkey(&hop.counterparty.node_id);
+			frontier.push(PathBuildingHop { cost: 1, node_id, parent_node_id: start });
+		}
+	}
+
+	let mut visited = HashMap::new();
+	while let Some(PathBuildingHop { cost, node_id, parent_node_id }) = frontier.pop() {
+		match visited.entry(node_id) {
+			hash_map::Entry::Occupied(_) => continue,
+			hash_map::Entry::Vacant(e) => e.insert(parent_node_id),
+		};
+		if node_id == dest_node_id {
+			let path = reverse_path(visited, our_node_id, dest_node_id)?;
+			log_info!(logger, "Got route to {:?}: {:?}", destination, path);
+			return Ok(path)
+		}
+		if let Some(node_info) = network_nodes.get(&node_id) {
+			for scid in &node_info.channels {
+				if let Some(chan_info) = network_channels.get(&scid) {
+					if let Some((_, successor)) = chan_info.as_directed_from(&node_id) {
+						// We may push a given successor multiple times, but the heap should sort its best entry
+						// to the top. We do this because there is no way to adjust the priority of an existing
+						// entry in `BinaryHeap`.
+						frontier.push(PathBuildingHop {
+							cost: cost + 1,
+							node_id: *successor,
+							parent_node_id: node_id,
+						});
+					}
+				}
+			}
+		}
+	}
+
+	Err(Error::PathNotFound)
+}
+
+#[derive(Debug, PartialEq)]
+/// Errors that might occur running [`find_path`].
+pub enum Error {
+	/// We failed to find a path to the destination.
+	PathNotFound,
+	/// We failed to convert this node id into a [`PublicKey`].
+	InvalidNodeId(secp256k1::Error),
+}
+
+impl fmt::Display for Error {
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		match self {
+			Error::PathNotFound => write!(f, "Failed to find a path to the destination"),
+			Error::InvalidNodeId(e) =>
+				write!(f, "Failed to convert a node id into a PublicKey with error: {}", e),
+		}
+	}
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
+
+#[derive(Eq, PartialEq)]
+struct PathBuildingHop {
+	cost: u64,
+	node_id: NodeId,
+	parent_node_id: NodeId,
+}
+
+impl PartialOrd for PathBuildingHop {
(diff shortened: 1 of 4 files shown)


valentinewallace:
I think I addressed all the feedback from #1669. Much cleaner! 

If there are rough conceptual ACKs, I can add more testing, add feature bit checks, and open the first commit in a separate PR.

TheBlueMatt (commented):
At first glance it looks like it'll work. We definitely need some ability to score nodes, if only a "skip nodes" set, as we'll want the ability to, say, fetch two non-overlapping paths.

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,176 @@
...
+			return Ok(reverse_path(visited, our_node_id, dest_node_id, logger)?)
+		}
+		if let Some(node_info) = network_nodes.get(&node_id) {
+			for scid in &node_info.channels {
+				if let Some(chan_info) = network_channels.get(&scid) {
+					if let Some((_, successor)) = chan_info.as_directed_from(&node_id) {
+						// We may push a given successor multiple times, but the heap should sort its best entry
+						// to the top. We do this because there is no way to adjust the priority of an existing
+						// entry in `BinaryHeap`.
+						frontier.push(PathBuildingHop {
```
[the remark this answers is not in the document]
valentinewallace: Yeah, it seems slower. 

Not checking if `successor` is in `visited`:
```
test routing::onion_message::benches::generate_routes_test                   ... bench:  27,532,191 ns/iter (+/- 13,250,916)
test routing::onion_message::benches::generate_routes_test                   ... bench:  25,474,037 ns/iter (+/- 13,319,206)
test routing::onion_message::benches::generate_routes_test                   ... bench:  27,410,199 ns/iter (+/- 15,179,713)
```

Checking:
```
test routing::onion_message::benches::generate_routes_test                   ... bench:  32,435,508 ns/iter (+/- 11,682,837)
test routing::onion_message::benches::generate_routes_test                   ... bench:  30,536,545 ns/iter (+/- 11,874,035)
test routing::onion_message::benches::generate_routes_test                   ... bench:  31,512,712 ns/iter (+/- 12,148,234)
```

valentinewallace:
> We definitely need some ability to score nodes, if only a "skip nodes" set, as we'll want the ability to, say, fetch two non-overlapping paths.

Sounds good, will likely do it in follow-up if that works

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,185 @@
...
+struct PathBuildingHop {
+	cost: u64,
+	node_id: NodeId,
+	parent_node_id: NodeId,
+}
+
+impl PartialOrd for PathBuildingHop {
+	fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+		// We need a min-heap, whereas `BinaryHeap`s are a max-heap, so compare the costs in reverse.
+		other.cost.partial_cmp(&self.cost)
```
naumenkogs, line 105: Assuming the cost is equal, can someone grind lower node-id to attract more message flows? After staring at rust documentation for 5 minutes, I'm not sure.

lightning/src/routing/onion_message.rs
```diff
@@ -60,6 +60,9 @@ pub fn find_path<L: Deref, GL: Deref>(
 			return Ok(reverse_path(visited, our_node_id, dest_node_id, logger)?)
 		}
 		if let Some(node_info) = network_nodes.get(&node_id) {
+			if let Some(node_ann) = &node_info.announcement_info {
+				if !node_ann.features.supports_onion_messages() { continue; }
+			} else if node_id != our_node_id { continue; }
```
naumenkogs, line 65: I think this line is particularly hard to understand, could you add a comment? Or even, do this check 3 lines before even accessing `announcement_info`.

lightning/src/routing/onion_message.rs
valentinewallace: Added a TODO for this

lightning/src/routing/onion_message.rs
valentinewallace: Lmk if the new version + comment make sense

lightning/src/routing/onion_message.rs
naumenkogs: So you confirm it's possible by looking at the rust doc?

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,328 @@
...
+			None => return Err(Error::PathNotFound),
+			Some(parent) => {
+				if *parent == our_node_id { break; }
+				curr = *parent;
+			}
+		}
+	}
+
+	path.reverse();
+	log_info!(logger, "Got route to {:?}: {:?}", destination, path);
```
tnull, line 148: As `reverse_path` is kind of a helper function, maybe this log could be placed in `find_route` itself?

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,328 @@
...
+			hash_map::Entry::Occupied(_) => continue,
+			hash_map::Entry::Vacant(e) => e.insert(parent_node_id),
+		};
+		if node_id == dest_node_id {
+			return Ok(reverse_path(visited, our_node_id, dest_node_id, logger)?)
+		}
+		if let Some(node_info) = network_nodes.get(&node_id) {
+			// Only consider the network graph if first_hops does not override it.
+			if valid_first_hops.contains(&node_id) || node_id == our_node_id {
+			} else if let Some(node_ann) = &node_info.announcement_info {
```
tnull, line 68: IIUC this is only to account for the case where the first hop features are different than the network-announced ones? What would be a case in which we can expect that?

Also, this currently seems to default to `supports_onion_messages == true` in case we haven't gotten a node announcement. Maybe this should rather be pessimistic and exclude a hop that may not support onion messages?

tnull (commented):
First pass, looks good! Haven't looked at tests yet.

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,328 @@
...
+/// If we have private channels, it may be useful to fill in `first_hops` with the results from
+/// [`ChannelManager::list_usable_channels`]. If `first_hops` is not filled in, connected peers
+/// without public channels will not be forwarded over.
+///
+/// [`ChannelManager::list_usable_channels`]: crate::ln::channelmanager::ChannelManager::list_usable_channels
+pub fn find_path<L: Deref, GL: Deref>(
+	our_node_pubkey: &PublicKey, destination: &PublicKey, network_graph: &NetworkGraph<GL>, first_hops: Option<&[&ChannelDetails]>, logger: L
+) -> Result<Vec<PublicKey>, Error> where L::Target: Logger, GL::Target: Logger
+{
+	log_trace!(logger, "Searching for an onion message path from origin {} to destination {} and {} first hops {}overriding the network graph",
```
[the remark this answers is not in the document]
valentinewallace: Note that the current version matches the equivalent log in `get_route`. We could do that but IIUC we'd have to add a call to `format!` to insert the number of first_hops, which allocates a String, which doesn't seem worth?

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,328 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Onion message pathfinding lives here.
```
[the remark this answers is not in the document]
valentinewallace: Lmk if the new docs look good

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,328 @@
...
+			}
+		}
+	}
+
+	Err(Error::PathNotFound)
+}
+
+#[derive(Debug, PartialEq)]
+/// Errored running [`find_path`].
+pub enum Error {
```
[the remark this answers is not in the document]
valentinewallace: I didn't do `Error` because I'm not sure what implementing `source` would look like here, but did `Display`

lightning/src/routing/onion_message.rs
valentinewallace: > Also, this currently seems to default to supports_onion_messages == true in case we haven't gotten a node announcement. Maybe this should rather be pessimistic and exclude a hop that may not support onion messages?

Indeed, that was a regression 🙈 

> IIUC this is only to account for the case where the first hop features are different than the network-announced ones? What would be a case in which we can expect that?

Basically our local channel view is expected to be the most up-to-date. Gossip can be slow but we'll be the first to know about feature upgrade/downgrades for our channel peers

lightning/src/routing/onion_message.rs
naumenkogs: Perhaps fine to keep a todo anyway because:
- the way nodes sorted before insertion into heap might play a role then
- heap implementation may change

lightning/src/routing/onion_message.rs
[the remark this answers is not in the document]
tnull: I believe when the error already implements `Display` and `Debug`, you should be fine with

```rust
#[cfg(feature = "std")]
impl std::error::Error for Error {}
```

lightning/src/routing/onion_message.rs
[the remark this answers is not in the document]
tnull: Fair enough, probably not worth.

lightning/src/routing/onion_message.rs
tnull: > Basically our local channel view is expected to be the most up-to-date. Gossip can be slow but we'll be the first to know about feature upgrade/downgrades for our channel peers

Ah, I see. Makes sense!

lightning/src/routing/onion_message.rs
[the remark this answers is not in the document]
tnull: Generally looks good. Is the current mix of `//!`, `///`, and `//` as intended though?

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,283 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! [Onion message] pathfinding lives here. Finding paths for onion messages is necessary for
```
tnull, line 10: nit: If you put the first sentence in its own paragraph, it will be the only one shown in  the 'Modules' overview of the [`lightning::routing` docs](https://docs.rs/lightning/latest/lightning/routing/index.html).
```suggestion
//! [Onion message] pathfinding lives here.
//!
//! Finding paths for onion messages is necessary for
```

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,283 @@
...
+//! retrieving invoices and fulfilling invoice requests from [offers]. It differs from payment
+//! pathfinding in that channel liquidity, fees, and knobs such as `htlc_maximum_msat` do not factor
+//! into path selection -- onion messages require a peer connection and nothing more. However, we
+//! still use the network graph because onion messages between channel peers are likely to be
+//! prioritized over those between non-channel peers.
+//!
+//! [Onion message]: crate::onion_message
+//! [offers]: <https://github.com/lightning/bolts/pull/798>
+
+// Note that we do not reuse `router::find_route` because it contains a lot of payment-specific
```
tnull, line 20: nit: Not sure we need this comment anymore since you already explained it in more detail above?

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,283 @@
...
+
+		let path = super::find_path(&our_id, &node_pks[2], &network_graph, None, Arc::clone(&logger)).unwrap();
+		assert_eq!(path.len(), 2);
+		// See routing::test_utils::build_graph ASCII graph, the first hop can be any of these
+		assert!(path[0] == node_pks[1] || path[0] == node_pks[7] || path[0] == node_pks[0]);
+		assert_eq!(path[1], node_pks[2]);
+	}
+
+	#[test]
+	fn three_hops() {
```
tnull, line 210: Not sure whether this test case is not a superset of `one_hop`/`two_hops`? Do we need all three, or can we just keep this one?

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,283 @@
...
+		assert_eq!(path[1], node_pks[2]);
+		assert_eq!(path[2], node_pks[0]);
+
+		// If we specify a channel to node1, that overrides our local channel view and that gets used
+		let our_chans = vec![get_channel_details(Some(42), node_pks[0].clone(), features, 250_000_000)];
+		let path = super::find_path(&our_id, &node_pks[0], &network_graph, Some(&our_chans.iter().collect::<Vec<_>>()), Arc::clone(&logger)).unwrap();
+		assert_eq!(path.len(), 1);
+		assert_eq!(path[0], node_pks[0]);
+	}
+}
```
tnull, line 283: Maybe test cases for a) checking that valid first hops are included and b) that invalid first hops are not considered (cf. `router::tests::invalid_first_hop_test`) might be nice.

lightning/src/routing/test_utils.rs
```diff
@@ -131,7 +167,7 @@ pub(super) fn id_to_feature_flags(id: u8) -> Vec<u8> {
 	}
 }
 
-pub(super) fn build_line_graph() -> (
+pub(super) fn build_line_graph(node_features: NodeFeatures) -> (
```
tnull, line 170: nit: Rename to make it consistent:
```suggestion
pub(super) fn build_line_graph_with_features(node_features: NodeFeatures) -> (
```

tnull (commented):
Mostly focused on tests this time.

TheBlueMatt (commented):
Basically LGTM, one real comment.

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,280 @@
...
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! [Onion message] pathfinding lives here.
+//!
+//! Finding paths for onion messages is necessary for retrieving invoices and fulfilling invoice
+//! requests from [offers]. It differs from payment pathfinding in that channel liquidity, fees, and
+//! knobs such as `htlc_maximum_msat` do not factor into path selection -- onion messages require a
+//! peer connection and nothing more. However, we still use the network graph because onion messages
+//! between channel peers are likely to be prioritized over those between non-channel peers.
```
ariard, line 16: So while I think relying on the existence of channels between peers to flow onion messages align well the incentives between bandwidth consumed and onion forwards accepted, I don't know if this is a direction we would like for the long-term in term of privacy. Indeed, overlapping both offers flows and HTLC flows, if we re-use the same heuristics than for classic pathfinding we might have overlap and show more "data trails" to any payment observers. Especially onion messages routing might invalidate some privacy-preserving enhancing we have implemented with HTLCs such as https://github.com/lightningdevkit/rust-lightning/pull/1286. The lack of timing padding in onion routing might cancel the CLTV delta offset. Of course, this is a subject we can think more and improve with time. Just to be aware of about the privacy regression we might have.

ariard (commented):
Nice if we can this back, figured out gonna be needed for #1848 and beyond. Will review soon.

valentinewallace:
> Nice if we can this back, figured out gonna be needed for #1848 and beyond. Will review soon.

Thanks, I don't know if I'll be prioritizing the follow-ups for a while (open to someone taking them over), but since this seems close(?) I'll try to get back to it in the new year

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,280 @@
...
+	let network_nodes = graph_lock.nodes();
+	let our_node_id = NodeId::from_pubkey(our_node_pubkey);
+	let dest_node_id = NodeId::from_pubkey(destination);
+	if our_node_id == dest_node_id { return Err(Error::InvalidDestination) }
+
+	// Add our start and first-hops to `frontier`.
+	let start = NodeId::from_pubkey(&our_node_pubkey);
+	let mut valid_first_hops = HashSet::new();
+	let mut frontier = BinaryHeap::new();
+	frontier.push(PathBuildingHop { cost: 0, node_id: start, parent_node_id: start });
```
ariard, line 57: Can define the reasoning behind `PathBuildingHop::cost`. I don't know if the onion bandwidth we're expecting from hops on the network is scaled on the channel capacity. The most rule of thumb metric I would guess.

ariard (commented):
Beyond offer transport, I think onion messages pathfinding could be useful to two more use-cases: bitcoin transaction relay and jamming credentials. For the former, I believe the set of heuristics is going to be really-specific, as you might avoid centralised hops to improve fault-tolerance. For the latter, the use-case requirements are more likely to be similar to offers, however you might do synthetic traffic to mask the economic volume.

Just to say, I don't think we'll need more than one onion messages pathfinding algorithms for other use-cases, however I believe they might need each to plug a `CustomHeuristicsOnionScore` similar to `scoring::Score`, or combine those `CustomHeuristicsOnionScore`. Thinking aloud for jamming.

ariard:
@naumenkogs if you're bored, I'm pretty sure onion message pathfinding will be a must to play with a bunch of things, notably jamming mitigations :)

lightning/src/routing/onion_message.rs
```diff
@@ -0,0 +1,280 @@
...
+		}
+		if let Some(node_info) = network_nodes.get(&node_id) {
+			// Only consider the network graph if first_hops does not override it.
+			if valid_first_hops.contains(&node_id) || node_id == our_node_id {
+			} else if let Some(node_ann) = &node_info.announcement_info {
+				if !node_ann.features.supports_onion_messages() || node_ann.features.requires_unknown_bits()
+				{ continue; }
+			} else { continue; }
+			for scid in &node_info.channels {
+				if let Some(chan_info) = network_channels.get(&scid) {
```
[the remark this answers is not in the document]
valentinewallace: I'm getting somewhat comparable times, with scid-heap narrowly taking the lead.

inserting node_ids into heap:
```
test routing::onion_message::benches::generate_simple_routes                 ... bench:  42,412,441 ns/iter (+/- 26,718,082)
test routing::onion_message::benches::generate_simple_routes                 ... bench:  40,244,529 ns/iter (+/- 27,297,264)
test routing::onion_message::benches::generate_simple_routes                 ... bench:  38,865,025 ns/iter (+/- 27,147,061)
test routing::onion_message::benches::generate_simple_routes                 ... bench:  43,582,987 ns/iter (+/- 23,421,789)
```

inserting scids into heap:
```
test routing::onion_message::benches::generate_simple_routes                 ... bench:  38,450,779 ns/iter (+/- 29,381,214)
test routing::onion_message::benches::generate_simple_routes                 ... bench:  37,930,562 ns/iter (+/- 27,023,358)
test routing::onion_message::benches::generate_simple_routes                 ... bench:  41,410,850 ns/iter (+/- 27,262,014)
test routing::onion_message::benches::generate_simple_routes                 ... bench:  39,008,820 ns/iter (+/- 26,009,606)
```

I pushed a WIP commit (https://github.com/lightningdevkit/rust-lightning/pull/1724/commits/28318ac76ca598ff9f37709202b9904dd7e9838a) with the scid-heap version if you want to test it out and/or check that I implemented it the way you expected.

valentinewallace:
I see @moneyball added this to the 119 milestone.

But I don't see the rush on this, because most cases will use direct-connect? It seems like this will only be useful once OMs are more widely supported across the network. @TheBlueMatt @jkczyz Let me know if I'm good to remove this from the milestone for now.

moneyball:
Yes I'll remove it. I had added it to 119 due to a misunderstanding of how it could work in the short-term.

PR #1726, opened 2022-09-16
BOLT 12 offer parsing

jkczyz:
Add common `bech32` parsing for BOLT 12 messages. The encoding is similar to `bech32` only without a checksum and with support for continuing messages across multiple parts.

Messages implementing `Bech32Encode` are parsed into a TLV stream, which is converted to the desired message content while performing semantic checks. Checking after conversion allows for more elaborate checks of data composed of multiple TLV records and for more meaningful error messages.

The parsed bytes are also saved to allow creating messages with mirrored data, even if TLV records are unknown.

Also, implement `Writeable` for `Offer` and have a corresponding `TryFrom` implementation for decoding the raw bytes.


lightning/src/offers/parse.rs
```diff
@@ -0,0 +1,112 @@
...
+	InvalidAmount,
+	/// A currency was provided that is not supported.
+	UnsupportedCurrency,
+	/// A required description was not provided.
+	MissingDescription,
+	/// A node id was not provided.
+	MissingNodeId,
+	/// An empty set of blinded paths was provided.
+	MissingPaths,
+	/// A quantity representing an empty range or that was outside of a valid range was provided.
```
valentinewallace, line 92: Bit confusing wording to parse

lightning/src/offers/offer.rs
```diff
@@ -384,12 +407,27 @@ impl OfferContents {
...
+	}
+}
+
 impl Writeable for OfferContents {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), io::Error> {
 		self.as_tlv_stream().write(writer)
 	}
 }
 
+impl TryFrom<Vec<u8>> for Offer {
```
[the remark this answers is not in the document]
jkczyz: Forgot to mention but also note that `Offer` needs to take ownership of the bytes.

lightning/src/offers/parse.rs
jkczyz: Rewrote the `InvalidQuantity` docs. PTAL.

jkczyz:
Updated to parse bech32 strings directly into the higher-level structs instead of an intermediary TLV stream format by adding a `TryFrom<Vec<u8>>` trait bound to `Bech32Encode` and dropping the associated type. This is more straightforward and also makes it so that checking that all bytes are read in #1738 only needs to be defined in one place.

jkczyz (commented):
All good feedback

lightning/src/offers/offer.rs
```diff
@@ -706,4 +831,266 @@ mod tests {
...
+				],
+			})
+			.build()
+			.unwrap();
+		if let Err(e) = offer.to_string().parse::<Offer>() {
+			panic!("error parsing offer: {:?}", e);
+		}
+
+		let mut builder = OfferBuilder::new("foo".into(), pubkey(42));
+		builder.offer.paths = Some(vec![]);
```
valentinewallace, line 928: I don't quite get why this should cause us to fail to parse

lightning/src/offers/offer.rs
```diff
@@ -706,4 +831,266 @@ mod tests {
...
+		match offer.to_string().parse::<Offer>() {
+			Ok(_) => panic!("expected error"),
+			Err(e) => assert_eq!(e, ParseError::InvalidSemantics(SemanticError::MissingPaths)),
+		}
+	}
+
+	#[test]
+	fn parses_offer_with_quantity() {
+		let offer = OfferBuilder::new("foo".into(), pubkey(42))
+			.supported_quantity(Quantity::one())
```
valentinewallace, line 940: is `fn one()` meant to be public?

lightning/src/offers/parse.rs
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		let data = Vec::<u8>::from_base32(&data)?;
+		Self::try_from(data)
+	}
+
+	/// Formats the message using bech32-encoding.
+	fn fmt_bech32_str(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+		bech32::encode_without_checksum_to_fmt(f, Self::BECH32_HRP, self.as_ref().to_base32())
+			.expect("HRP is valid").unwrap();
```
valentinewallace, line 51: nit: 
```suggestion
			.expect("HRP is invalid").unwrap();
```

valentinewallace (commented):
Nothing major :)

lightning/src/offers/offer.rs
jkczyz: I likely inferred it from this part of the spec:

```
  - if it is connected only by private channels:
    - MUST include `offer_paths` containing one or more paths to the node from
      publicly reachable nodes.
  - otherwise:
    - MAY include `offer_paths`.
```

jkczyz: True... I guess it is more that we fail if given a `paths` TLV record with length 0. We are fine parsing it if the `paths` TLV record doesn't exist. Could remove this check if you prefer.

jkczyz: Removed the check.

lightning/src/offers/offer.rs
```diff
@@ -425,19 +465,96 @@ tlv_stream!(OfferTlvStream, OfferTlvStreamRef, {
...
+
+impl TryFrom<ParsedMessage<OfferTlvStream>> for Offer {
+	type Error = ParseError;
+
+	fn try_from(offer: ParsedMessage<OfferTlvStream>) -> Result<Self, Self::Error> {
+		let ParsedMessage { bytes, tlv_stream } = offer;
+		let contents = OfferContents::try_from(tlv_stream)?;
+		Ok(Offer { bytes, contents })
+	}
+}
```
valentinewallace, line 488: I find this a bit more readable/concise, kinda jumping around the file right now: 
```
diff --git a/lightning/src/offers/offer.rs b/lightning/src/offers/offer.rs
index 5876acf6..a68a143f 100644
--- a/lightning/src/offers/offer.rs
+++ b/lightning/src/offers/offer.rs
@@ -400,7 +400,9 @@ impl TryFrom<Vec<u8>> for Offer {

        fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
                let parsed_message = ParsedMessage::<OfferTlvStream>::try_from(bytes)?;
-               Offer::try_from(parsed_message)
+               let ParsedMessage { bytes, tlv_stream } = parsed_message;
+               let contents = OfferContents::try_from(tlv_stream)?;
+               Ok(Offer { bytes, contents })
        }
 }

@@ -477,16 +479,6 @@ impl FromStr for Offer {
        }
 }

-impl TryFrom<ParsedMessage<OfferTlvStream>> for Offer {
-       type Error = ParseError;
-
-       fn try_from(offer: ParsedMessage<OfferTlvStream>) -> Result<Self, Self::Error> {
-               let ParsedMessage { bytes, tlv_stream } = offer;
-               let contents = OfferContents::try_from(tlv_stream)?;
-               Ok(Offer { bytes, contents })
-       }
-}
-
 impl TryFrom<OfferTlvStream> for OfferContents {
        type Error = SemanticError;

jkczyz: Yeah, good call.

valentinewallace (approved):

PR #1731, opened 2022-09-19
Move router test utils into their own module

valentinewallace:
Used in #1724 to test onion message pathfinding


tnull (commented):
Generally looks good, but it could get a bit confusing that there are now two `test_utils` that are even both referenced in `router.rs`. Maybe we could solve this for the time being by removing the `use` statements and always using the full `routing::test_utils`/`util::test_utils` prefix to make it explicit?

valentinewallace:
> Generally looks good, but it could get a bit confusing that there are now two `test_utils` that are even both referenced in `router.rs`. Maybe we could solve this for the time being by removing the `use` statements and always using the full `routing::test_utils`/`util::test_utils` prefix to make it explicit?

Hmm I feel like that would add a lot of verbosity since both of those test util modules are used in a ton of places.

Do you think it would help if I modified the import to be: `use util::test_utils as ln_test_utils;` so that all the current e.g. `test_utils::TestScorer` would be `ln_test_utils::TestScorer`?

dunxen (dismissed):
Makes sense and LGTM!

tnull:
> Do you think it would help if I modified the import to be: `use util::test_utils as ln_test_utils;` so that all the current e.g. `test_utils::TestScorer` would be `ln_test_utils::TestScorer`?

Yup, `use util::test_utils as ln_test_utils;` would def. do the trick. No hard requirement though, feel free to leave it out if you think it gets too verbose. Also, feel free to squash.

valentinewallace:
> Yup, `use util::test_utils as ln_test_utils;` would def. do the trick. No hard requirement though, feel free to leave it out if you think it gets too verbose, moving some to `bench_utils` already helped.
> 
> Also, feel free to squash.

Sg, I went ahead with that. I kinda prefer to keep the first big commit as a "pure code move" (modulo imports/exposing stuff as `pub(super)`) so I kept the commits separate

valentinewallace:
I don't think that CI failure is related to the PR 🤔

dunxen:
> I don't think that CI failure is related to the PR 🤔 

Yeah seems some no-std dependency broke the MSRV. (Assuming it's the same issue I have on one PR)

PR #1745, opened 2022-09-26
Make writing InvoicePayer tests simpler by automatically forwarding HLTC update messages

adamritter:
To make testing routing through multiple rounds of payment attempts with real channel balances simpler, I created a helper function to forward HTLC update and RevokeAndAck messages and refactored a complex test case.

 I didn't delete the original no_extra_retries_on_back_to_back_fail function, but I'm happy to delete it if the checks in it are not essential.

### DIFF
--- a/lightning-invoice/src/payment.rs
+++ b/lightning-invoice/src/payment.rs
@@ -2425,6 +2425,7 @@ mod tests {
 		assert!(expected_events.borrow().is_empty());
 	}
 
+	///	Helper function to forward HTLC update and RevokeAndAck messages
 	fn handle_message_send_events(sender:&Node, node_indices_by_id: &HashMap<PublicKey, usize>, nodes: &Vec<Node>) {
 		let mut to_notify : Vec<&Node>=Vec::new();
 		for msg in sender.node.get_and_clear_pending_msg_events() {
--- a/lightning-invoice/src/payment.rs
+++ b/lightning-invoice/src/payment.rs
@@ -2424,4 +2424,143 @@ mod tests {
 		nodes[0].node.process_pending_events(&invoice_payer);
 		assert!(expected_events.borrow().is_empty());
 	}
+
+	fn handle_message_send_events(sender:&Node, node_indices_by_id: &HashMap<PublicKey, usize>, nodes: &Vec<Node>) {
+		let mut to_notify : Vec<&Node>=Vec::new();
+		for msg in sender.node.get_and_clear_pending_msg_events() {
+			match msg {
+				MessageSendEvent::SendRevokeAndACK { node_id, msg } => {
+					println!("Sendrevoke {:?}", msg);
+					let idx = node_indices_by_id[&node_id];
+					let other_node=&nodes[idx];
+					other_node.node.handle_revoke_and_ack(&sender.node.node_id(), &msg);
+					to_notify.push(other_node);
+					// handle_message_send_events(&other_node, node_indices_by_id, nodes);
+				}
+				MessageSendEvent::UpdateHTLCs { node_id, updates } => {
+					println!("Update {:?}", updates);
+					let idx = node_indices_by_id[&node_id];
+					let other_node=&nodes[idx];
+					for add_htlc in updates.update_add_htlcs {
+						other_node.node.handle_update_add_htlc(&sender.node.node_id(), &add_htlc);
+					}
+					for update_fail_htlc in updates.update_fail_htlcs {
+						other_node.node.handle_update_fail_htlc(&sender.node.node_id(), &update_fail_htlc);
+					}
+					other_node.node.handle_commitment_signed(&sender.node.node_id(), &updates.commitment_signed);
+					to_notify.push(other_node);
+
+					// handle_message_send_events(&other_node, node_indices_by_id, nodes);
+				},
+				_ => panic!("Unexpected event: {:?}", msg),
+			}
+		}
+		// clear monitors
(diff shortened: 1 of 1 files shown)


lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
...
+		};
+		let invoice_payer = InvoicePayer::new(nodes[0].node, router, nodes[0].logger, event_handler, Retry::Attempts(1));
+
+		assert!(invoice_payer.pay_invoice(&create_invoice_from_channelmanager_and_duration_since_epoch(
+			&nodes[1].node, nodes[1].keys_manager, Currency::Bitcoin, Some(100_010_000), "Invoice".to_string(),
+			duration_since_epoch(), 3600).unwrap())
+			.is_ok());
+		let entries=nodes.iter().enumerate().map(|(i, n)| (n.node.node_id(), i));
+		let node_map:HashMap<PublicKey, usize>=entries.collect();
+		handle_message_send_events(&nodes[0], &node_map, &nodes);
```
tnull, line 2534: As mentioned by @TheBlueMatt in chat, it would probably make sense to reuse the `commitment_signed_dance!` macro here, instead of introducing a new recursive helper function.

https://github.com/lightningdevkit/rust-lightning/blob/7bc52aa62cb3c32f51b242ad3fa0ae16aee80cec/lightning/src/ln/functional_test_utils.rs#L1146-L1228

You could consider using this in combination with a loop here, but likely it would make it harder to follow who sends to whom. Since the macro allows to remove the intermediary steps between HTLC updates/failures, this should increase readability quite a bit already.

lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
...
+		}
+		// clear monitors
+		let mut added_monitors = sender.chain_monitor.added_monitors.lock().unwrap();
+		added_monitors.clear();
+	}
+
+	#[test]
+	/// A simpler version of no_extra_retries_on_back_to_back_fail test that demonstrates
+	/// automatically handling and forwarding message send events during a payment inside the test case.
+	fn no_extra_retries_on_back_to_back_fail_automatically_handle_message_send_events() {
```
tnull, line 2471: If the coverage will be the same, it makes sense to entirely replace `no_extra_retries_on_back_to_back_fail` with this refactored function. Would be nice to keep the documenting comments there that explain the reasoning behind the test case.

lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
 		nodes[0].node.process_pending_events(&invoice_payer);
 		assert!(expected_events.borrow().is_empty());
 	}
+
+	///	Helper function to forward HTLC update and RevokeAndAck messages
+	fn handle_message_send_events(sender:&Node, node_indices_by_id: &HashMap<PublicKey, usize>, nodes: &Vec<Node>) {
```
tnull, line 2429: nit
```suggestion
	fn handle_message_send_events(sender: &Node, node_indices_by_id: &HashMap<PublicKey, usize>, nodes: &Vec<Node>) {
```

lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
 		nodes[0].node.process_pending_events(&invoice_payer);
 		assert!(expected_events.borrow().is_empty());
 	}
+
+	///	Helper function to forward HTLC update and RevokeAndAck messages
```
tnull, line 2428: nit
```suggestion
	/// Helper function to forward HTLC update and RevokeAndAck messages
```

lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
...
+	fn handle_message_send_events(sender:&Node, node_indices_by_id: &HashMap<PublicKey, usize>, nodes: &Vec<Node>) {
+		let mut to_notify : Vec<&Node>=Vec::new();
```
tnull, line 2430: nit
```suggestion
		let mut to_notify: Vec<&Node> = Vec::new();
```

lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
...
+		for msg in sender.node.get_and_clear_pending_msg_events() {
+			match msg {
+				MessageSendEvent::SendRevokeAndACK { node_id, msg } => {
+					println!("Sendrevoke {:?}", msg);
```
tnull, line 2434: Print statements should mainly be used for local testing purposes. You may leave them in testcode, however, it then would be nice for them to be full sentences and properly formatted.

lightning-invoice/src/payment.rs
```diff
@@ -2424,4 +2424,143 @@ mod tests {
...
+					let idx = node_indices_by_id[&node_id];
+					let other_node=&nodes[idx];
```
tnull, line 2436:
```suggestion
					let other_node = &nodes[idx];
```

tnull (commented):
After having another look I now understand what you're trying to do here. While not critical, the `no_extra_retries_on_back_to_back_fail` test case could indeed use some refactoring to improve readability. You however should be able to achieve this by reusing said `commitment_signed_dance!" macro.

Other than that I highlighted some formatting nits.

adamritter:
Not really, we can take it out until there's a more complete solution for
evaluating routers inside payments.

PR #1748, opened 2022-09-30
Support custom onion messages

valentinewallace:
In ser_macros, we now allow decode_tlv_stream to decode 1 custom TLV within a
specified range of TLV types. We also need to fail to mid-deserialization if there is
more than 1 TLV in that specified range, otherwise peers could waste our
processing time.

This change is needed to support offers and async payment onion messages, for
the above peers-can-waste-cpu reason.

TODO: more docs, ~fix fuzz~


jkczyz:
Continuing from our offline discussion yesterday, given for offer messages we need the bytes when parsing anyhow, can't we leave `ser_macros.rs` as is and just read each possible message as `Vec<u8>`? Should be cheap to read a sequence of bytes if we aren't parsing them.

jkczyz:
Ah, I guess for custom message support we may need something like this, though.

valentinewallace:
> Continuing from our offline discussion yesterday, given for offer messages we need the bytes when parsing anyhow, can't we leave `ser_macros.rs` as is and just read each possible message as `Vec<u8>`? Should be cheap to read a sequence of bytes if we aren't parsing them.

Per:
> We also need to fail to deserialize if there is
more than 1 TLV in that specified range, otherwise peers could waste our
processing time.

My impression is that peers could still waste a significant amount of our processing time if we don't satisfy this requirement 🤷‍♀️. This sorta makes sense to me because it could end up being a lot of vec allocs even if we just read bytes. But yeah, 100% also custom OMs.

jkczyz:
Thinking about this more, this actually prevents the more general case (outside of onion messages) where there can be more than one custom type in a TLV stream. Given we aren't later going to parse these messages unless the code actually understands the type, I think we only want to pass supported custom types to the macro (rather than a range). This would limit what actually gets parsed.

Otherwise, in the more general case, a malicious sender could still provide a message with many custom types. For onion message payloads, we may still want to configure them as `Vec<u8>` at least for offer messages since we need the raw bytes. So the number of allocations is limited to the number of supported custom types. The question then is about configuring payload parsing with user-supported custom types.

valentinewallace:
> Otherwise, in the more general case, a malicious sender could still provide a message with many custom types. 

We'll fail mid-deserialization if more than one type in the given range is set, so thankfully this shouldn't be an issue. 

I'm a bit confused how we'd support user-provided custom types if we only parse types we understand? I think it's OK to punt the general case of multiple custom TLVs (I'm not clear on whether we ever plan to support this) but open to ideas here.

lightning/src/onion_message/functional_tests.rs
```diff
@@ -80,103 +80,111 @@ fn pass_along_path(path: &Vec<MessengerNode>, expected_path_id: Option<[u8; 32]>
...
 #[test]
 fn two_unblinded_two_blinded() {
 	let nodes = create_nodes(5);
+	let test_tlv = Tlv { tag: 4242, value: vec![42; 32] };
 
 	let secp_ctx = Secp256k1::new();
 	let blinded_route = BlindedRoute::new(&[nodes[3].get_node_pk(), nodes[4].get_node_pk()], &*nodes[4].keys_manager, &secp_ctx).unwrap();
 
-	nodes[0].messenger.send_onion_message(&[nodes[1].get_node_pk(), nodes[2].get_node_pk()], Destination::BlindedRoute(blinded_route), None).unwrap();
+	nodes[0].messenger.send_onion_message(&[nodes[1].get_node_pk(), nodes[2].get_node_pk()], Destination::BlindedRoute(blinded_route),test_tlv, None).unwrap();
```
ariard, line 106: nit: missing space here

lightning/src/util/ser_macros.rs
```diff
@@ -175,7 +175,8 @@ macro_rules! decode_tlv {
 }
 
 macro_rules! decode_tlv_stream {
-	($stream: expr, {$(($type: expr, $field: ident, $fieldty: tt)),* $(,)*}) => { {
+		($stream: expr, {$(($type: expr, $field: ident, $fieldty: tt)),* $(,)*}
+		 $(, ($custom_type: tt, $custom_value: ident, $custom_range_min: expr, $custom_range_max: expr))?) => { {
```
ariard, line 179: Catching up with the conv, IIUC each custom TLV inside the onion would have min/max bounds defined at the protocol-level ? Otherwise if DoS protections are per-implementation, you might fail deser of a message gauged by valid as the sending implementation. I still have in mind the use-case of sending Lightning transactions inside onions, and a commitment transaction size can dynamic on the number of HTLC outputs.

valentinewallace:
> I'm suggesting to additionally have the decoder directly use LDK's parameterized custom handler to parse any types it supports. Compare this to your draft proposal which instead passes to the custom handler any TLV record (even if just one in the onion message case) and having it decide if it knows how to decode the type. That said, it may be a difficult to accomplish the alternative with macros.
> 
> For `send_onion_message`, you would give it some `T: Writeable` instead of a `Tlv`.

Ah, I think I see what you're saying. To restate: in this scheme, we would be initializing each known type (including the custom type) as `None` and reading one of them, failing to deserialize if we read a known type when any of the other known types have already been read? 

The one concern I have is that `T: Writeable` doesn't give us the TLV's `type` number. And they may support reading multiple custom types, so the types would have to be passed in as a list somehow and individually initialized/checked for. May be surmountable though.

> > I think it's OK to punt the general case of multiple custom TLVs (I'm not clear on whether we ever plan to support this) but open to ideas here.
> 
> My hesitancy here is that it is adding onion message semantics to `decode_tlv_stream`.

We could have a separate `decode_onion_message_tlv_stream` and DRY with a `decode_tlv_stream_inner`, not sure if that addresses what you're thinking

lightning/src/util/ser_macros.rs
valentinewallace: >  IIUC each custom TLV inside the onion would have min/max bounds defined at the protocol-level ?

I don't think the bounds are spec'd for custom TLVs, but yup. 

> I still have in mind the use-case of sending Lightning transactions inside onions, and a commitment transaction size can dynamic on the number of HTLC outputs.

Cool, that should be supported. Currently we support reading 1 TLV within the type range `64..u64::max_value()` (can be an invoice_request, etc or custom). Lmk if there's a change desired here.

jkczyz:
> Ah, I think I see what you're saying. To restate: in this scheme, we would be initializing each known type (including the custom type) as `None` and reading one of them, failing to deserialize if we read a known type when any of the other known types have already been read?
> 

Sorta. I'm just saying to treat custom types like standard types since they should be known at compile-time, even if configured by the user. Then at the onion-message layer, enforce that there can only be one.

Given that, I'm not really convinced there should be a concern with excessive heap allocations for onion message payloads since the number of allocations would be be limited to known types assuming each are parsed into `Vec<u8>`. 

> The one concern I have is that `T: Writeable` doesn't give us the TLV's `type` number.

Yeah, would need to either pass a type or do something like `wire::Message` and have `T` further bound by a trait defining the type as a constant.

> And they may support reading multiple custom types, so the types would have to be passed in as a list somehow and individually initialized/checked for. May be surmountable though.
> 

Yeah, I think calling `decode_tlv_stream` with known custom types would be the trickiest part.

> We could have a separate `decode_onion_message_tlv_stream` and DRY with a `decode_tlv_stream_inner`, not sure if that addresses what you're thinking. I guess I don't see how the changes _prevent_ multiple custom TLV support in the future

Ah, sorry. I guess "prevents" was a bit too strong or phrased poorly. I meant custom message support as written can't be reused for the more general case of multiple custom types. So we would have to add another param to `decode_tlv_stream` (and more complexity) later. Whereas, if `decode_tlv_stream` supported the general case instead and handled the onion message semantics (i.e., only one type) higher up, then it could be re-used without needing to add more complexity.

ViktorTigerstrom (commented):
Started reviewing this yesterday before the comment about first proceeding with the in-range-read macro. I'm leaving some comments in case there's value in them once this is picked up again :).

lightning/src/util/ser_macros.rs
ViktorTigerstrom: > IIUC each custom TLV inside the onion would have min/max bounds defined at the protocol-level ? 
IIUC the closest spec is that the generalized TLV format defines that types between 2^16 and u64::max_value() are reserved for custom types.
https://github.com/lightning/bolts/blob/16973e2b857e853308cafd59e42fa830d75b1642/01-messaging.md?plain=1#L131

lightning/src/onion_message/packet.rs
```diff
@@ -156,30 +158,45 @@ impl Writeable for (Payload, [u8; 32]) {
...
+		let (mut message_type, mut message_bytes) = (None, None);
+		let mut decode_custom_tlv = |msg_type, mut msg_reader: &mut FixedLengthReader<&mut &mut FixedLengthReader<&mut R>>| {
+			if msg_type >= 64 {
+				// Don't allow reading more than one data TLV from an onion message.
+				if message_bytes.is_some() || message_type.is_some() {
+					return Err(DecodeError::InvalidValue)
+				}
+				message_type = Some(msg_type);
+				decode_tlv!(msg_reader, message_bytes, vec_type);
+				return Ok(true)
```
valentinewallace, line 192: @jkczyz Should this bool be an enum? Should we get rid of the result and always return a trinary enum? Let me know what you prefer

lightning/src/onion_message/packet.rs
```diff
@@ -188,8 +193,15 @@ impl ReadableArgs<SharedSecret> for Payload {
 				Ok(Payload::Forward(ForwardControlTlvs::Unblinded(tlvs)))
```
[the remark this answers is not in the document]
valentinewallace: >  would we want this to fail and return an Err if message_type is set to an even number and it's a Payload::Forward?

Good catch, adding this!

Re (2) I didn't quite get that, are you asking if we plan to support users sending custom `control` TLVs? I believe that'd be out-of-scope for now

fuzz/src/onion_message.rs
```diff
@@ -117,36 +117,37 @@ mod tests {
...
...000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b200000000000000000000000000000000000000000000000000000000000000";
```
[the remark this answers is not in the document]
valentinewallace: Ah, yeah when padding is added the fuzz vectors will need to be updated again (since presumably most/all OMs will contain padding for at least one hop, which then changes the hmac of the payload). I can send you the test harness I set up to generate these vectors when the PR is open :) (no rush on opening it tho)

lightning/src/onion_message/packet.rs
jkczyz: I had been thinking the custom decoder could signal an unknown type by returning a variant of `DecodeError` (possibly a new variant). This would presumed to be returned only if none of the bytes were consumed (i.e., only when checking the type). Otherwise, another custom decoder could not process it if the previous one consumed some bytes. If we make a trait for the custom decoder, we could have a separate method for checking if it understands the type and another method for decoding, perhaps?

jkczyz: Ah, I think that I misread the `?` repetition as `*` and thought this would handle known payloads (e.g., `invoice_request`) as a separate custom decoder. I guess logic for reading such known types would be added to the closure?

Related: Will handling be separate from decoding? If so, will we need the type still to figure out if handler for a known type is needed or a custom type, IIUC.

valentinewallace:
I think we should be able to clean up the tests to not use log assertions anymore after this, but would appreciate some conceptual feedback on the latest approach first

fuzz/src/onion_message.rs
```diff
@@ -49,6 +50,42 @@ pub extern "C" fn onion_message_run(data: *const u8, datalen: usize) {
...
+		Ok(CUSTOM_MESSAGE_CONTENTS.write(w)?)
+	}
+}
+
+impl CustomMessageReadable for TestCustomMessage {
+	fn read<R: io::Read>(message_type: u64, buffer: &mut R) -> Result<Option<Self>, DecodeError> where Self: Sized {
+		if message_type == CUSTOM_MESSAGE_TYPE {
+			let mut buf = Vec::new();
+			buffer.read_to_end(&mut buf)?;
+			assert_eq!(buf, CUSTOM_MESSAGE_CONTENTS);
```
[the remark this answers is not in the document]
valentinewallace: Oops, thanks

lightning/src/onion_message/packet.rs
```diff
@@ -92,25 +92,70 @@ impl LengthReadable for Packet {
...
+
+impl CustomMessageReadable for () {
+	fn read<R: io::Read>(_message_type: u64, _buffer: &mut R) -> Result<Option<Self>, DecodeError> where Self: Sized {
+		Ok(None)
+	}
+}
+
+impl OnionMessageContents for () {
+	fn tlv_type(&self) -> u64 { u64::max_value() }
+}
```
jkczyz, line 158: We probably don't want to allow writing `()` as the onion message. Could we either separate the traits or use an enum with no variants? Then this could be `unreachable`.

jkczyz (commented):
Feel free to squash

lightning/src/onion_message/messenger.rs
```diff
@@ -55,43 +61,67 @@ use prelude::*;
...
-/// onion_messenger.send_onion_message(&intermediate_hops, Destination::BlindedRoute(blinded_route), reply_path);
+/// # let your_custom_message = YourCustomMessage {};
+/// let message = Message::Custom(your_custom_message);
+/// onion_messenger.send_onion_message(&intermediate_hops, Destination::BlindedRoute(blinded_route), message, reply_path);
 /// ```
 ///
 /// [offers]: <https://github.com/lightning/bolts/pull/798>
 /// [`OnionMessenger`]: crate::onion_message::OnionMessenger
-pub struct OnionMessenger<Signer: Sign, K: Deref, L: Deref>
+pub struct OnionMessenger<Signer: Sign, K: Deref, L: Deref, CMH: Deref>
```
jkczyz, line 113: Docs about only supporting "sending and receiving empty onion messages" need updating.

lightning/src/onion_message/messenger.rs
```diff
@@ -130,26 +160,37 @@ pub enum SendError {
 	BufferFull,
 }
 
-impl<Signer: Sign, K: Deref, L: Deref> OnionMessenger<Signer, K, L>
+/// Handler for custom onion messages.
```
jkczyz, line 163: Would be good to have a more descriptive explanation of how to implement this either here or maybe better in the `OnionMessenger` docs. Doesn't have to be very long so long as the whole picture is described somewhere.

lightning/src/util/ser_macros.rs
```diff
@@ -175,7 +175,8 @@ macro_rules! decode_tlv {
 }
 
 macro_rules! decode_tlv_stream {
-	($stream: expr, {$(($type: expr, $field: ident, $fieldty: tt)),* $(,)*}) => { {
+		($stream: expr, {$(($type: expr, $field: ident, $fieldty: tt)),* $(,)*}
+		 $(, $decode_custom_tlv: expr)?) => { {
```
jkczyz, line 179: This macro is probably requires some docs now.

lightning/src/util/ser_macros.rs
```diff
@@ -175,7 +175,8 @@ macro_rules! decode_tlv {
 }
 
 macro_rules! decode_tlv_stream {
-	($stream: expr, {$(($type: expr, $field: ident, $fieldty: tt)),* $(,)*}) => { {
+		($stream: expr, {$(($type: expr, $field: ident, $fieldty: tt)),* $(,)*}
```
jkczyz, line 178: Unneeded indentation?

lightning/src/onion_message/packet.rs
```diff
@@ -92,25 +92,51 @@ impl LengthReadable for Packet {
...
-// enum Message {
-// 	InvoiceRequest(InvoiceRequest),
-// 	Invoice(Invoice),
-//	InvoiceError(InvoiceError),
-//	CustomMessage<T>,
-// }
+#[derive(Debug)]
+/// The contents of an onion message. In the context of offers, this would be the invoice, invoice
+/// request, or invoice error.
+pub enum Message<T> where T: OnionMessageContents {
```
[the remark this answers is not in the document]
jkczyz: Maybe name it `OnionMessageContents` and rename that `OnionMessageType`?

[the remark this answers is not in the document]
valentinewallace: Went with ` Message --> OnionMessageContents` and `OnionMessageContents --> CustomOnionMessageContents`

jkczyz (dismissed):

valentinewallace:
Disallowed any `data` TLVs in forwarding hops for now. Waiting on spec clarification here: https://github.com/lightning/bolts/pull/759/files#r999879154

Only diff is: 
```
diff --git a/lightning/src/onion_message/packet.rs b/lightning/src/onion_message/packet.rs
index 960b349f..74d253bf 100644
--- a/lightning/src/onion_message/packet.rs
+++ b/lightning/src/onion_message/packet.rs
@@ -230,7 +230,7 @@ impl<T: CustomOnionMessageContents> ReadableArgs<SharedSecret> for Payload<T> {
                match read_adapter {
                        None => return Err(DecodeError::InvalidValue),
                        Some(ChaChaPolyReadAdapter { readable: ControlTlvs::Forward(tlvs)}) => {
-                               if message_type.is_some() && message_type.unwrap() % 2 == 0 {
+                               if message_type.is_some() {
                                        return Err(DecodeError::InvalidValue)
                                }
                                Ok(Payload::Forward(ForwardControlTlvs::Unblinded(tlvs)))

File: lightning/src/onion_message/packet.rs [hunk 13]
```diff
@@ -155,41 +183,66 @@ impl Writeable for (Payload, [u8; 32]) {
...
+				Ok(None) => Ok(false),
+				Err(e) => Err(e),
+			}
 		});
 		rd.eat_remaining().map_err(|_| DecodeError::ShortRead)?;
 
 		match read_adapter {
 			None => return Err(DecodeError::InvalidValue),
 			Some(ChaChaPolyReadAdapter { readable: ControlTlvs::Forward(tlvs)}) => {
+				if message_type.is_some() && message_type.unwrap() % 2 == 0 {
```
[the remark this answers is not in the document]
valentinewallace: Disallowed this

jkczyz (approved):

PR #1753, opened 2022-10-05
Avoid generating redundant claims after initial confirmation

wpaulino:
These claims will never be valid as a previous claim has already confirmed. If a previous claim is reorged out of the chain, a new claim will be generated bypassing the new behavior.

While this doesn't change much for our existing transaction-based claims, as broadcasting an already confirmed transaction acts as a NOP, it prevents us from yielding redundant event-based claims, which will be introduced as part of the anchors patchset.

### DIFF
--- a/lightning/src/chain/onchaintx.rs
+++ b/lightning/src/chain/onchaintx.rs
@@ -430,7 +430,43 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
 		where F::Target: FeeEstimator,
 					L::Target: Logger,
 	{
-		if cached_request.outpoints().len() == 0 { return None } // But don't prune pending claiming request yet, we may have to resurrect HTLCs
+		let request_outpoints = cached_request.outpoints();
+		if request_outpoints.is_empty() {
+			// Don't prune pending claiming request yet, we may have to resurrect HTLCs. Untractable
+			// packages cannot be aggregated and will never be split, so we cannot end up with an
+			// empty claim.
+			debug_assert!(cached_request.is_malleable());
+			return None;
+		}
+		// If we've seen transaction inclusion in the chain for all outpoints in our request, we
+		// don't need to continue generating more claims. We'll keep tracking the request to fully
+		// remove it once it reaches the confirmation threshold, or to generate a new claim if the
+		// transaction is reorged out.
+		let mut all_inputs_have_confirmed_spend = true;
+		for outpoint in &request_outpoints {
+			if let Some(first_claim_txid_height) = self.claimable_outpoints.get(outpoint) {
+				// We check for outpoint spends within claims individually rather than as a set
+				// since requests can have outpoints split off.
+				if !self.onchain_events_awaiting_threshold_conf.iter()
+					.any(|event_entry| if let OnchainEvent::Claim { claim_request } = event_entry.event {
+						first_claim_txid_height.0 == claim_request
+					} else {
+						// The onchain event is not a claim, keep seeking until we find one.
+						false
+					})
+				{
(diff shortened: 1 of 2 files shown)


lightning/src/chain/onchaintx.rs
```diff
@@ -382,7 +382,36 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
 		where F::Target: FeeEstimator,
 					L::Target: Logger,
 	{
-		if cached_request.outpoints().len() == 0 { return None } // But don't prune pending claiming request yet, we may have to resurrect HTLCs
+		let request_outpoints = cached_request.outpoints();
+		if request_outpoints.is_empty() {
+			// Don't prune pending claiming request yet, we may have to resurrect HTLCs
+			return None;
```
ariard, line 388: If the package is `Untractable`, I think this case is pathological ? Like for the commitment transaction, we cannot aggregate them, so all the inputs (here only one) are claimed and we discard the claim request. If I'm correct we can log_error or assert that case.

lightning/src/chain/onchaintx.rs
```diff
@@ -382,7 +382,39 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
+						}
+					} else {
+						// The request's outpoint spend does not exist yet.
+						return false;
+					}
+				}
+				// All of the request's outpoints were spent by the claim event.
+				true
+			} else {
+				false
```
ariard, line 413: nit: "The onchain event is not a claim. Keep seeking if we have a request containing all the outpoints".

lightning/src/chain/onchaintx.rs
```diff
@@ -382,7 +382,39 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
+		// it reaches the confirmation threshold, or to generate a new claim if the transaction is
+		// reorged out.
+		if self.onchain_events_awaiting_threshold_conf.iter()
+			.any(|event_entry| if let OnchainEvent::Claim { claim_request } = event_entry.event {
+				for outpoint in &request_outpoints {
+					if let Some(first_claim_txid_height) = self.claimable_outpoints.get(outpoint) {
+						if first_claim_txid_height.0 != claim_request {
+							// The request's outpoint spend does not correspond to the current claim event.
+							return false;
+						}
```
ariard, line 404: nit: "There is a confirmed claim txid referencing this outpoint".

ariard (commented):
SGTM

arik-so (dismissed):
Seems fine to me, but I'm not quite sure I have sufficient context 🫥

lightning/src/chain/onchaintx.rs
```diff
@@ -430,7 +430,43 @@ impl<ChannelSigner: Sign> OnchainTxHandler<ChannelSigner> {
...
+			debug_assert!(cached_request.is_malleable());
+			return None;
+		}
+		// If we've seen transaction inclusion in the chain for all outpoints in our request, we
+		// don't need to continue generating more claims. We'll keep tracking the request to fully
+		// remove it once it reaches the confirmation threshold, or to generate a new claim if the
+		// transaction is reorged out.
+		let mut all_inputs_have_confirmed_spend = true;
+		for outpoint in &request_outpoints {
+			if let Some(first_claim_txid_height) = self.claimable_outpoints.get(outpoint) {
```
arik-so, line 447: btw, this variable name is a bit confusing, because it sounds like it's the height of the txid, but it's actually the (txid, height) tuple

arik-so: that said, I don't have any better ideas ¯\_(ツ)_/¯

arik-so (approved):

PR #1770, opened 2022-10-14
Add logging in route hint filtering

jkczyz:
Knowing why a channel was not included as an invoice route hint can be valuable. Add logging on the decisions made when filtering channels.

### DIFF
--- a/lightning-invoice/src/utils.rs
+++ b/lightning-invoice/src/utils.rs
@@ -54,14 +54,20 @@ use sync::Mutex;
 /// [`ChannelManager::create_inbound_payment`]: lightning::ln::channelmanager::ChannelManager::create_inbound_payment
 /// [`ChannelManager::create_inbound_payment_for_hash`]: lightning::ln::channelmanager::ChannelManager::create_inbound_payment_for_hash
 /// [`PhantomRouteHints::channels`]: lightning::ln::channelmanager::PhantomRouteHints::channels
-pub fn create_phantom_invoice<Signer: Sign, K: Deref>(
-	amt_msat: Option<u64>, payment_hash: Option<PaymentHash>, description: String, invoice_expiry_delta_secs: u32,
-	phantom_route_hints: Vec<PhantomRouteHints>, keys_manager: K, network: Currency,
-) -> Result<Invoice, SignOrCreationError<()>> where K::Target: KeysInterface {
+pub fn create_phantom_invoice<Signer: Sign, K: Deref, L: Deref>(
+	amt_msat: Option<u64>, payment_hash: Option<PaymentHash>, description: String,
+	invoice_expiry_delta_secs: u32, phantom_route_hints: Vec<PhantomRouteHints>, keys_manager: K,
+	logger: L, network: Currency,
+) -> Result<Invoice, SignOrCreationError<()>>
+where
+	K::Target: KeysInterface,
+	L::Target: Logger,
+{
 	let description = Description::new(description).map_err(SignOrCreationError::CreationError)?;
 	let description = InvoiceDescription::Direct(&description,);
-	_create_phantom_invoice::<Signer, K>(
-		amt_msat, payment_hash, description, invoice_expiry_delta_secs, phantom_route_hints, keys_manager, network,
+	_create_phantom_invoice::<Signer, K, L>(
+		amt_msat, payment_hash, description, invoice_expiry_delta_secs, phantom_route_hints,
+		keys_manager, logger, network,
 	)
 }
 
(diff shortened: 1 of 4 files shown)


lightning-invoice/src/utils.rs
```diff
@@ -381,31 +399,43 @@ where
...
 			// look at the public channels instead.
+			log_debug!(logger, "Not including invoice route hints on account of public channel {}",
+				log_bytes!(channel.channel_id));
 			return vec![]
 		}
 
 		if channel.inbound_capacity_msat >= min_inbound_capacity {
-			min_capacity_channel_exists = true;
+			if !min_capacity_channel_exists {
+				log_debug!(logger, "Channel with enough inbound capacity exists for invoice route hints");
```
jurvis, line 426: nit: should we print channel id here too like in the diffs above?

lightning-invoice/src/utils.rs
```diff
@@ -326,14 +341,15 @@ where
...
 {
-	let route_hints = filter_channels(channelmanager.list_usable_channels(), amt_msat);
-
 	// `create_inbound_payment` only returns an error if the amount is greater than the total bitcoin
 	// supply.
 	let (payment_hash, payment_secret) = channelmanager
 		.create_inbound_payment(amt_msat, invoice_expiry_delta_secs)
 		.map_err(|()| SignOrCreationError::CreationError(CreationError::InvalidAmount))?;
 	let our_node_pubkey = channelmanager.get_our_node_id();
+	let channels = channelmanager.list_usable_channels();
```
jurvis, line 350: Are we using this value other than in +L373? Should we call it inline in that call like -L329?

lightning-invoice/src/utils.rs
```diff
@@ -381,31 +399,43 @@ where
...
 	for channel in channels.iter() {
 		if channel.get_inbound_payment_scid().is_none() || channel.counterparty.forwarding_info.is_none() {
+			log_debug!(logger, "Ignoring channel {} for invoice route hints", log_bytes!(channel.channel_id));
 			continue;
 		}
 
 		if channel.is_public {
 			// If any public channel exists, return no hints and let the sender
 			// look at the public channels instead.
+			log_debug!(logger, "Not including invoice route hints on account of public channel {}",
```
ViktorTigerstrom, line 419: This wording will be pretty confusing for phantom invoices as well. Say a phantom invoice is generated using 3 participating nodes, where only one of the nodes have public channels, but the other 2 have only private channels. That invoice will include invoice route hints for the channels of the 2 participating nodes with private channels only, yet log this message due to the 3rd participating node with public channel(s), hence incorrectly indicating that no invoice route hints were included. 

Also note that phantom invoices will include invoice route hints for every participating node even if all of them have public channels. Phantom hops with the node id for every participating node will be included in the invoice route hints due to:
https://github.com/lightningdevkit/rust-lightning/blob/67726097c28f1ec5598ab9465e4de5a44225e56b/lightning-invoice/src/utils.rs#L178

Perhaps we could phrase this sentence as something like:
"Non of the considered channels are included in the invoice route hints on account of public channel {}."

Not sure what the best way to phrase this is though :)!

lightning-invoice/src/utils.rs
```diff
@@ -381,31 +399,43 @@ where
...
 /// need to find the path by looking at the public channels instead
-fn filter_channels(channels: Vec<ChannelDetails>, min_inbound_capacity_msat: Option<u64>) -> Vec<RouteHint>{
+fn filter_channels<L: Deref>(
+	channels: Vec<ChannelDetails>, min_inbound_capacity_msat: Option<u64>, logger: &L
+) -> Vec<RouteHint> where L::Target: Logger {
 	let mut filtered_channels: HashMap<PublicKey, &ChannelDetails> = HashMap::new();
 	let min_inbound_capacity = min_inbound_capacity_msat.unwrap_or(0);
 	let mut min_capacity_channel_exists = false;
 
+	log_debug!(logger, "Considering {} channels for invoice route hints", channels.len());
```
ViktorTigerstrom, line 409: Hmm I think it'd be pretty useful to also add some logging in `_create_phantom_invoice` for this line to not be too confusing when generating phantom invoices. Say you're generating a phantom invoice using 3 participating nodes with 3 channels each or something. Then this line will be repeated 3 times ("Considering 3 channels for invoice route hints") instead of saying that saying that 9 channels were considered which I think is what people would expect instead.

I think that will be pretty confusing with no context, if you don't also have a good understanding of how the code for phantom invoice generation is structured, and that the channels for each participated node is filtered separately.

I think that can become clearer if there's also some logging in `_create_phantom_invoice` that indicates that route hints will be generated for X participating nodes.
The context will also be clearer by also adding a line of logging in the `_create_phantom_invoice` `for loop` right before the `filter_channels` call, which indicates that channels for participating node with node id {} are about to be considered for invoice route hints.

lightning-invoice/src/utils.rs
```diff
@@ -430,12 +460,22 @@ fn filter_channels(channels: Vec<ChannelDetails>, min_inbound_capacity_msat: Opt
...
+		.into_iter()
+		.map(|(_, channel)| channel)
+		.filter(|channel| {
+			let has_enough_inbound_capacity = channel.inbound_capacity_msat >= min_inbound_capacity;
+			let include_channel = !min_capacity_channel_exists || has_enough_inbound_capacity;
+			if !include_channel {
+				log_debug!(logger, "Ignoring channel {} without enough capacity for invoice route hints",
+					log_bytes!(channel.channel_id));
+			}
+			include_channel
```
ViktorTigerstrom, line 473: This will clash a little with the cleanup in #1769 :). I have no preference for which version we go with, but could be worth considering by either version.

ViktorTigerstrom (commented):
Looks good and is definitely helpful :)! Just leaving some suggestions for clarifications for phantom invoice route hints filtering.

lightning-invoice/src/utils.rs
jkczyz: Prefer to leave it out as it (and any others meeting the criteria) will be printed later after filtering.

lightning-invoice/src/utils.rs
jkczyz: I moved it to keep all the `ChannelManager` calls together. Note that `our_node_pubkey` is only used once, too.

jkczyz:
> Feel free to squash, basically LGTM but the super verbose channel-inclusion-exclusion stuff may make more sense as `TRACE`, or at least the why-was-it-excluded ones.

Hmm... seems like why a channel was excluded is more important than which channels were included given the latter can be gleaned from the invoice? I'm fine with making everything `log_trace`, if you'd like.

And anything logged outside `filter_channels` isn't all that important without having the logging in `filter_channels`.

jkczyz:
Squashed and changed all logging to `log_trace` for now.

lightning-invoice/src/utils.rs
```diff
@@ -381,39 +405,57 @@ where
...
+			}
 		}
+
 		match filtered_channels.entry(channel.counterparty.node_id) {
 			hash_map::Entry::Occupied(mut entry) => {
 				let current_max_capacity = entry.get().inbound_capacity_msat;
 				if channel.inbound_capacity_msat < current_max_capacity {
 					continue;
 				}
+				log_trace!(logger, "Preferring channel {} ({} msats) over {} ({} msats) for invoice route hints",
```
ViktorTigerstrom, line 456: Could also be worth mentioning the counterparty node id for the channel here, due to the filtering only including 1 channel per counterparty node.

ViktorTigerstrom (dismissed):
LGTM, leaving one suggestion that's just a nice to have!

lightning-invoice/src/utils.rs
```diff
@@ -381,39 +405,57 @@ where
...
+			if !online_channel_exists {
+				log_trace!(logger, "Channel with connected peer exists for invoice route hints");
+				online_channel_exists = true;
+			}
 		}
+
 		match filtered_channels.entry(channel.counterparty.node_id) {
 			hash_map::Entry::Occupied(mut entry) => {
 				let current_max_capacity = entry.get().inbound_capacity_msat;
 				if channel.inbound_capacity_msat < current_max_capacity {
```
ViktorTigerstrom, line 453: Actually sorry, we should probably log here that the channel is ignored,?

ViktorTigerstrom:
That's ok for me @TheBlueMatt! The new changes looks good. They could possibly be dryed up a little instead of being repeated twice, but then again being this explicit makes it easier to read, so the current version is fine for me:). LGTM once squashed.

ViktorTigerstrom (approved):

lightning-invoice/src/utils.rs
```diff
@@ -353,6 +375,8 @@ where
 	if let Some(amt) = amt_msat {
 		invoice = invoice.amount_milli_satoshis(amt);
 	}
+
+	let route_hints = filter_channels(channels, amt_msat, &logger);
```
valentinewallace, line 379: Some changes in this method seem unrelated

valentinewallace (approved):

lightning-invoice/src/utils.rs
jkczyz: I need the `filter_channels` logging to come after the logging above, the latter of which needs `payment_hash`.

PR #1771, opened 2022-10-14
Add onion message blinded control tlv payload padding

ViktorTigerstrom:
Adds padding to blinded control tlv payloads in blinded routes. This addresses one of the follow-ups in #1607.

TODO:
This also needs to update fuzz tests.

### DIFF
--- a/lightning/src/onion_message/blinded_route.rs
+++ b/lightning/src/onion_message/blinded_route.rs
@@ -200,3 +200,127 @@ impl Writeable for ReceiveTlvs {
 		Ok(())
 	}
 }
+
+#[cfg(test)]
+mod test {
+	use bitcoin::secp256k1::{PublicKey, SecretKey, Secp256k1};
+	use ::get_control_tlv_length;
+	use super::{ForwardTlvs, ReceiveTlvs, blinded_hops};
+	use util::ser::{VecWriter, Writeable};
+
+	#[test]
+	fn padding_is_correctly_serialized() {
+		let max_length = get_control_tlv_length!(true, true);
+
+		let dummy_next_node_id = PublicKey::from_slice(&hex::decode("030101010101010101010101010101010101010101010101010101010101010101").unwrap()[..]).unwrap();
+		let dummy_blinding_override = PublicKey::from_slice(&hex::decode("030202020202020202020202020202020202020202020202020202020202020202").unwrap()[..]).unwrap();
+		let dummy_path_id = [1; 32];
+
+		let no_padding_tlv = ForwardTlvs {
+			next_node_id: dummy_next_node_id,
+			next_blinding_override: Some(dummy_blinding_override),
+			total_length: max_length,
+		};
+
+		let blinding_override_padding_tlv = ForwardTlvs {
+			next_node_id: dummy_next_node_id,
+			next_blinding_override: None,
+			total_length: max_length,
+		};
+
+		let recieve_tlv_padding_tlv = ReceiveTlvs {
+			path_id: Some(dummy_path_id),
+			total_length: max_length,
+		};
+
+		let full_padding_tlv = ReceiveTlvs {
+			path_id: None,
+			total_length: max_length,
+		};
+
+		let mut w = VecWriter(Vec::new());
+		no_padding_tlv.write(&mut w).unwrap();
+		let serialized_no_padding_tlv = w.0;
+		// As `serialized_no_padding_tlv` is the longest tlv, no padding is expected.
+		// Expected data tlv is:
+		// 1. 4 (type) for `next_node_id`
+		// 2. 33 (length) for the length of a point/public key
+		// 3. 33 bytes of the `dummy_next_node_id`
+		// 4. 8 (type) for `next_blinding_override`
+		// 5. 33 (length) for the length of a point/public key
+		// 6. 33 bytes of the `dummy_blinding_override`
+		let expected_serialized_no_padding_tlv_payload = &hex::decode("04210301010101010101010101010101010101010101010101010101010101010101010821030202020202020202020202020202020202020202020202020202020202020202").unwrap()[..];
+		assert_eq!(serialized_no_padding_tlv, expected_serialized_no_padding_tlv_payload);
+		assert_eq!(serialized_no_padding_tlv.len(), max_length as usize);
+
+		w = VecWriter(Vec::new());
+		blinding_override_padding_tlv.write(&mut w).unwrap();
+		let serialized_blinding_override_padding_tlv = w.0;
+		// As `serialized_blinding_override_padding_tlv` has no `next_blinding_override`, 35 bytes
+		// of padding is expected (the serialized length of `next_blinding_override`).
+		// Expected data tlv is:
+		// 1. 1 (type) for padding
+		// 2. 33 (length) given the length of a the missing `next_blinding_override`
+		// 3. 33 0 bytes of padding
+		// 4. 4 (type) for `next_node_id`
+		// 5. 33 (length) for the length of a point/public key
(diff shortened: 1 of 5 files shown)


lightning/src/onion_message/functional_tests.rs
```diff
@@ -180,3 +182,75 @@ fn peer_buffer_full() {
...
+
+	let only_unblinded_payloads = packet_payloads_and_keys(&secp_ctx, &[nodes[0].get_node_pk(), nodes[1].get_node_pk()], Destination::Node(nodes[2].get_node_pk()), None,  &session_priv).unwrap().0;
+	let one_unblinded_and_three_blinded_payloads = packet_payloads_and_keys(&secp_ctx, &[nodes[1].get_node_pk()], Destination::BlindedRoute(three_blinded_hops), None,  &session_priv).unwrap().0;
+	// When more that one unblinded payload exists, the blinded payloads should be the same length
+	// as the largest unblinded payload.
+	let multiple_unblinded_and_blinded_payloads = packet_payloads_and_keys(&secp_ctx, &[nodes[0].get_node_pk(), nodes[1].get_node_pk()], Destination::BlindedRoute(two_blinded_hops), None,  &session_priv).unwrap().0;
+
+	// Verify that the blinded contol tlv payloads returned from `packet_payloads_and_keys` have
+	// the same length, and that the payload for every blinded payload matches the length of the
+	// largest unblinded payload length.
```
ViktorTigerstrom, line 202: The spec is a little unclear in regarding this, but the test vectors seems to indicate that this should be the case.

lightning/src/onion_message/blinded_route.rs
```diff
@@ -13,9 +13,11 @@ use bitcoin::secp256k1::{self, PublicKey, Secp256k1, SecretKey};
 
 use chain::keysinterface::KeysInterface;
 use super::utils;
+use ::get_control_tlv_length;
```
ViktorTigerstrom, line 16: This seems bad. Will check why `#[macro_export]` places the `get_control_tlv_length` in the root of the lightning crate for `onion_message::utils`, but not for in `ln::functional_test_utils` and similar.

lightning/src/onion_message/packet.rs
```diff
@@ -239,15 +247,57 @@ impl Readable for ControlTlvs {
...
 	#[inline]
 	fn read<R: Read>(reader: &mut R) -> Result<Self, DecodeError> {
+		let mut padding_length = 0;
 		loop {
 			let mut buf = [0; 8192];
-			if reader.read(&mut buf[..])? == 0 { break; }
+			let read_bytes_len = reader.read(&mut buf[..])?;
+			if read_bytes_len == 0 { break; }
+			for n in 0..read_bytes_len {
+				if buf[n] != 0 as u8 { return Err(DecodeError::InvalidValue); }
```
valentinewallace, line 264: I don't think the padding is required to be 0s, or at least I can't find it in the spec ?

valentinewallace (commented):
In terms of overall approach, I think we can simplify it by hardcoding the padding length used in `ReceiveControlTlvs`, and never including padding in `ForwardTlvs` for now. 

We'll need `const`s for the max expected encrypted payload size (basically the `sizeof(next_node_id + tlv type/len)`) and the receive TLV padding size, and a test that fails if either of those `const`s changes (which should fit well into the equal-length-payloads test you wrote, IIUC)

However, since the spec test vectors include padding for every hop and I'm not sure why, feel free to wait on Rusty getting back to me on the question I linked below

lightning/src/onion_message/blinded_route.rs
```diff
@@ -54,10 +56,11 @@ impl BlindedRoute {
...
 	///
 	/// Errors if less than two hops are provided or if `node_pk`(s) are invalid.
-	//  TODO: make all payloads the same size with padding + add dummy hops
-	pub fn new<K: KeysInterface, T: secp256k1::Signing + secp256k1::Verification>
-		(node_pks: &[PublicKey], keys_manager: &K, secp_ctx: &Secp256k1<T>) -> Result<Self, ()>
-	{
+	//  TODO: Add dummy hops
+	pub fn new<K: KeysInterface, T: secp256k1::Signing + secp256k1::Verification> (
+		node_pks: &[PublicKey], keys_manager: &K, secp_ctx: &Secp256k1<T>,
+		include_next_blinding_override_padding: bool
```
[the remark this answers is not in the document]
ViktorTigerstrom: Nice, thanks for asking :)! 

IIUC another case where we also need to support padding for next blinding override in blinded `ForwardTlvs` is for  "concatenation of two blinded routes", included in the route blinding test vectors [bolt04/route-blinding-test.json](https://github.com/lightning/bolts/pull/759/files#diff-c7c1e9ac1c9959d9dbf2544f0717edd705f47079a58e88dacb838453e4206446) and detailed in [04-onion-routing.md](https://github.com/lightning/bolts/pull/759/files#diff-ad65f0beaac5cef88f5fd7a8b9ca36cbc5a790f4815b4b50b9ea794a55aaf012).
Though perhaps it's too early to start preparing for support of that :sweat_smile:.

lightning/src/onion_message/packet.rs
ViktorTigerstrom: Oh yeah, my bad! Good catch :)

ViktorTigerstrom:
Thanks a lot for the review @valentinewallace! Definitely agree that it'd be great if we can remove the need for support for `next_blinding_override` padding in `BlindedRoute`, to simplify things!

> In terms of overall approach, I think we can simplify it by hardcoding the padding length used in `ReceiveControlTlvs`, and never including padding in `ForwardTlvs` for now.

Hmmm sadly even if the `next_blinding_override` padding is removed from the onion message test vectors it'll be hard to support them if we aim to create the entire route using our `ControlTlvs` if we also effectively hard code no padding for `ForwardTlvs` and a set max size for `ReceiveControlTlvs`, as `Bob's` payload contains a custom tag, and all the other payloads would need to add padding to account for that.
Or do we for test vector coverage aim to skip the step of creating the packet using our `ControlTlvs` and verifying that the created packet matches the test vector `onion_message_packet`, and instead opt for just the second step of decoding the raw test vector `onion_message_packet`, and verifying that each hop is decrypted correctly?

> However, since the spec test vectors include padding for every hop and I'm not sure why, feel free to wait on Rusty getting back to me on the question I linked below

Sounds good, thanks for asking!

lightning/src/onion_message/blinded_route.rs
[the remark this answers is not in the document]
valentinewallace: > "concatenation of two blinded routes", included in the route blinding test vectors [bolt04/route-blinding-test.json](https://github.com/lightning/bolts/pull/759/files#diff-c7c1e9ac1c9959d9dbf2544f0717edd705f47079a58e88dacb838453e4206446) and detailed in [04-onion-routing.md](https://github.com/lightning/bolts/pull/759/files#diff-ad65f0beaac5cef88f5fd7a8b9ca36cbc5a790f4815b4b50b9ea794a55aaf012).
Though perhaps it's too early to start preparing for support of that 😅.

I've always disagreed with that spec wording, bc that just means sending from a list of unblinded hops to a blinded route (i.e. `send_onion_message(vec![intermed_node_1, intermed_node_2], Destination::BlindedRoute(..))` in the current codebase)

ViktorTigerstrom:
> Needs rebase and some comments addressed.
Yes, will pick this up again :). Though we're still waiting for the spec clarifications Val requested (https://github.com/lightningdevkit/rust-lightning/pull/1771#discussion_r997271129), but I'll address the comments I can address meanwhile.

valentinewallace:
Superceded by #3177

PR #1773, opened 2022-10-15
Cleanup lock orders

ViktorTigerstrom:
This is based on #1772, so it can be marked as blocked on dependent pr for now. Will also leave it as draft until #1772 is merged.

The first commit moves the lock_order of `pending_inbound_payments` as well as `pending_outbound_payments` to before the `channel_state` lock, which is to be removed. This will simplify #1507.

The second commit `pending_background_events` into a separate lock order branch.

### DIFF
--- a/lightning/src/ln/channelmanager.rs
+++ b/lightning/src/ln/channelmanager.rs
@@ -401,13 +401,6 @@ pub(super) struct ChannelHolder<Signer: Sign> {
 	/// SCIDs being added once the funding transaction is confirmed at the channel's required
 	/// confirmation depth.
 	pub(super) short_to_chan_info: HashMap<u64, (PublicKey, [u8; 32])>,
-	/// Map from payment hash to the payment data and any HTLCs which are to us and can be
-	/// failed/claimed by the user.
-	///
-	/// Note that while this is held in the same mutex as the channels themselves, no consistency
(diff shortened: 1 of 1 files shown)


lightning/src/ln/channelmanager.rs
```diff
@@ -681,23 +674,25 @@ pub type SimpleRefChannelManager<'a, 'b, 'c, 'd, 'e, M, T, F, L> = ChannelManage
 //  |
 //  |__`forward_htlcs`
 //  |
-//  |__`channel_state`
+//  |__`pending_background_events`
```
ViktorTigerstrom, line 677: I'd just like to ask if `process_background_events` through `timer_tick_occurred` can be called by a thread while another thread has other locks acquired?  If so, I'll drop the second commit that moves this.

ViktorTigerstrom: I've realized in hindsight that this question is rather stupid, as "background threads" acquiring `pending_background_events` is fine even if other threads are holding other locks at the same time. We just need to ensure that every single thread acquires the locks according to the lock order to avoid deadlocks.

ViktorTigerstrom:
Closing this due to most of this already having been addressed through https://github.com/lightningdevkit/rust-lightning/pull/1772.

PR #1779, opened 2022-10-18
Add KeysInterface::get_node_id method

valentinewallace:
Useful since we're working on getting rid of KeysInterface::get_node_secret to complete support for remote signing.

Will be used in upcoming work to check whether an outbound onion message blinded path has our node id as the introduction node id


lightning/src/chain/keysinterface.rs
```diff
@@ -412,6 +412,17 @@ pub trait KeysInterface {
 	/// This method must return the same value each time it is called with a given `Recipient`
 	/// parameter.
 	fn get_node_secret(&self, recipient: Recipient) -> Result<SecretKey, ()>;
+	/// Get node id based on the provided [`Recipient`]. This public key corresponds to the secret in
+	/// [`get_node_secret`].
+	///
+	/// This method must return the same value each time it is called with a given `Recipient`
+	/// parameter.
```
ariard, line 419: Could precise an API error is return if the type of `Recipient` isn't supported by the implementation. Good to warn caller.

ariard (commented):
ACK 17c0fc8

dunxen (dismissed):

dunxen (approved):

PR #1780, opened 2022-10-18
Bump workspace to rust edition 2018

wpaulino:
Mostly motivated by the need of async/await.

To reproduce, follow https://doc.rust-lang.org/cargo/commands/cargo-fix.html#edition-migration.


dunxen:
Nice. Could you maybe list the combination of commands and args you used? Seems I keep missing some areas. Struggling to reproduce.

dunxen:
> It was quite troublesome, I had to apply the migration multiple times due to numerous imports gated behind different features and config flags. I then committed the changes after each iteration and squashed them all into a single commit.

Yeah I think that's pretty much the trouble I'm dealing with. Not getting all the combos of config and features. I'll attempt again in the morning. Think I just had fuzzing left then maybe I can share a repro.

tnull:
Seems this needs a rebase now?

dunxen:
Alright, I've reproduced this but pretty much also lost track of all the steps. Phew!

lightning-invoice/src/de.rs
```diff
@@ -715,15 +715,15 @@ impl From<ParseError> for ParseOrSemanticError {
...
+impl From<crate::SemanticError> for ParseOrSemanticError {
 	fn from(e: SemanticError) -> Self {
 		ParseOrSemanticError::SemanticError(e)
 	}
 }
 
 #[cfg(test)]
 mod test {
-	use de::ParseError;
+	use crate::de::ParseError;
```
tnull, line 726: Could use `super` here and in quite a few places below.
```suggestion
	use super::ParseError;
```

tnull (commented):
Looks good, now my finger hurts from scrolling.

lightning-invoice/src/de.rs
tnull: Yeah, fair enough.

dunxen:
Oof. All those conflicts 😢

Basically looks good to me unless we still wanted something split out of here into a follow up, or are we happy with the whole package?

dunxen (approved):
Not sure if we wanted to split any of these changes out as discussed on discord, but this LGTM. Bring on the merge conflicts?

G8XSU (approved):

jkczyz (approved):
Everything looks crate!

jkczyz:
> Everything looks crate!

Except the CI error... looks transient.

PR #1783, opened 2022-10-19
Signature weight should be estimated at 73 based on BOLT-3

devrandom:


dunxen (approved):
True story:

https://github.com/lightning/bolts/blob/16973e2b857e853308cafd59e42fa830d75b1642/03-transactions.md?plain=1#L407
https://bitcoin.stackexchange.com/a/77192/106541

lightning/src/ln/channel.rs
```diff
@@ -1676,7 +1676,7 @@ impl<Signer: Sign> Channel<Signer> {
 		1 +                                            // witness element count
 		4 +                                            // 4 element lengths (2 sigs, multisig dummy, and witness script)
 		self.get_funding_redeemscript().len() as u64 + // funding witness script
-		2*(1 + 71);                                    // two signatures + sighash type flags
+		2*(1 + 71 + 1);                                // two signatures + sighash type flags + sig lengths
```
jkczyz, line 1679: The length is accounted for two lines up. The reason for the difference is that signatures are variable length.

devrandom: ah, I see.  so my comment is wrong, but the change is correct, right?

devrandom: i.e. it should be `2*(1 + 72)`

devrandom: done

jkczyz (approved):

dunxen (approved):

devrandom:
So this is not a protocol level issue, because the fees for mutual-close are negotiated as total fees rather than feerate.  The only real concern is that we would have an unrelayable tx, or that fee negotiations will fail, right?

Looking at the actual difference between 72 and 73 - it's all about low-S, and BOLT-3 in general doesn't assume it.  Wouldn't it be a bit strange to not assume low-S for other tx types, but assume it for mutual-close?

devrandom:
Submitted https://github.com/lightning/bolts/issues/1034

PR #1787, opened 2022-10-20
Handle events asynchronously in the BackgroundProcessor's async variant

wpaulino:
Fixes https://github.com/lightningdevkit/rust-lightning/issues/1674.

### DIFF
--- a/lightning/src/util/events.rs
+++ b/lightning/src/util/events.rs
@@ -1400,6 +1400,10 @@ pub trait OnionMessageProvider {
 ///
 /// Events are processed by passing an [`EventHandler`] to [`process_pending_events`].
 ///
+/// Implementations of this trait may also feature an async version of event handling, as shown with
+/// [`ChannelManager::process_pending_events_async`] and
+/// [`ChainMonitor::process_pending_events_async`].
+///
 /// # Requirements
 ///
 /// When using this trait, [`process_pending_events`] will call [`handle_event`] for each pending
@@ -1426,6 +1430,8 @@ pub trait OnionMessageProvider {
 /// [`handle_event`]: EventHandler::handle_event
 /// [`ChannelManager::process_pending_events`]: crate::ln::channelmanager::ChannelManager#method.process_pending_events
 /// [`ChainMonitor::process_pending_events`]: crate::chain::chainmonitor::ChainMonitor#method.process_pending_events
+/// [`ChannelManager::process_pending_events_async`]: crate::ln::channelmanager::ChannelManager::process_pending_events_async
+/// [`ChainMonitor::process_pending_events_async`]: crate::chain::chainmonitor::ChainMonitor::process_pending_events_async
 pub trait EventsProvider {
 	/// Processes any events generated since the last call using the given event handler.
 	///
@@ -1434,6 +1440,10 @@ pub trait EventsProvider {
 }
 
 /// A trait implemented for objects handling events from [`EventsProvider`].
+///
+/// An async variation also exists for implementations of [`EventsProvider`] that support async
+/// event handling. The async event handler should satisfy the generic bounds: `F:
+/// core::future::Future, H: Fn(Event) -> F`.
 pub trait EventHandler {
 	/// Handles the given [`Event`].
 	///


lightning/src/routing/gossip.rs
```diff
@@ -274,11 +274,11 @@ where C::Target: chain::Access, L::Target: Logger
 	}
 }
 
-impl<L: Deref> EventHandler for NetworkGraph<L> where L::Target: Logger {
-	fn handle_event(&self, event: &Event) {
+impl<L: Deref> NetworkGraph<L> where L::Target: Logger {
+	pub fn handle_event(&self, event: &Event) {
```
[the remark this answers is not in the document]
jkczyz: Also, probably can have this just take a `&NetworkUpdate` and move the event handling portion to the caller now.

lightning-invoice/src/payment.rs
```diff
@@ -733,8 +733,69 @@ where
...
+pub struct InvoicePayerEventHandler<P: Deref, R: Router, L: Deref, T: Time, E: EventHandler>
+where
+	P::Target: Payer,
+	L::Target: Logger,
+{
+	/// A utility for paying [`Invoice`]s and sending spontaneous payments.
+	pub invoice_payer: InvoicePayerUsingTime<P, R, L, T>,
+	/// An event handler to continue handling any events intercepted by the InvoicePayer.
+	pub event_handler: E,
+}
```
jkczyz, line 755: Hmm... this does't make for great UX when calling methods on `InvoicePayer`. Isn't clear to me why this is required. Could you clarify?

lightning-invoice/src/payment.rs
```diff
@@ -659,77 +658,136 @@ fn has_expired(route_params: &RouteParameters) -> bool {
...
 		}
 
-		// Delegate to the decorated event handler unless the payment is retried.
-		self.event_handler.handle_event(event)
+		Some(event)
+	}
+}
+
+/// A utility for paying [`Invoice`]s and sending spontaneous payments that intercepts relevant
+/// events and forwards them to the EventHandler.
```
tnull, line 735:
```suggestion
/// events and forwards them to the [`EventHandler`].
```

lightning/src/util/events.rs
```diff
@@ -1339,6 +1339,8 @@ pub trait OnionMessageProvider {
 ///
 /// Events are processed by passing an [`EventHandler`] to [`process_pending_events`].
 ///
+/// Implementations of this event may also feature an async version of event handling.
```
tnull, line 1342: Possibly?
```suggestion
/// Implementations of this may also feature an async version of event handling.
```

lightning-invoice/src/payment.rs
jkczyz: I guess it doesn't need to be tuple. Just seems like exposing the wrapper is unnecessary if it holds a reference instead.

jkczyz: To clarify, I'm saying just hide the wrapper in `BackgroundProcessor` if `InvoicePayer` no longer implements `EventHandler`. Each the user and `BackgroundProcessor` would have a reference to `InvoicePayer`. The user for paying and the `BackgroundProcessor` for delegating event handling.

lightning-background-processor/src/lib.rs
```diff
@@ -217,24 +217,24 @@ impl<
 	L: Deref,
 > EventHandler for DecoratingEventHandler<'a, E, PGS, RGS, G, A, L>
 where A::Target: chain::Access, L::Target: Logger {
-	fn handle_event(&self, event: &Event) {
+	fn handle_event(&self, event: Event) {
 		if let Some(network_graph) = self.gossip_sync.network_graph() {
-			network_graph.handle_event(event);
+			if let Event::PaymentPathFailed { ref network_update, .. } = event {
+				if let Some(network_update) = network_update {
```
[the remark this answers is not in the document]
jkczyz: Would simply adding a helper function be sufficient? There's nothing `async` about this block of code.

lightning-invoice/src/payment.rs
```diff
@@ -659,77 +661,136 @@ fn has_expired(route_params: &RouteParameters) -> bool {
...
+    }
+}
+
+/// A utility for paying [`Invoice`]s and sending spontaneous payments that intercepts relevant
+/// events and forwards them to the async event handler.
+///
+/// See [module-level documentation] for details.
+///
+/// [module-level documentation]: crate::payment
+pub struct AsyncInvoicePayerEventHandler<P: Deref, R: Router, L: Deref, T: Time, F: Future, H: Fn(Event) -> F>
```
jkczyz, line 772: Should we define a trait instead of only supporting a function?

lightning-invoice/src/payment.rs
jkczyz: We could implement `Deref` on it to avoid needing to access the `invoice_payer` field. I'd rather have methods on `InvoicePayer` consume `self`, pass an `{Async}EventHandler`, and return the `Deref`able wrapper than exposing the details to the user.

Another, perhaps preferable idea: Keep the `E` type parameter on `InvoicePayer` and have two different `impl` blocks for `new` each with different trait bounds.... probably would need different function names. And similarly, only implement `EventHandler` and the async trait on `impl` blocks with the respective trait bounds. I *think* this would work though each trait may need a common supertrait like we do in `Features` (not sure).

jkczyz: > Not sure I understood this?

Don't have the user manually construct a wrapper using `pub` fields. Have a method on `InvoicePayer` do it instead.
> 
> Hmm, ugh. Let's try to keep the class hierarchy simply for now, and clean it all up by moving retry logic into ChannelManager instead.

This doesn't change the class hierarchy. It just makes some methods/traits conditionally implemented based on the parameterized type. If anything, it keeps the class hierarchy as it was rather than adding wrapper types.

jkczyz: Yeah, looks good. Will need to update the description in the module-level docs now that `InvoicePayer` no longer directly implements `EventHandler`.

jkczyz: Sure thing, happy to. We could also do the other approach here: https://github.com/lightningdevkit/rust-lightning/pull/1787#discussion_r1013460196

lightning/src/chain/chainmonitor.rs
```diff
@@ -475,10 +475,26 @@ where C::Target: chain::Filter,
...
+	///
+	/// See the trait-level documentation of [`EventsProvider`] for requirements.
+	///
+	/// [`EventsProvider`]: crate::util::events::EventsProvider
+	pub async fn process_pending_events_async<Future: core::future::Future, H: Fn(Event) -> Future>(&self, handler: H) {
+		let mut pending_events = Vec::new();
+		for monitor_state in self.monitors.read().unwrap().values() {
+			pending_events.append(&mut monitor_state.monitor.get_and_clear_pending_events());
+		}
+		for event in pending_events.drain(..) {
```
jkczyz, line 494: No need to use `drain`. Without it `into_iter` will be used.

lightning/src/chain/chainmonitor.rs
```diff
@@ -475,10 +475,26 @@ where C::Target: chain::Filter,
 	pub fn get_and_clear_pending_events(&self) -> Vec<events::Event> {
 		use crate::util::events::EventsProvider;
 		let events = core::cell::RefCell::new(Vec::new());
-		let event_handler = |event: &events::Event| events.borrow_mut().push(event.clone());
+		let event_handler = |event: events::Event| events.borrow_mut().push(event.clone());
```
jkczyz, line 478: `clone` no longer needed

lightning/src/ln/channelmanager.rs
```diff
@@ -5675,6 +5675,37 @@ impl<M: Deref, T: Deref, K: Deref, F: Deref, L: Deref> ChannelManager<M, T, K, F
...
+		}
+
+		let mut pending_events = mem::replace(&mut *self.pending_events.lock().unwrap(), vec![]);
+		if !pending_events.is_empty() {
+			result = NotifyOption::DoPersist;
+		}
+
+		for event in pending_events.drain(..) {
+			handler(event).await;
+		}
```
jkczyz, line 5703: `drain` not needed here, either, and can then remove `mut` from `pending_events`.

lightning-invoice/src/payment.rs
```diff
@@ -744,77 +752,105 @@ fn has_expired(route_params: &RouteParameters) -> bool {
...
 }
 
-impl<P: Deref, R: ScoringRouter, L: Deref, E: EventHandler, T: Time> EventHandler for InvoicePayerUsingTime<P, R, L, E, T>
+impl<P: Deref, R: ScoringRouter, L: Deref, E: sealed::BaseEventHandler, T: Time> InvoicePayerUsingTime<P, R, L, E, T>
 where
 	P::Target: Payer,
 	L::Target: Logger,
 {
-	fn handle_event(&self, event: &Event) {
+	fn handle_event_internal(&self, event: Event) -> Option<Event> {
```
jkczyz, line 760: I don't have a strong opinion here, but since this is no longer an `EventHandler` implementation, you can still take the `Event` by reference and return a `bool` instead of moving in and back out. But maybe the new interface is clearer.

lightning/src/util/events.rs
```diff
@@ -1400,6 +1400,8 @@ pub trait OnionMessageProvider {
 ///
 /// Events are processed by passing an [`EventHandler`] to [`process_pending_events`].
 ///
+/// Implementations of this trait may also feature an async version of event handling.
```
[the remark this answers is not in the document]
jkczyz: Linking the implementations SGTM. May want to link the async functions as well.

jkczyz (approved):

